PR #9, opened 2020-07-21
Add Pull Request template.

carrieedwards:
Add standard PR template required by New Relic Product.


.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
+_Before contributing, please read our [contributing guidelines](./CONTRIBUTING.rst) and [code of conduct](./CODE_OF_CONDUCT.md)._
```
a-feld, line 1: These links are broken since the contributing and code of conduct files are not at this level. I think these links need to be absolute paths though given they're going to be in a template.

PR #23, opened 2020-10-02
Use is_alive in place of deprecated isAlive

stianjensen:
threading.Thread.isAlive() is removed in python3.9

> The isAlive() method of threading.Thread has been removed. It was deprecated since Python 3.8. Use is_alive() instead. (Contributed by Dong-hee Na in bpo-37804.)

https://docs.python.org/3.9/whatsnew/3.9.html#removed


a-feld (approved):
Thanks for the contribution @stianjensen !

PR #26, opened 2020-10-12
Fix DeprecationWarning in import_hook

sigvef:

# Overview

For Python >=3.4, using find_loader is deprecated, and importing it
generates a DeprecationWarning. Mirroring wrapt's implementation of a
similar construct in its importer.py (on which this code seems modeled),
we can try to use the new find_spec first,  then the old python 2.7
fallback, as before. Since the newrelic python agent only supports 2.7+
and 3.5+, there is no need to keep support for the deprecated
find_loader at all.

The immediate result of this change is that one less DeprecationWarning
is fired at runtime when using the newrelic python agent. There are
still other DeprecationWarnings originating in this file that are not
fixed by this commit.

# Testing
exercise your changes.

### DIFF
--- a/newrelic/api/import_hook.py
+++ b/newrelic/api/import_hook.py
@@ -18,11 +18,6 @@ import sys
 
 _logger = logging.getLogger(__name__)
 
-try:
-    from importlib import find_loader
-except ImportError:
-    find_loader = None
-
 _import_hooks = {}
 
 _ok_modules = (
@@ -173,25 +168,14 @@ class ImportHookFinder:
         self._skip[fullname] = True
 
         try:
-            # For Python 3 we need to use find_loader() from the new
-            # importlib module.
-
-            if find_loader:
-                loader = find_loader(fullname, path)
-
-                if loader:
-                    return _ImportHookChainedLoader(loader)
-
-            else:
+            try:
+                from importlib.util import find_spec
+                return _ImportHookChainedLoader(find_spec(fullname).loader)
+            except AttributeError:
+                return None
+            except ImportError:
                 __import__(fullname)
-
-                # If we get this far then the module we are
-                # interested in does actually exist and so return
-                # our loader to trigger import hooks and then return
-                # the module.
-
                 return _ImportHookLoader()
-
         finally:
             del self._skip[fullname]
 
--- a/tests/agent_unittests/test_import_hook.py
+++ b/tests/agent_unittests/test_import_hook.py
@@ -1,0 +1,43 @@
+from pytest import raises
+
+
+def test_import_hook_finder():
+    """
+    This asserts the behavior of ImportHookFinder.find_module. It behaves
+    differently depending on whether or not the module it is looking for
+    exists, has been registered with an import hook, and across different
+    python versions.
+    """
+    from newrelic.api.import_hook import ImportHookFinder, register_import_hook
+    import sys
+
+    PY2 = sys.version_info[0] == 2
+
+    finder = ImportHookFinder()
+
+    # a dummy hook just to be able to register hooks for modules
+    def hook(*args, **kwargs):
+        pass
+
+    # Finding a module that does not exist and is not registered returns None.
+    module = finder.find_module('some_module_that_does_not_exist')
+    assert module is None
+
+    # Finding a module that does not exist and is registered behaves
+    # differently on python 2 vs python 3.
+    register_import_hook('some_module_that_does_not_exist', hook)
+    if PY2:
+        with raises(ImportError):
+            module = finder.find_module('some_module_that_does_not_exist')
+    else:
+        module = finder.find_module('some_module_that_does_not_exist')
+        assert module is None
+
+    # Finding a module that exists, but is not registered returns None.
+    module = finder.find_module('newrelic')
+    assert module is None
+
+    # Finding a module that exists, and is registered, finds that module.
+    register_import_hook('newrelic', hook)
+    module = finder.find_module('newrelic')
+    assert module is not None


newrelic/api/import_hook.py
```diff
@@ -173,25 +168,16 @@ def find_module(self, fullname, path=None):
...
-
-                return _ImportHookLoader()
-
+            try:
+                from importlib.util import find_spec
+                return _ImportHookChainedLoader(find_spec(fullname).loader)
+            except ImportError:
+                try:
+                    from importlib import find_loader
+                    return _ImportHookChainedLoader(find_loader(fullname, path))
```
a-feld, line 177: This is a change in behavior, the previous implementation returned `None` for the case where `find_loader` returned `None`.

a-feld (changes requested):
Hi @sigvef , thank you for the PR! I ran some more extensive testing and came across a couple of issues that we should fix / discuss. Looking forward to getting this merged!

sigvef:
I've added a test case for find_module to help verify that we don't introduce any breaking changes here.

As it turns out, the newrelic python agent doesn't support python 3 below 3.5 at all, so we can completely remove the deprecated branch.

newrelic/api/import_hook.py
sigvef: Nice catch. This branch (the <=py3.4 fallback) is now completely removed, since the newrelic python agent doesn't run on <=py3.4, and is therefore never run anyway.

newrelic/api/import_hook.py
```diff
@@ -173,25 +168,16 @@ def find_module(self, fullname, path=None):
...
-                # If we get this far then the module we are
-                # interested in does actually exist and so return
-                # our loader to trigger import hooks and then return
-                # the module.
-
-                return _ImportHookLoader()
-
+            try:
+                from importlib.util import find_spec
+                return _ImportHookChainedLoader(find_spec(fullname).loader)
```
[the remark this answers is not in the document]
sigvef: True. I've added test coverage for this now, and handled the AttributeError case.

tests/agent_unittests/test_import_hook.py
```diff
@@ -0,0 +1,43 @@
...
+    def hook(*args, **kwargs):
+        pass
+
+    # Finding a module that does not exist and is not registered returns None.
+    module = finder.find_module('some_module_that_does_not_exist')
+    assert module is None
+
+    # Finding a module that does not exist and is registered behaves
+    # differently on python 2 vs python 3.
+    register_import_hook('some_module_that_does_not_exist', hook)
```
a-feld, line 28: Calling `register_import_hook` from within the test here has side effects for other tests. Is there a way we can exercise this code without introducing edits to the hooks that persist across tests?

newrelic/api/import_hook.py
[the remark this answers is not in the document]
a-feld: The try/except cases here may be catching a bit too much: I'm looking at the documentation for [find_spec](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec) and in 3.7+ it says it can raise a `ModuleNotFoundError` (subclass of `ImportError`), which would cause the `__import__(fullname)` code to run.

Looking at the implementation, I think the intent was to only run the "else" case when `from importlib.util import find_spec` raises an import error? If so, we should probably adjust the logic to use the `else` case of the try to run `return _ImportHookChainedLoader(find_spec(fullname).loader)`.

```python
try:
  from importlib.util import find_spec
except ImportError:
  __import__(fullname)
  return _ImportHookLoader()
else:
  spec = find_spec(fullname)
  if spec:
    return _ImportHookChainedLoader(spec.loader)
```

Thoughts on this approach? Should we protecting against the `ValueError` case? You've probably looked into more detail with `find_spec` than I have at this point, and I'm not sure what code would cause a `ValueError` to be raised. 😄

tests/agent_unittests/test_import_hook.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    This asserts the behavior of ImportHookFinder.find_module. It behaves
+    differently depending on whether or not the module it is looking for
+    exists, has been registered with an import hook, and across different
+    python versions.
+    """
+    from newrelic.api.import_hook import ImportHookFinder, register_import_hook
+    import sys
+
+    PY2 = sys.version_info[0] == 2
```
a-feld, line 14: (suggestion, this is entirely optional): you should be able to make use of our vendored version of `six` which provides this helper 😄 

```python
import newrelic.packages.six as six
```

tests/agent_unittests/test_import_hook.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_import_hook_finder():
+    """
+    This asserts the behavior of ImportHookFinder.find_module. It behaves
+    differently depending on whether or not the module it is looking for
+    exists, has been registered with an import hook, and across different
+    python versions.
+    """
+    from newrelic.api.import_hook import ImportHookFinder, register_import_hook
+    import sys
```
a-feld, line 12: (suggestion) These test local imports can probably be moved to the module scope (outside of the test)

a-feld (changes requested):
Thanks for your patience @sigvef ! We've been busy working on Starlette instrumentation 😄 
I have a few more questions / suggestions before we can get this merged. Thanks for your engagement, we're excited for this update!

tests/agent_unittests/test_import_hook.py
sigvef: Hm. Do you think clearing `_import_hooks` is enough?

sigvef:
https://github.com/newrelic/newrelic-python-agent/pull/26#discussion_r506561887

Narrowing down the exception scope is a good idea. I haven't run into the ValueError case in the wild, but the code should handle it, given that the documentation states that it can happen. I've updated the code accordingly. Do you think that approach works?

tests/agent_unittests/test_import_hook.py
sigvef: No, `_import_hooks` is already populated at the start of the test. Perhaps we would need to restructure a bit to allow for testing without side effects, then.

a-feld (approved):
It looks like the tests were failing with the changes that were pushed this morning. I discussed with the team on how to disposition this PR and we concluded that we should merge the PR in its earlier state (I force pushed the branch to the commit prior to this morning) and we'll own making further changes to this PR. Thanks for your contribution!

sigvef:
Sounds good! Thanks for getting this merged.

PR #33, opened 2020-11-24
Adding Caching to GHA Testing

TimPansino:

# Overview
* Adds caching of .tox directories and pip cache directories across test runs.
* Consolidates runners to have all versions of python installed.
* Runs all tox environments in parallel for each test directory.

# Related Github Issue
N/A

# Testing

This update affects all tests currently running.

Sidekick: @a-feld

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,7 +39,7 @@ jobs:
     - name: Get Date
       id: get-date
       run: |
-        echo "::set-output name=date::$(/bin/date -u "+%Y-%m")"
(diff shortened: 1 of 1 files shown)


a-feld (commented):
This looks great! Just some napkin math:
* We have ~60-ish test suites to port over
* The cache entry size according to github ranges from 100MB-200MB for our minimal pip-cache + tox dirs
* We can store at most 5GB

I think we'll probably be at or above the limit by the time we port our test suites. Is there any way to bring the cache size down?

Maybe excluding the site-packages directory while also saving the pip cache would result in a size reduction?

a-feld (approved):

PR #38, opened 2020-12-03
PYTHON-3827 Flask Tests

TimPansino:
# Overview

* Ports internal test suite to public repo.
* Reduces testing matrix by:
  * Removing C extension based testing.
  * Removing unnecessary version tests while retaining 100% coverage

# Related Github Issue

# Testing
exercise your changes.

### DIFF
--- a/tests/framework_flask/tox.ini
+++ b/tests/framework_flask/tox.ini
@@ -7,9 +7,6 @@ envlist =
     pypy-flask{0006,0007,0008,0009,0010,0011,0012,0100},
     pypy3-flask{0010,0012,0100,master},
 
-[jenkins]
-mostrecent = flask0012
-
 [pytest]
 usefixtures = session_initialization requires_data_collector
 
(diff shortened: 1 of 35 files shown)


tests/framework_flask/tox.ini
```diff
@@ -0,0 +1,31 @@
+[tox]
+setupdir = {toxinidir}/../..
+envlist =
+    {pypy,py27,py35,pypy3}-flask0012,
```
a-feld, line 4: Question: should we include `pypy3` in this test line? `pypy3` is 3.6+ compatible at this point in time, and moving to py37 compatibility. If flask only supports py35 at this version, wouldn't `pypy3` potentially cause issues?

a-feld (changes requested):

tests/framework_flask/tox.ini
TimPansino: Potentially, I'm ok with removing it. It's a very old version of flask anyway, so unlikely to have issues related to it.

a-feld (approved):

PR #66, opened 2020-12-14
psycopg2cffi Tests

TimPansino:
Sidekick: @carrieedwards 

# Overview

* Import psycopg2cffi tests from internal repo.
* Add namespacing to tests to fix concurrency issues.

### DIFF
--- a/tests/testing_support/fixtures.py
+++ b/tests/testing_support/fixtures.py
@@ -2055,6 +2055,22 @@ def validate_transaction_error_trace_count(num_errors):
     return _validate_transaction_error_trace_count
 
 
+def validate_transaction_slow_sql_count(num_slow_sql):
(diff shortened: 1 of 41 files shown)


tests/testing_support/fixtures.py
```diff
@@ -2055,6 +2055,22 @@ def _validate_transaction_error_trace_count(wrapped, instance, args,
     return _validate_transaction_error_trace_count
 
 
+def validate_transaction_slow_sql_count(num_slow_sql):
```
carrieedwards, line 2058: Is this validator one we should also put in its own file to clear up fixtures.py?

TimPansino: Yep, good call.

carrieedwards (approved):
Looks good to me!

carrieedwards (commented):
Looks good to me!

PR #68, opened 2020-12-15
pymongo Tests

TimPansino:
Sidekick: @a-feld 

# Overview

* Add job for mongodb tests.
* Import pymongo tests from internal repo.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -523,6 +523,101 @@ jobs:
         TOX_PARALLEL_NO_SPINNER: 1
         PY_COLORS: 0
 
+  mongodb:
+
+    strategy:
+      matrix:
+        test-directory:
+          - datastore_pymongo
+
+    runs-on: ubuntu-latest
+
+    services:
+      mongodb:
+        image: mongo:3.6.4
+        ports:
+          - 8080:27017
+          - 8081:27017
+        # Set health checks to wait until mongodb has started
+        options: >-
+          --health-cmd "echo 'db.runCommand(\"ping\").ok' | mongo localhost:27017/test --quiet || exit 1"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Get Date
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -523,6 +523,101 @@ jobs:
...
+    strategy:
+      matrix:
+        test-directory:
+          - datastore_pymongo
+
+    runs-on: ubuntu-latest
+
+    services:
+      mongodb:
+        image: mongo:3.6.4
```
a-feld, line 537: Why use mongo:3.6.4 instead of mongo:4 or mongo:4.4?

tests/datastore_pymongo/test_pymongo.py
```diff
@@ -0,0 +1,219 @@
...
+        db.my_collection.initialize_unordered_bulk_op()
+        db.my_collection.initialize_ordered_bulk_op()
+        db.my_collection.bulk_write([pymongo.InsertOne({'x': 1})])
+        db.my_collection.insert_one({'x': 300})
+        db.my_collection.insert_many([{'x': 1} for i in range(20, 25)])
+        db.my_collection.replace_one({'x': 1}, {'x': 2})
+        db.my_collection.update_one({'x': 1}, {'$inc': {'x': 3}})
+        db.my_collection.update_many({'x': 1}, {'$inc': {'x': 3}})
+        db.my_collection.delete_one({'x': 4})
+        db.my_collection.delete_many({'x': 4})
```
a-feld, line 43: It looks like there are destructive operations executing on the same DB. Is this safe for concurrent executions via tox -p auto?

TimPansino: I hadn't encountered any issues with concurrency but reading the docs it definitely looks like it could fail. I'll get it changed to be parameterized.

.github/workflows/tests.yml
TimPansino: `parallel_scan` was removed in mongo 4, and 3.6.4 is what we were using internally.

a-feld (approved):

PR #70, opened 2020-12-15
Add psycopg2 tests.

a-feld:
# Overview
Adds testing for psycopg2.

### DIFF
--- a/tests/datastore_psycopg2/test_obfuscation.py
+++ b/tests/datastore_psycopg2/test_obfuscation.py
@@ -3,10 +3,8 @@ import pytest
 from newrelic.api.background_task import background_task
 from newrelic.core.database_utils import SQLConnections
 
-from testing_support.validators.validate_database_node import (
-        validate_database_node)
-from testing_support.validators.validate_sql_obfuscation import (
-        validate_sql_obfuscation)
+from testing_support.validators.validate_database_node import validate_database_node
+from testing_support.validators.validate_sql_obfuscation import validate_sql_obfuscation
 from utils import DB_SETTINGS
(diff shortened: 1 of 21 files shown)


tests/datastore_psycopg2/test_obfuscation.py
```diff
@@ -0,0 +1,117 @@
...
+
+    connection = psycopg2.connect(
+            database=DB_SETTINGS['name'], user=DB_SETTINGS['user'],
+            password=DB_SETTINGS['password'], host=DB_SETTINGS['host'],
+            port=DB_SETTINGS['port'])
+
+    try:
+        cursor = connection.cursor()
+
+        cursor.execute('create table if not exists a (b text, c text)')
```
TimPansino, line 25:
```suggestion
        cursor.execute('create schema if not exists schema_%s' % str(os.getpid()))
        cursor.execute('set search_path to schema_%s' % str(os.getpid()))
        cursor.execute('drop table if exists a')
        cursor.execute('create table if not exists a (b text, c text)')
```

I'm seeing local failures with postgres having issues creating this table and getting duplicate key constraints, looks like the atomicity on DDL they define in their spec isn't flawless. Let's namespace this to a schema by PID using something like this.

TimPansino (changes requested):

tests/datastore_psycopg2/test_obfuscation.py
TimPansino: This should also be duplicated below to the test_explain_plans

TimPansino (approved):

PR #73, opened 2020-12-16
pysolr Tests

TimPansino:
Sidekick: @umaannamalai 

# Overview

* Import pysolr tests from internal repo.
* Upgrade tests with namespacing for concurrency issues.
* Add solr workflow to GHA

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -426,6 +426,103 @@ jobs:
         TOX_PARALLEL_NO_SPINNER: 1
         PY_COLORS: 0
 
+  solr:
+
+    strategy:
+      matrix:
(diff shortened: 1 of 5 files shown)


TimPansino:
Interesting notes to be aware of: 

* All our `solr` tests require a core (collection) to already exist.
* Github Actions does not support the `command` functionality of docker run, preventing our original usage of the official `solr` container on docker hub as it does not have any means to initialize a core at startup without a command.
* bitnami/solr provides this through an environment variable but is a much larger image size. It's apache licensed and available here https://github.com/bitnami/bitnami-docker-solr

umaannamalai (approved):

PR #78, opened 2020-12-16
Breakout Validators

TimPansino:
Sidekick: @carrieedwards 

# Overview

* Move previously added validators out of `fixtures.py` and `external_fixtures.py` to separate files.


carrieedwards (approved):
Looks good to me!

PR #86, opened 2020-12-17
feedparser Tests

TimPansino:
Sidekick: @carrieedwards 

# Overview

* Import feedparser tests from internal repo.
* Add testing for new major version of feedparser.


carrieedwards (approved):
Looks good to me!

PR #92, opened 2020-12-21
Unpin setup-python after bugfix

TimPansino:
Sidekick: @carrieedwards 

# Overview

* Unpin setup-python github action after bugfix.


carrieedwards (approved):
Looks good to me

PR #94, opened 2020-12-21
Parallel Coverage

TimPansino:
# Overview

* Implement minor changes to coverage to allow it to be run in parallel.
  * This allows us to run coverage on GHA should we choose to do so in the near future.


carrieedwards (approved):
Looks good to me!

PR #97, opened 2021-01-04
Add django ASGI instrumentation. #31

a-feld:
# Overview

This PR adds automatic transaction support for Django's ASGI application.

# Related Github Issue
#31

### DIFF
--- a/newrelic/hooks/adapter_asgiref.py
+++ b/newrelic/hooks/adapter_asgiref.py
@@ -10,15 +10,19 @@ class ContextOf(object):
     def __init__(self, trace_cache_id):
         self.trace_cache = trace_cache()
         self.trace = self.trace_cache._cache.get(trace_cache_id)
+        self.thread_id = None
+        self.restore = None
 
     def __enter__(self):
         if self.trace:
-            self.thread_id = self.trace_cache.thread_start(self.trace)
+            self.thread_id = self.trace_cache.current_thread_id()
+            self.restore = self.trace_cache._cache.get(self.thread_id)
+            self.trace_cache._cache[self.thread_id] = self.trace
         return self
 
     def __exit__(self, exc, value, tb):
-        if self.trace:
-            self.trace_cache.thread_stop(self.thread_id)
+        if self.restore:
+            self.trace_cache._cache[self.thread_id] = self.restore
 
 
 async def context_wrapper_async(awaitable, trace_cache_id):
--- a/newrelic/hooks/adapter_asgiref.py
+++ b/newrelic/hooks/adapter_asgiref.py
@@ -1,0 +1,44 @@
+from newrelic.common.object_wrapper import wrap_function_wrapper
+from newrelic.core.trace_cache import trace_cache
+
+
+def _bind_thread_handler(loop, source_task, *args, **kwargs):
+    return source_task
+
+
+class ContextOf(object):
+    def __init__(self, trace_cache_id):
+        self.trace_cache = trace_cache()
+        self.trace = self.trace_cache._cache.get(trace_cache_id)
+
+    def __enter__(self):
+        if self.trace:
+            self.thread_id = self.trace_cache.thread_start(self.trace)
+        return self
+
+    def __exit__(self, exc, value, tb):
+        if self.trace:
+            self.trace_cache.thread_stop(self.thread_id)
+
+
+async def context_wrapper_async(awaitable, trace_cache_id):
+    with ContextOf(trace_cache_id):
+        return await awaitable
+
+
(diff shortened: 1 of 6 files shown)


newrelic/hooks/framework_django.py
```diff
@@ -1222,3 +1235,13 @@ def instrument_django_core_handlers_exception(module):
         module.handle_uncaught_exception = (
                 wrap_handle_uncaught_exception(
                 module.handle_uncaught_exception))
+
+
+def instrument_django_core_handlers_asgi(module):
+    import django
+
+    framework = ('Django', django.get_version())
+
```
TimPansino, line 1244: Are these lines actually necessary? Looks like they're unused.

newrelic/hooks/framework_django_py3.py
```diff
@@ -0,0 +1,43 @@
...
+def _bind_get_response(request, *args, **kwargs):
+    return request
+
+
+async def _nr_wrapper_BaseHandler_get_response_async_(
+        wrapped, instance, args, kwargs):
+    response = await wrapped(*args, **kwargs)
+
+    if current_transaction() is None:
+        return response
```
TimPansino, line 17: Coverage is never true here.

newrelic/hooks/framework_django_py3.py
```diff
@@ -0,0 +1,43 @@
...
+
+    request = _bind_get_response(*args, **kwargs)
+
+    if hasattr(request, '_nr_exc_info'):
+        if not ignore_status_code(response.status_code):
+            record_exception(*request._nr_exc_info)
```
TimPansino, line 23: Coverage is never true here.

newrelic/hooks/framework_django_py3.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def _nr_wrap_converted_middleware_async_(middleware, name):
+
+    @function_wrapper
+    async def _wrapper(wrapped, instance, args, kwargs):
+        transaction = current_transaction()
+
+        if transaction is None:
+            return await wrapped(*args, **kwargs)
```
TimPansino, line 36: Coverage is never true here.

TimPansino (commented):
Left a few coverage related comments.

newrelic/hooks/framework_django.py
a-feld: They're used, an assertion is missing. Nice catch! 😄

newrelic/hooks/framework_django.py
```diff
@@ -458,8 +463,14 @@ def instrument_django_core_handlers_base(module):
...
-    wrap_function_wrapper(module, 'BaseHandler.get_response',
-            _nr_wrapper_BaseHandler_get_response_)
+    if six.PY3 and hasattr(module.BaseHandler, '_get_response') and hasattr(module.BaseHandler, '_get_response_async'):
+        wrap_function_wrapper(module, 'BaseHandler._get_response',
+                _nr_wrapper_BaseHandler_get_response_)
+        wrap_function_wrapper(module, 'BaseHandler._get_response_async',
+                _nr_wrapper_BaseHandler_get_response_async_)
+    else:
+        wrap_function_wrapper(module, 'BaseHandler.get_response',
+                _nr_wrapper_BaseHandler_get_response_)
```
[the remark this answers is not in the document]
a-feld: This logic is actually incorrect so I'm glad we took a second look at this. I believe the correct thing to do is to be wrapping `BaseHandler.get_response` and `BaseHandler.get_response_async` because the intent is to have an instrumentation point that is used just before the request is complete so that we can ignore errors based on status code.

newrelic/hooks/adapter_asgiref.py
```diff
@@ -0,0 +1,48 @@
...
+def main_wrap_wrapper(wrapped, instance, args, kwargs):
+    awaitable = wrapped(*args, **kwargs)
+    return context_wrapper_async(awaitable, trace_cache().current_thread_id())
+
+
+def instrument_asgiref_sync(module):
+    wrap_function_wrapper(module, 'SyncToAsync.thread_handler',
+        thread_handler_wrapper)
+    wrap_function_wrapper(module, 'AsyncToSync.main_wrap',
+        main_wrap_wrapper)
```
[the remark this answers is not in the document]
a-feld: I think it's ok to leave these unguarded so that in the future if these instrumentation points change and a user loads asgiref, an error will occur resulting in a log message getting printed. [The failure of the instrumentation point won't crash the application](https://github.com/newrelic/newrelic-python-agent/blob/f6a91024fc5a3cec571bb06eb65607538b7d316b/newrelic/config.py#L1128).

TimPansino (approved):

Issue #35, opened 2020-11-30
Crash on ASGI request with Referer header w/ non-ASCII characters


eykd:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

**Description**

We have a Starlette/uvicorn-based application that crashes w/ a traceback inside `newrelic` when it encounters a `Referer` header  w/ non-ASCII characters.

As best as I can tell from the ASGI spec and from uvicorn source, the headers in the ASGI scope are expected to be pairs of bytestrings, and `ASGIWebTransaction` captures them raw off the scope:

https://github.com/newrelic/newrelic-python-agent/blob/ddcc93c764d01d2e0082e4d2caeab11e796a2019/newrelic/api/asgi_application.py#L235

But later, in `WebTransaction._update_agent_attributes()`, the `referer` header is pulled directly from the raw headers:

https://github.com/newrelic/newrelic-python-agent/blob/ddcc93c764d01d2e0082e4d2caeab11e796a2019/newrelic/api/web_transaction.py#L326-L328

And then the raw bytestring is passed to `urllib.parse.urlsplit()` which tries to coerce it to `ascii`, when the expected encoding for headers is `latin-1`.

https://github.com/newrelic/newrelic-python-agent/blob/ddcc93c764d01d2e0082e4d2caeab11e796a2019/newrelic/api/web_transaction.py#L128-L130

Here's the traceback.

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 30: ordinal not in range(128)
  File "<string>", line 5, in wrapper
  File "starlette/exceptions.py", line 82, in __call__
    raise exc from None
  File "starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "starlette/routing.py", line 582, in __call__
    await route.handle(scope, receive, send)
  File "starlette/routing.py", line 243, in handle
    await self.app(scope, receive, send)
  File "starlette/routing.py", line 57, in app
    await response(scope, receive, send)
  File "starlette/responses.py", line 143, in __call__
    await send({"type": "http.response.body", "body": self.body})
  File "starlette/exceptions.py", line 68, in sender
    await send(message)
  File "starlette/middleware/errors.py", line 156, in _send
    await send(message)
  File "newrelic/api/asgi_application.py", line 140, in send_inject_browser_agent
    return await self.send(message)
  File "newrelic/api/asgi_application.py", line 264, in send
[...]

PR #114, opened 2021-01-13
Fix crash on non-ascii http referer. #35

a-feld:
Thanks to @eykd for the detailed bug report! ❤️ 

This PR fixes a crash that occurs when non-ascii HTTP referers are passed to ASGI applications. The crash is fixed by ensuring the query string value is a decoded string (rather than bytes) prior to calling urlsplit when removing the query string from the referer. A regression test is included.

Fixes: #35


carrieedwards (approved):
Looks good to me!

PR #124, opened 2021-01-19
Improve startup performance by deferring load of urllib3.

a-feld:
Loading urllib3 can be expensive since there are regex compiles at
import time.

urllib3 is not immediately needed for operations at initialize,
therefore this commit delays the loading of urllib3 and later only loads
those imports on a background thread.


TimPansino (approved):

TimPansino:
LGTM

PR #129, opened 2021-01-21
Fix event hook response processing regression

umaannamalai:

# Overview
This Pull Request changes the httpx event hook logic to use the external trace to process responses rather than the current trace.

# Related Github Issue

# Testing
exercise your changes.


a-feld (changes requested):
This needs a regression test. This error should be reproducible by setting enabled=false and verifying that the agent does not crash.

a-feld (approved):

PR #158, opened 2021-03-02
Refactor NewRelicContextFormatter, add log_record_to_dict

chrislawlor:

# Overview
Refactors the `api.log.NewRelicContextFormatter`, so that converting the log record to a dict, and serializing to JSON, are achievable separately.

## Reasoning

This change makes it easy to add the data exposed by `NewRelicContextFormatter` in custom formatters.

Without this change, a custom formatter has essentially two options:

1. Call `NewRelicContextFormatter.format`, deserialize the returned JSON, extend as desired, then re-serialize.
2. Copy the logic from `NewRelicContextFormatter`. This is undesirable for many reasons, not the least of which is that we can only assume that it will change over time, potentially breaking log integration when it does.

With this change in place, a custom formatter can call `NewRelicContextFormatter.log_record_to_dict` to get the correct values in dict format, extend as desired, then serialize.

# Related Github Issue

I haven't opened one as this is such a small change, it seems appropriate to discuss here rather than the issue tracker.

# Testing

No tests added as functionality is essentially unchanged. I can add a specific test for `log_record_to_dict` if desired.

exercise your changes.

### DIFF
--- a/newrelic/api/log.py
+++ b/newrelic/api/log.py
@@ -60,7 +60,8 @@ class NewRelicContextFormatter(Formatter):
     def __init__(self, *args, **kwargs):
         super(NewRelicContextFormatter, self).__init__()
 
-    def format(self, record):
+    @classmethod
+    def log_record_to_dict(cls, record):
         output = {
             "timestamp": int(record.created * 1000),
             "message": record.getMessage(),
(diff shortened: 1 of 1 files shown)


newrelic/api/log.py
```diff
@@ -88,4 +89,6 @@ def format(self, record):
 
         if record.exc_info:
             output.update(format_exc_info(record.exc_info))
-        return json.dumps(output)
+
+    def format(self, record):
+        return json.dumps(self.log_record_to_dict(record))
```
TimPansino, line 94:
```suggestion
        return output

    def format(self, record):
        return json.dumps(self.log_record_to_dict(record))
```

You're missing a return statement on `log_record_to_dict` 🙂 

Other than that, LGTM.

TimPansino (changes requested):

newrelic/api/log.py
chrislawlor: How embarrassing 😊 

Fixed.

TimPansino (approved):
LGTM. Thank you for the contribution!

jarcoal:
Any chance we could get a new release cut for this?

TimPansino:
Sorry about the long wait, we should have cut a patch release for this. We're in the middle of a feature that we expect to release this week, this PR will be included in that release.

PR #186, opened 2021-05-06
Add expected errors settings

umaannamalai:
This PR adds new settings for expected error status codes and classes.

### DIFF
--- a/newrelic/config.py
+++ b/newrelic/config.py
@@ -189,12 +189,12 @@ def _map_console_listener_socket(s):
 
 
 def _merge_ignore_status_codes(s):
-    return newrelic.core.config._parse_error_status_codes(
+    return newrelic.core.config._parse_status_codes(
         s, _settings.error_collector.ignore_status_codes
     )
 
 def _merge_expected_status_codes(s):
-    return newrelic.core.config._parse_error_status_codes(
+    return newrelic.core.config._parse_status_codes(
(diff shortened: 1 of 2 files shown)


newrelic/config.py
```diff
@@ -189,10 +189,15 @@ def _map_console_listener_socket(s):
 
 
 def _merge_ignore_status_codes(s):
-    return newrelic.core.config._parse_ignore_status_codes(
+    return newrelic.core.config._parse_error_status_codes(
```
TimPansino, line 192: If we're changing the name of this, can we just make it `_parse_status_codes`? There's really nothing error specific about it. The name bothers me a bit.

newrelic/config.py
```diff
@@ -1850,15 +1864,19 @@ def _process_error_trace_configuration():
             (module, object_path) = function.split(":", 1)
 
             ignore_errors = []
+            expected_classes = []
 
             if _config_object.has_option(section, "ignore_errors"):
                 ignore_errors = _config_object.get(section, "ignore_errors").split()
```
TimPansino, line 1870: I assume then that we're renaming this setting in a future PR with all the associated rip out?

umaannamalai: Correct. This will get renamed in the work associated with deprecating the `ignore_errors` setting.

TimPansino (approved):

PR #212, opened 2021-05-12
Update error trace with status code and expected errors

umaannamalai:
This PR adds status code and expected error class parameters to the error trace wrapper.

Closes #212

### DIFF
--- a/newrelic/api/error_trace.py
+++ b/newrelic/api/error_trace.py
@@ -14,24 +14,25 @@
 
 import functools
 
-from newrelic.api.time_trace import current_trace, record_exception
+from newrelic.api.time_trace import current_trace, notice_error
 from newrelic.common.object_wrapper import FunctionWrapper, wrap_object
 
 
 class ErrorTrace(object):
-
-    def __init__(self, ignore_errors=[], **kwargs):
+    def __init__(self, ignore=None, expected=None, status_code=None, **kwargs):
         parent = None
         if kwargs:
             if len(kwargs) > 1:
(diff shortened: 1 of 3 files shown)


newrelic/api/error_trace.py
```diff
@@ -14,24 +14,25 @@
...
 
-from newrelic.api.time_trace import current_trace, record_exception
+from newrelic.api.time_trace import current_trace, notice_error
 from newrelic.common.object_wrapper import FunctionWrapper, wrap_object
 
 
 class ErrorTrace(object):
-
-    def __init__(self, ignore_errors=[], **kwargs):
+    def __init__(self, ignore=None, expected=None, status_code=None, **kwargs):
```
TimPansino, line 22: Was thinking about this API, can we keep `ignore_errors` in addition to the others?

We should add a deprecation warning to the ignore_errors argument and forward the value to ignore_classes if ignore_classes is Falsey, or we could merge them if they're both iterables. Would be less likely to break existing customer code bases.

newrelic/api/error_trace.py
```diff
@@ -13,25 +13,38 @@
...
         if parent is None:
             parent = current_trace()
 
         self._transaction = parent and parent.transaction
         self._ignore_errors = ignore_errors
+        self._ignore = ignore if ignore is not None else ignore_errors
+        self._expected = expected
+        self._status_code = status_code
+
+        if self._ignore_errors:
```
TimPansino, line 40: Can change this to a `if ignore_errors` and drop the line above where we store it since this is never accessed again.

TimPansino (approved):

PR #213, opened 2021-05-12
Add ignore/expected status codes API to notice_error

### LINKED ISSUE #202: Implement expected_status_codes, and rip out old ignore_status_codes logic.
* make status_code a callable in expected/ignored rule matching
* change api of expected/ignored rule matching to take in exc_info instead of string names
* encapsulate logic for extracting string info from exceptions and reuse in both record_exception, notice_error, and rule matching
* Coerce status_code into an int

Something like 
```python
    if callable(status_code):
        status_code = status_code(exc, value, tb)

    if status_code is not None:
        try:
            status_code = int(status_code)
            if status_code in status_codes_rules:
                return True
        except:
            # Warning here
            pass
```

TimPansino:

# Overview
* Adds support for ignore/expected status_codes within notice_error.
* Encapsulate parsing of message, and names from exceptions for reusability.
* Add testing for status_codes functionality.
* Add testing for mixed ignore and expected settings with overrides.
* Expand test file to include all current cases.

# Related Github Issue
Closes #202

# Testing
exercise your changes.

### DIFF
--- a/newrelic/core/config.py
+++ b/newrelic/core/config.py
@@ -1083,50 +1083,36 @@ def _remove_ignored_configs(server_settings):
 
 
 def ignore_status_code(status):
+    """Legacy function kept here for compatibility."""
     return status in _settings.error_collector.ignore_status_codes
 
 
 def is_expected_error(
-    module=None,
-    name=None,
-    message=None,
+    exc_info,
     status_code=None,
-    fullname=None,
 ):
     return error_matches_rules(
         "expected",
-        module=module,
-        name=name,
-        message=message,
+        exc_info,
         status_code=status_code,
-        fullname=fullname,
     )
 
 
 def should_ignore_error(
-    module=None,
-    name=None,
-    message=None,
+    exc_info,
     status_code=None,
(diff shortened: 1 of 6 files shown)


newrelic/core/config.py
```diff
@@ -1145,29 +1132,34 @@ def error_matches_rules(
     if not settings:
         return False
 
-    # TODO Remove default None after settings are implemented
-    classes_rules = getattr(settings.error_collector, "%s_classes" % rules_prefix, set())
-    status_codes_rules = getattr(
-        settings.error_collector, "%s_status_codes" % rules_prefix, set()
-    )
+    # Retrieve settings based on prefix
+    classes_rules = getattr(settings.error_collector, "%s_classes" % rules_prefix)
```
umaannamalai, line 1136: Do we still want a default here to avoid any potential AttributeErrors?

TimPansino: Yeah that's fair, I'll put that back in

umaannamalai (approved):

PR #221, opened 2021-05-19
Add metric logic for expected errors.

### LINKED ISSUE #190: Add logic to avoid incrementing error metrics for expected errors

umaannamalai:
Closed #190

### DIFF
--- a/newrelic/core/stats_engine.py
+++ b/newrelic/core/stats_engine.py
@@ -734,7 +734,7 @@ class StatsEngine(object):
         # Regardless of whether we record the trace or the event we still
         # want to increment the metric Errors/all unless the error was marked
         # as expected
-        if self._error_event(error_details)[0].get("error.expected"):
+        if is_expected:
             self.record_time_metric(TimeMetric(name='ErrorsExpected/all', scope='',
                 duration=0.0, exclusive=None))
         else:
--- a/newrelic/core/stats_engine.py
+++ b/newrelic/core/stats_engine.py
@@ -732,8 +732,13 @@ class StatsEngine(object):
             self.__transaction_errors.append(error_details)
 
         # Regardless of whether we record the trace or the event we still
-        # want to increment the metric Errors/all
-        self.record_time_metric(TimeMetric(name='Errors/all', scope='',
+        # want to increment the metric Errors/all unless the error was marked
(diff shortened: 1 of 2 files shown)


newrelic/core/transaction_node.py
```diff
@@ -261,7 +268,8 @@ def apdex_metrics(self, stats):
         frustrating = 0
 
         if self.errors:
-            frustrating = 1
+            if False in (error.expected for error in self.errors):
```
TimPansino, line 271: Let's flatten this to `if self.errors and False in (error.expected for error in self.errors):`. Otherwise the else statement will be missed when an expected error occurs.

newrelic/core/transaction_node.py
```diff
@@ -398,7 +406,8 @@ def apdex_perf_zone(self):
             return None
 
         if self.errors:
-            return 'F'
+            if False in (error.expected for error in self.errors):
```
TimPansino, line 409: And then also flatten here.

newrelic/core/stats_engine.py
```diff
@@ -732,8 +732,13 @@ def notice_error(self, error=None, attributes={}, expected=None, ignore=None, st
             self.__transaction_errors.append(error_details)
 
         # Regardless of whether we record the trace or the event we still
-        # want to increment the metric Errors/all
-        self.record_time_metric(TimeMetric(name='Errors/all', scope='',
+        # want to increment the metric Errors/all unless the error was marked
+        # as expected
+        if self._error_event(error_details)[0].get("error.expected"):
```
TimPansino, line 737: We already have `is_expected` available in this function, we can just reduce this to `if is_expected:`

TimPansino (changes requested):

newrelic/core/transaction_node.py
umaannamalai: That's the way I actually had it originally but I started second guessing the logic in the case an error was expected. I'll switch it back!

newrelic/core/stats_engine.py
umaannamalai: Ah good call. I had written this before that functionality was added to notice_error

TimPansino (approved):

PR #230, opened 2021-05-28
Add non PID specific total memory metric.

umaannamalai:
Currently, the agent yields up total memory metrics for each detected PID. This PR adds a non PID specific total memory metric for compatibility with the query being used in the NR UI for the _All hosts by memory usage_ graph (found in the Charts section of the Summary page).

### DIFF
--- a/newrelic/samplers/memory_usage.py
+++ b/newrelic/samplers/memory_usage.py
@@ -25,8 +25,16 @@ from newrelic.samplers.decorators import data_source_generator
 PID = os.getpid()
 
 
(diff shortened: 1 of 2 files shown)


TimPansino:
Should we not also yield up a `Memory/Physical/Utilization` metric?

umaannamalai:
> Should we not also yield up a `Memory/Physical/Utilization` metric?

I'm not opposed to doing this. I'll add this in.

TimPansino (approved):

Issue #236, opened 2021-06-02
Pyramid (transaction naming): exception views implicitly take precedence


Sybret:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

**Description**
When a [custom exception view](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/views.html#custom-exception-views) is triggered, the final [view_handler_wrapper](https://github.com/newrelic/newrelic-python-agent/blob/13508df5a2945f1e47f72eac35beb577f3166475/newrelic/hooks/framework_pyramid.py#L88) call sets the current transaction's name to be the exception view. This final call overrides the original view that served the request and results in a loss of context within the APM transaction view.

**Expected Behavior**
The transaction name would be set to the original view which served the request.

**Proposed Fix**
Use a higher priority=2 when calling set_transaction_name
* in the view_handler_wrapper (for non-exception views only)
* for both calls in the default_view_mapper_wrapper._wrapper since we have an instance of the registered view_callable in both cases

similar to how exception handler prioritization is handled in the Flask hooks module.

**Steps to Reproduce**

1. Register a single route and view to be called that raises Exception() in the view function body
2. Register an exception view with that context `config.add_view(context=Exception, view=...)`
3. Run the pyramid app using the `newrelic-admin run-program` wrapper
4. Curl the route registered in Step 1
5. View the transaction in the NewRelic APM -> Transactions Web UI

**Your Environment**
Newrelic versions tested: 6.4.0.157 and the current main branch

**Additional context**
[TIP]:  # ( Add any other context about the problem here. For example, relevant community posts or support tickets. )

PR #237, opened 2021-06-02
Pyramid (transaction naming): lower exception view precedence

Sybret:
# Overview
For the Pyramid framework, use a higher priority=2 when calling set_transaction_name
* in the view_handler_wrapper (for non-exception views only)
* for both calls in the default_view_mapper_wrapper._wrapper since we have an instance of the registered view_callable in both cases

similar to how exception handler prioritization is handled in the Flask hooks module.

# Related Github Issue
https://github.com/newrelic/newrelic-python-agent/issues/236

# Testing
exercise your changes.

TODO:
- [ ] add unit tests

### DIFF
--- a/newrelic/hooks/framework_pyramid.py
+++ b/newrelic/hooks/framework_pyramid.py
@@ -98,7 +98,11 @@ def view_handler_wrapper(wrapped, instance, args, kwargs):
 
     name = callable_name(view_callable)
 
-    transaction.set_transaction_name(name)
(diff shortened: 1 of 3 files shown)


kat-star:
Hi @Sybret. Thank you for your contribution. We like the change you submitted, and we plan to incorporate it in our next release.

Sybret:
Excellent, thanks @kat-star @TimPansino !

kat-star (approved):
LGTM

PR #272, opened 2021-07-06
 Add field resolver metrics and span attributes.

umaannamalai:
This PR adds field resolver metrics and span attributes.
Note: Some test metrics named after field resolvers are commented out since resolver name in those cases will be replaced with deepest path in a future PR.

# Related Github Issue
#263

### DIFF
--- a/tests/framework_graphql/test_application.py
+++ b/tests/framework_graphql/test_application.py
@@ -13,16 +13,25 @@
 # limitations under the License.
 
 import pytest
-from testing_support.fixtures import validate_transaction_metrics, validate_transaction_errors
+from testing_support.fixtures import (
+    validate_transaction_metrics,
+    validate_transaction_errors,
+    override_application_settings,
+    dt_enabled,
+)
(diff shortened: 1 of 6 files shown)


tests/framework_graphql/test_application.py
```diff
@@ -122,3 +126,25 @@ def _test():
...
+
+    @validate_transaction_metrics(
+         "_target_application:resolve_hello",
+         rollup_metrics=field_resolver_metrics,
+         scoped_metrics=field_resolver_metrics,
+         background_task=True,
+     )
+    @validate_span_events(exact_agents={'graphql.field.name': 'hello'}, count=1)
+    @validate_span_events(exact_agents={'graphql.field.parentType': 'Query'}, count=1)
+    @validate_span_events(exact_agents={'graphql.field.path': 'hello'}, count=1)
```
TimPansino, line 143: You can collapse these into a single validator, no need to duplicate it

TimPansino (approved):

PR #303, opened 2021-07-29
Add GraphQL instrumentation support for Graphene.

umaannamalai:
This PR adds initial GraphQL instrumentation to the Python Agent. With this PR, the Graphene library and Graphene schemas created in Starlette/ FastAPI (run with an AsyncioExecutor) will be auto instrumented.


TimPansino (approved):
🚀

PR #321, opened 2021-08-17
Update Default Agent Logging

umaannamalai:
This PR updates the generated newrelic.ini to output logging to stderr by default. This update will allow users to receive immediate feedback when running the agent and customize logging afterward as needed.

### DIFF
--- a/newrelic/newrelic.ini
+++ b/newrelic/newrelic.ini
@@ -49,7 +49,7 @@ monitor_mode = true
 # possible to say "stderr" and output to standard error output.
 # This is set by default would normally result in output appearing
 # in your web server log.
-log_file = stderr
+log_file = stdout
 
 # Sets the level of detail of messages sent to the log file, if
 # a log file location has been provided. Possible values, in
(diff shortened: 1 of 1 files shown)


newrelic/newrelic.ini
```diff
@@ -41,18 +41,15 @@ app_name = Python Application
...
-#log_file = /tmp/newrelic-python-agent.log
+# Sets the name of a file to log agent messages to. Whatever you
+# set this to, you must ensure that the permissions for the
+# containing directory and the file itself are correct, and
+# that the user that your web application runs as can write out
+# to the file. If not able to out a log file, it is also
+# possible to say "stderr" and output to standard error output.
+# This is set by default would normally result in output appearing
+# in your web server log.
+log_file = stderr
```
TimPansino, line 52: Should we not use stdout as the default? If you're going to set one that's what I would set.

umaannamalai: I'm good with that

lrafeei (approved):

PR #335, opened 2021-08-30
CAT/DT--DT by default

lrafeei:
# Overview
DT by default, CAT disabled, fix tests that fail due to assuming otherwise

# Related Github Issue
https://github.com/newrelic/newrelic-python-agent/issues/318


TimPansino (approved):
🚀

PR #342, opened 2021-09-03
Enable DT in default newrelic.ini

umaannamalai:
This updates the default newrelic.ini to set distributed_tracing.enabled to True as a part of the DT by default effort.


TimPansino (approved):
🚀

PR #383, opened 2021-09-24
Fix notice_error Rules Matching Settings Handling

### LINKED ISSUE #378: Ignore Errors should not read global settings

TimPansino:
# Overview
* Fix handling of settings in `error_matches_rules` to use the same settings as the call to `notice_error`, to be sure application settings are properly observed.
* Reorder `complete_root` and `notice_error` calls in `Transaction.__exit__` to fix missing settings issues.
* Remove `error.expected` attribute from logs in context IF there is no active transaction, as there is no reliable way to retrieve an application from within the context formatter otherwise.

# Related Github Issue
Closes #378

# Testing
* Modified existing tests to ensure they failed when these changes were not present.
* Added new tests for logs in context to capture inside/outside transaction behavior of errors

### DIFF
--- a/newrelic/core/config.py
+++ b/newrelic/core/config.py
@@ -41,8 +41,8 @@ except ImportError:
 try:
     import grpc
 
-    from newrelic.core.infinite_tracing_pb2 import (
-        Span as _,  # NOQA # pylint: disable=W0611,C0412
+    from newrelic.core.infinite_tracing_pb2 import (  # pylint: disable=W0611,C0412  # noqa: F401
+        Span,
     )
 except ImportError:
     grpc = None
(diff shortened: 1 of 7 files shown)


newrelic/core/config.py
```diff
@@ -1103,53 +1102,71 @@ def ignore_status_code(status):
...
+
+        if not settings:
+            # Pull from active trace if no settings on transaction
+            trace = tc.current_trace()
+            settings = trace and trace.settings
+
+            if not settings:
+                # Unable to find rules to match with
+                _logger.error(
+                    "Failed to retrieve exception rules: No settings supplied, or found on transaction or trace."
```
umaannamalai, line 1161: Should we fallback to `global_settings()` if none are found on the txn or trace?

TimPansino: I wouldn't, I don't want any chances of a bug where serverside overrides are ignored.

lrafeei (approved):

PR #443, opened 2021-12-21
Add support for Redis v4.

umaannamalai:
This PR adds new instrumentation to support Redis v4. It also updates the tox.ini to test on both v3 and v4 of Redis.

### DIFF
--- a/tests/datastore_redis/test_instance_info.py
+++ b/tests/datastore_redis/test_instance_info.py
@@ -91,7 +91,10 @@ _instance_info_from_url_tests = [
     (('redis://localhost:1234/garbage',), {}, ('localhost', '1234', '0')),
 ]
 
-if REDIS_PY_VERSION >= (2, 7, 5):
+# Behavior to default port to 6379 was added in v2.7.5
+# (https://github.com/redis/redis-py/commit/cfe1041bbb8a8887531810429879bffbe705b03a)
+# and removed in v4.0.0b1 (https://github.com/redis/redis-py/blame/v4.0.0b1/redis/connection.py#L997)
+if (3, 5, 3) >= REDIS_PY_VERSION >= (2, 7, 5):
     _instance_info_from_url_tests.append(
         (('redis://127.0.0.1',), {}, ('127.0.0.1', '6379', '0'))
     )
--- a/tox.ini
+++ b/tox.ini
@@ -81,7 +81,8 @@ envlist =
     mongodb-datastore_pymongo-{py27,py36,py37,py38,py39,py310,pypy}-pymongo{03},
     mysql-datastore_pymysql-{py27,py36,py37,py38,py39,py310,pypy,pypy3},
     solr-datastore_pysolr-{py27,py36,py37,py38,py39,py310,pypy,pypy3},
-    redis-datastore_redis-{py27,py36,py37,py38,py39,py310,pypy,pypy3},
+    redis-datastore_redis-{py27,py36,py37,py38,pypy,pypy3}-redis03,
+    redis-datastore_redis-{py36,py37,py38,py39,py310,pypy3}-redislatest,
     solr-datastore_solrpy-{py27,pypy}-solrpy{00,01},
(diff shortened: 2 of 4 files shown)


tests/datastore_redis/test_instance_info.py
```diff
@@ -45,95 +44,96 @@ class DisabledSSLConnection(redis.SSLConnection, DisabledConnection):
...
 ]
 
-if REDIS_PY_VERSION >= (2, 7, 5):
-    _instance_info_from_url_tests.append(
-        (('redis://127.0.0.1',), {}, ('127.0.0.1', '6379', '0'))
-    )
+# Behavior to default port to 6379 was added in v2.7.5
+# (https://github.com/redis/redis-py/commit/cfe1041bbb8a8887531810429879bffbe705b03a)
+# and removed in v4.0.0b1 (https://github.com/redis/redis-py/blame/v4.0.0b1/redis/connection.py#L997)
+if (3, 5, 3) >= REDIS_PY_VERSION >= (2, 7, 5):
```
TimPansino, line 100: I don't think this is how this operator works. `(3, 5, 3) >= REDIS_PY_VERSION` evaluates to True or False, which gets coerced to 1 or 0 before being compared to `>= (2, 7, 5)`. So the full expression will always be False.

```suggestion
if (3, 5, 3) >= REDIS_PY_VERSION and REDIS_PY_VERSION >= (2, 7, 5):
```

tox.ini
```diff
@@ -208,7 +209,8 @@ deps =
     datastore_pymongo-pymongo03: pymongo<4.0
     datastore_pymysql: PyMySQL<0.11
     datastore_pysolr: pysolr<4.0
-    datastore_redis: redis
+    datastore_redis-redislatest: redis==4.0.2
```
TimPansino, line 212: This should be unpinned

```suggestion
    datastore_redis-redislatest: redis
```

tests/datastore_redis/test_instance_info.py
TimPansino: Ironically I just found this in another codebase, this might actually work. I'm not quite sure why it works, I didn't expect it to.

TimPansino: Fascinating. It's actually baked into the syntax of the language. I retract that suggestion.

https://docs.python.org/3/reference/expressions.html#comparisons

tox.ini
```diff
@@ -210,8 +211,10 @@ deps =
     datastore_pymongo-pymongo03: pymongo<4.0
     datastore_pymysql: PyMySQL<0.11
     datastore_pysolr: pysolr<4.0
-    datastore_redis: redis
-    datastore_redis: rb
+    datastore_redis-redislatest: redis
+    datastore_redis-redis040: redis<4.1
```
TimPansino, line 215: Let's keep 2 digit number for all versions so this doesn't get confusing.

```suggestion
    datastore_redis-redis0400: redis<4.1
```

tox.ini
```diff
@@ -82,7 +82,8 @@ envlist =
     mongodb-datastore_pymongo-{py27,py36,py37,py38,py39,py310,pypy}-pymongo{03},
     mysql-datastore_pymysql-{py27,py36,py37,py38,py39,py310,pypy,pypy3},
     solr-datastore_pysolr-{py27,py36,py37,py38,py39,py310,pypy,pypy3},
-    redis-datastore_redis-{py27,py36,py37,py38,py39,py310,pypy,pypy3},
+    redis-datastore_redis-{py27,py36,py37,py38,pypy,pypy3}-redis03,
+    redis-datastore_redis-{py36,py37,py38,py39,py310,pypy3}-redis{040,latest},
```
TimPansino, line 86: Let's keep 2 digit number for all versions so this doesn't get confusing.

```suggestion
    redis-datastore_redis-{py36,py37,py38,py39,py310,pypy3}-redis{0400,latest},
```

TimPansino (approved):

TimPansino:
Testing successful with redis versions 4,5,6.0, and 6.2.

PR #477, opened 2022-02-07
Update newrelic.ini

DmytroLitvinov:
# Overview
Change protocol from HTTP to HTTPS (found it when integrated `check no http urls` pre-commit hooks )

### DIFF
--- a/newrelic/newrelic.ini
+++ b/newrelic/newrelic.ini
@@ -10,7 +10,7 @@
 # find for Microsoft Windows INI files. For further information on the
 # configuration file format see the Python ConfigParser documentation at:
 #
-#    http://docs.python.org/library/configparser.html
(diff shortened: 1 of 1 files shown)


TimPansino (approved):

TimPansino:
Thanks for the fix, we'll get that merged into the next release.

Ignore the random test failure, we'll fix that elsewhere.

DmytroLitvinov:
Hi @TimPansino ,
Should I sign CLA? Have not done it yet.

TimPansino:
> Hi @TimPansino , Should I sign CLA? Have not done it yet.

That would be preferable, I don't think there's a strict legal requirement for changing 1 letter of a URL but I'm sure legal would prefer if you did.

DmytroLitvinov:
Signed CLA :+1:

PR #489, opened 2022-02-25
Add aredis testing.

umaannamalai:
This PR adds testing for the new aredis instrumentation.


TimPansino (approved):
LGTM. Manually tested on 1.0.0 as well as with hiredis extra, no code path changes between these versions. I'm good with keeping just the latest in tox.

PR #555, opened 2022-05-26
Record log event

lrafeei:
Co-authored-by: Lalleh Rafeei lrafeei@users.noreply.github.com
Co-authored-by: Uma Annamalai umaannamalai@users.noreply.github.com
Co-authored-by: Tim Pansino timpansino@users.noreply.github.com


newrelic/api/transaction.py
```diff
@@ -1821,6 +1850,34 @@ def record_custom_event(event_type, params, application=None):
...
+        transaction = current_transaction()
+        if transaction:
+            transaction.record_log_event(message, level, timestamp)
+        else:
+            application = application_instance(activate=False)
+
+            if application and application.enabled:
+                application.record_log_event(message, level, timestamp, priority=priority)
+            else:
+                _logger.debug(
```
umaannamalai, line 1871: Should this debug message be edited? Since record_log_event is not currently a public API that customers can use, they will not be able to supply their own application object to the call.

TimPansino: I would leave it in, people will almost certainly use the API whether we make it officially public or not. Either way, it's a debug message. It might help someone or us in the future to leave it like that.

PR #569, opened 2022-06-15
ARedis Concurrency Bug Regression Test

TimPansino:
# Overview
* Add a regression test that reproduces probably concurrency bugs in async redis instance reporting.

# Related Github Issue
N/A

# Testing
* This test is marked XFail to prove the existence of the bug and demonstrate a working regression test. The fix will come in a separate PR which removes the XFail marker.


TimPansino:
Note to run this test locally you'll need to open a redis instance on ports 8080 and 8081, and set GITHUB_ACTIONS=true in your env vars.

umaannamalai (approved):

PR #582, opened 2022-07-11
Update data usage supportability metrics.

umaannamalai:


tests/agent_unittests/test_http_client.py
```diff
@@ -289,32 +289,46 @@ def test_http_payload_compression(server, client_cls, method, threshold):
         compression_threshold=threshold,
     ) as client:
         with InternalTraceContext(internal_metrics):
-            status, data = client.send_request(payload=payload, params={"method": "test"})
+            status, data = client.send_request(payload=payload, params={"method": "method1"})
+
+    with client_cls(
```
hmstepanek, line 294: I'm admittedly having a really hard time following this test which is nothing against what you've done here-it was that way before you touched it in this PR. Here are my thoughts on it as I currently understand it (it might be better to have you walk me through it over zoom to make sure my understanding is correct here):
* All the if-else statements in this test make it really hard to follow. My general preference is if I find myself wanting to add an if-else statement I break the test into two or more separate tests. That being said, you didn't originally write this test so I certainly am not saying you need to fix it as part of this PR.
* I like that you changed the method name from test to `method1`. I think that makes it easier to follow.
* I'm not sure why we added a second call to method2 here. Could you explain your thinking there?
* I'm not sure why we only check method2 for byte compression and not method1 in the `threshold < 20` statement. Could we just change that to method1 and eliminate the second call to method2 all-together?

newrelic/common/agent_http.py
```diff
@@ -524,24 +524,31 @@ def _supportability_request(params, payload, body, compression_time):
...
         # *********
 
-        if agent_method and body:
+        if agent_method and payload:
             # Compression was applied
             if compression_time is not None:
                 internal_metric(
-                    "Supportability/Python/Collector/ZLIB/Bytes/%s" % agent_method,
-                    len(payload),
+                    "Supportability/Python/Collector/%s/ZLIB/Bytes" % agent_method,
```
hmstepanek, line 536: It might be better to add this in a commit message but for my own sake and the sake of any future developer reading this, I'm going to explain what's going on here as it wasn't immediately clear just reading the code:

We previously were not following the agent spec in a couple of ways which this code now fixes:
1. The method name should come before the /ZLIB/Bytes path: `Supportability/Python/Collector/ZLIB/Bytes/%s` -> `Supportability/Python/Collector/%s/ZLIB/Bytes`
1. The spec wants a count independent of the method: `Supportability/Python/Collector/ZLIB/Bytes`.

hmstepanek (commented):
Good catch on the bug with the body and payload being reversed! In general looks good. I had some questions about the testing which might be easier to talk over zoom about rather than on the PR so lmk what you want to do here!

hmstepanek (approved):

PR #583, opened 2022-07-19
Enable log forwarding by default.

umaannamalai:
This PR flips the switch on `application_logging.forwarding.enabled` to turn it on by default.


hmstepanek (changes requested):
Super excited to see this on by default! Just a minor change here, otherwise looks good. I see there are some other lint failures here that aren't related to the code changed in this PR. Personally, I'm totally fine with not making those changes as they are unrelated to the code that was changed but wasn't sure what our team policy was on that?

hmstepanek (approved):
Looks good!

hmstepanek (approved):

PR #587, opened 2022-07-21
Enable log forwarding by default. 

umaannamalai:
This PR enables log forwarding by default and corrects the Python agent data usage supportability metric naming scheme. It also addresses a bug related to compressed byte calculation.


hmstepanek (approved):
Looks good.

lrafeei (approved):
🚀

PR #589, opened 2022-07-22
Remove unknown/useless pylint disable options

hmstepanek:
Fix the following errors we've been seeing in CI:
```
pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'B101' (unknown-option-value)
pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'E122' (unknown-option-value)
pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'E126' (unknown-option-value)
pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'E127' (unknown-option-value)
pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'E128' (unknown-option-value)
pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'E203' (unknown-option-value)
pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'E501' (unknown-option-value)
pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'E722' (unknown-option-value)
pyproject.toml:1:0: W0012: Unknown option value for '--disable', expected a valid pylint message and got 'W504' (unknown-option-value)
```


pyproject.toml
```diff
@@ -6,7 +6,7 @@ include = '\.pyi?$'
 profile = "black"
 
 [tool.pylint.messages_control]
-disable = "B101,C0103,C0114,C0115,C0116,C0209,C0302,C0415,E0401,E1120,E122,E126,E127,E128,E203,E501,E722,R0201,R0205,R0401,R0801,R0902,R0903,R0904,R0911,R0912,R0913,R0914,R0915,R1705,R1710,R1725,W0201,W0212,W0223,W0402,W0603,W0612,W0613,W0702,W0703,W0706,W504,line-too-long,redefined-outer-name"
```
umaannamalai, line 9: Do you think we should keep R0201? This one checks if a method could actually be written as a function.

hmstepanek: R0201 has moved to [this](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-self-use.html) new rule which is dependent on an extension. I do see that we make use of the staticmethod decorator a lot in the code already which is basically what this check would have us do on any code that violates the rule. I don't really have a preference as long as enabling this extension doesn't require us to fix a bunch of violations.

umaannamalai (approved):
🚀

pyproject.toml
umaannamalai: I think we're fine leaving this as is!

hmstepanek: For what it's worth, I think that's the right call. I just ran it locally and there are quite a few violations on this rule in the code base already.

PR #598, opened 2022-08-03
Hypercorn ASGI Server Instrumentation

TimPansino:
# Overview
* Add instrumentation for Hypercorn ASGI server, a project by Pallets (maintainers of Flask) which split from Quart. 
* Supports both ASGI 2/3
* Supports WSGI on v0.14+
* Support included for trio, asyncio, and uvloop workers.
* Testing included for all current minor versions v0.10+.


tests/adapter_hypercorn/test_hypercorn.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+@override_application_settings({"transaction_name.naming_scheme": "framework"})
+def test_hypercorn_500(port, app):
+    @validate_transaction_errors(["builtins:ValueError"])
+    @validate_transaction_metrics(callable_name(app))
+    @raise_background_exceptions()
+    @wait_for_background_threads()
+    def _test():
+        try:
```
hmstepanek, line 120: Here you can take advantage of pytest's raises method:
```
        with pytest.raises(HTTPError):
            urlopen("http://localhost:%d/exc" % port)
```

tests/testing_support/sample_asgi_applications.py
```diff
@@ -12,27 +12,36 @@
...
 
     async def __call__(self, receive, send):
+        if self.scope["type"] == "lifespan":
+            return await handle_lifespan(self.scope, receive, send)
+
         if self.scope["type"] != "http":
             raise ValueError("unsupported")
 
         if self.scope["path"] == "/exc":
+            print("1 - THOWING ERROR")
```
hmstepanek, line 36:
```suggestion
```

newrelic/core/environment.py
```diff
@@ -170,6 +170,20 @@ def environment_settings():
...
 
+    if not dispatcher and "hypercorn" in sys.modules:
+        dispatcher.append(("Dispatcher", "hypercorn"))
+        hypercorn = sys.modules["hypercorn"]
+
+        if hasattr(hypercorn, "__version__"):
+            dispatcher.append(("Dispatcher Version", hypercorn.__version__))
+        else:
+            try:
+                from importlib.metadata import version
```
hmstepanek, line 181: I know this is an agent so sometimes we violate certain best practices (in this particular case importing at run time) that but I think that can be avoided here by referencing sys.modules["importlib"].metadata.version instead of importing it.

hmstepanek, line 181: I'm not sure if this is an issue or not cause I haven't tested it. I'm guessing probably not since the tests are passing but since there is a try pass here I just wanted to bring this up in case it might be an issue:
`importlib` is not available in python versions earlier than 3.8 so would there ever be a case here where `hypercorn.__version__` does not exist on python versions earlier than 3.8?

newrelic/core/environment.py
```diff
@@ -170,6 +170,20 @@ def environment_settings():
         if hasattr(uvicorn, "__version__"):
             dispatcher.append(("Dispatcher Version", uvicorn.__version__))
 
+    if not dispatcher and "hypercorn" in sys.modules:
+        dispatcher.append(("Dispatcher", "hypercorn"))
+        hypercorn = sys.modules["hypercorn"]
+
+        if hasattr(hypercorn, "__version__"):
+            dispatcher.append(("Dispatcher Version", hypercorn.__version__))
+        else:
```
hmstepanek, line 179: Could you add a comment above this section explaining what's going on here? I know primarily because you just talked about this in the presentation today but I don't think it's going to be as obvious years later when someone looks at this code again. It would also be helpful to provide the version(s) for which this special code is needed so at potentially a much later date we can remove this special code forever when we no longer need to support the version that requires it.

hmstepanek (changes requested):
I really appreciate the time you took muddling through some of the finer issues with instrumenting/testing this. Overall it looks great! I had a couple small suggestions/concerns but nothing major.

newrelic/core/environment.py
TimPansino: Moved things around and used the same get_version callable as the sys.modules code.

hmstepanek (dismissed):

newrelic/core/environment.py
TimPansino: Done

hmstepanek (approved):
Looks good. Looks like the tests are failing [here](https://github.com/newrelic/newrelic-python-agent/runs/8142202156?check_suite_focus=true#step:5:2433) but unrelated to this PR.

TimPansino:
🚢

PR #599, opened 2022-08-03
Upgrade MegaLinter to v6

TimPansino:
# Overview
* Upgrade MegaLinter to v6 using automated upgrader
* Also happened to upgrade all calls to actions/checkout to v3, leaving that in as well.


hmstepanek (approved):
Looks good.

PR #601, opened 2022-08-03
Fix PyPy coroutines testing to exclude uvloop

TimPansino:
# Overview
* Removes uvloop from PyPy testing for `coroutines_asyncio` as uvloop is not listed as compatible with PyPy and cannot be installed on PyPy3.7+.

# Related Github Issue
Blocks #590


Nayuken (commented):
After applying these changes Coroutines test are passing using pypy3.7

umaannamalai (approved):

PR #624, opened 2022-09-20
url obfuscation

dpacheconr:
# Overview
New agent capability - obfuscate sensitive information in URL's using regex patterns


hmstepanek:
Closing as a won't fix. This is a one-off customer issue and would require agreement amount all the agent teams to make it a feature.

PR #626, opened 2022-09-21
Remove kafka metric instrumentation.

umaannamalai:
Remove kafka metric instrumentation

We are not gonna do this for now. If customers want it we can add it later. This would push a lot of metrics up and customers might want to disable this which means we would need to add a config option. We don't want to do this extra work work right now without knowing whether customers would even use it and find it useful. Also of note there are other NR monitoring solutions that push up metric data like this.


hmstepanek (approved):
Looks good. I'm gonna merge this even though the tests are failing.

The test failures:
1. Looks like aiohttp released a new version which is causing failures (totally unrelated to this PR).
2. kafka tests are failing because of flakiness but we have a fix for this in another PR already so I'm not gonna worry about.

PR #644, opened 2022-10-03
Fix instumentation for sanic 22.0.9

hmstepanek:
# Overview
Fix instumentation for sanic 22.0.9

In 22.9.0 sanic introduced an API to control middleware ordering which broke 2 things:
* The addition of the new middleware mixin included a new method called `finalize_middleware` that sets the middleware to be used on the request.route during the `app._startup`. In order to register new middleware the `finalize_middleware` must be called, otherwise the app will use the previously registered middleware that was registered before `app._startup`.
* The middleware mixin wrapped certain middleware functions causing the name of the transaction and function trace to be the name of the mixin instead of the middleware function.

# Related Github Issue
https://issues.newrelic.com/browse/NR-53906

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -137,7 +137,7 @@ envlist =
     python-framework_pyramid-{pypy,py27,py38}-Pyramid0104,
     python-framework_pyramid-{pypy,py27,pypy37,py37,py38,py39,py310}-Pyramid0110-cornice,
     python-framework_pyramid-{py37,py38,py39,py310,pypy37}-Pyramidmaster,
-    python-framework_sanic-{py38,pypy37}-sanic{190301,1906,1812,1912,200904,210300,2109,2112,2203},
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -337,8 +337,8 @@ deps =
     framework_sanic-sanic2109: sanic<21.10
     framework_sanic-sanic2112: sanic<21.13
     framework_sanic-sanic2203: sanic<22.4
-    ; Temporarily pin this to the second to last release
-    framework_sanic-saniclatest: sanic==22.6.2
+    framework_sanic-sanic2290: sanic<22.9.1
```
umaannamalai, line 340: Can we change this to specifically test the latest version of sanic so we're alerted when a new release has breaking changes?

hmstepanek: I think this is already doing that.
```
framework_sanic-sanic2290: sanic<22.9.1
framework_sanic-saniclatest: sanic
```

umaannamalai: Ah yes, all good then!

umaannamalai (approved):
🚀

PR #656, opened 2022-10-10
Graphene-Django introspection fix

lrafeei:
This PR adds:
1. Instrumentation for graphene-django
1. A fix to the user seeing an introspection field (__schema) after loading/reloading a GraphiQL page while running New Relic.


newrelic/hooks/framework_graphql.py
```diff
@@ -482,6 +485,13 @@ def wrap_resolve_field(wrapped, instance, args, kwargs):
     else:
         field_path = field_path.key
 
+    # This is to catch the specific case of the user loading/reloading
```
hmstepanek, line 488: Thank you for the comment here!

tests/component_graphenedjango/wsgi.py
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+try:
+    from django.core.wsgi import get_wsgi_application
+
+    application = get_wsgi_application()
+
+except ImportError:
```
hmstepanek, line 20: Could we add a comment here as to under what circumstances this import error happens?

hmstepanek (dismissed):
Had one request for a comment otherwise looks good! I'm just gonna pre-approve this.

tests/component_graphenedjango/wsgi.py
lrafeei: Done!  This seems to account for something way outside of our support window, but I don't see the harm in having it exist for the time being.

lrafeei:
GraphQL v2 is no longer supported, so this PR is moot.  If future support for Graphene-Django is needed, a new PR will be submitted

PR #658, opened 2022-10-11
Add usage tracking metrics for Kafka clients.

umaannamalai:
This PR adds a new messagebroker metric for Kafka clients supported by the agent to be used for usage tracking.

### DIFF
--- a/.mega-linter.yml
+++ b/.mega-linter.yml
@@ -31,6 +31,8 @@ PYTHON_PYLINT_CONFIG_FILE: pyproject.toml
 PYTHON_ISORT_CONFIG_FILE: pyproject.toml
 PYTHON_BANDIT_CONFIG_FILE: pyproject.toml
 PYTHON_BANDIT_FILTER_REGEX_EXCLUDE: "tests"
-PYTHON_BANDIT_PRE_COMMANDS: ["pip install bandit[toml]"]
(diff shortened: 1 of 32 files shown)


hmstepanek (changes requested):
In general this looks great!

Given that it is common practice to have one service produce and a completely separate service consume I think we should capture this info on the consumer as well (since a service may never call the Producer, only the consumer).

I'd also suggest converting [these lines](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/core/environment.py#L43-L50) into a function, and importing/using that function in this bit of code here. While the `__version__` probably does work currently, at a certain point it may not and it would be nice to have all this code in one place. Not only that but I think it would be better to use the more universally recommended method of capturing the version (via `importlib`). Using `__version__` in the tests is totally fine IMO as I think it gives us a means of double checking, I'm only suggesting this for the actual instrumentation code.

TimPansino:
I think it would be prudent to send up package metrics for both producer and consumer for informational purposes, particularly since the consumer side are generally applications driven by kafka instead of just external traces like the producer side.

hmstepanek (approved):
Looks good! I just had one concern which I think will be fine but just wanted to mention it in case you thought it might be a problem.

PR #665, opened 2022-10-18
Add usage metrics for Daphne and Hypercorn.

umaannamalai:
This PR adds a dispatcher metric for Daphne and Hypercorn to track usage.

### DIFF
--- a/tests/adapter_daphne/test_daphne.py
+++ b/tests/adapter_daphne/test_daphne.py
@@ -112,7 +112,12 @@ def server_and_port():
 
 @override_application_settings({"transaction_name.naming_scheme": "framework"})
 def test_daphne_200(port, app):
-    @validate_transaction_metrics(callable_name(app))
+    @validate_transaction_metrics(
+        callable_name(app),
+        custom_metrics=[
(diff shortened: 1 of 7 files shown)


tests/adapter_daphne/test_daphne.py
```diff
@@ -112,7 +112,12 @@ async def fake_app(*args, **kwargs):
 
 @override_application_settings({"transaction_name.naming_scheme": "framework"})
 def test_daphne_200(port, app):
-    @validate_transaction_metrics(callable_name(app))
+    @validate_transaction_metrics(
```
hmstepanek, line 115: In the instrumentation there are 2 cases for both Daphne and Hypercorn that we need to test. One where there is already an active transaction and one where we start one. I think we should make sure we test both cases here if we can. I'm not sure why it wasn't already tested in here actually.

hmstepanek (changes requested):
Just one request about adding tests for the case where a transaction already exists. Other wise looks good

hmstepanek (approved):
This looks good. We are gonna go ahead and merge this and add test for the already active transaction test case in a separate PR.

PR #666, opened 2022-10-19
Increase default custom event limit.

umaannamalai:
This PR triples the default custom event limit in the agent to 3600 events per minute.


hmstepanek (approved):
Note I have changed this to merge into the feature branch instead of directly onto main.

PR #672, opened 2022-10-25
Validator transfer from fixtures.py to validators directory, Part 1

lrafeei:
This is part 1 of 5 for the transfer of the validators from `fixtures.py` to the validators directory.


hmstepanek (changes requested):
Thank you for finding time to do this! We don't have a lot of time on this team to dedicate to this type of work and it's a huge undertaking but a big help!


It looks like the old validators just got commented out instead of actually being removed from the fixtures.py-let's actually delete these.

I noticed there's a bunch of lint errors. Now would be a great opportunity to fix these. I don't think we necessarily need to fix all of them cause there are a ton so use your discretion but I think the unused import errors and out of order imports are ones we can definitely easily address as part of this PR.

lrafeei:
> Thank you for finding time to do this! We don't have a lot of time> Thank you for finding time to do this! We don't have a lot of time on this team to dedicate to this type of work and it's a huge undertaking but a big help!
> 
> It looks like the old validators just got commented out instead of actually being removed from the fixtures.py-let's actually delete these.
> 
> I noticed there's a bunch of lint errors. Now would be a great opportunity to fix these. I don't think we necessarily need to fix all of them cause there are a ton so use your discretion but I think the unused import errors and out of order imports are ones we can definitely easily address as part of this PR.

Thanks for looking at this!  So, I eventually removed the commented out code in the subsequent parts.  My thinking was that we'd incrementally merge them into the feature branch and merge the cleaned up feature branch into main (so none of the blocks of commented-out code would make it into main)

hmstepanek (approved):

PR #684, opened 2022-11-05
Fix failing botocore tests

hmstepanek:
# Overview
Fix failing botocore tests

In botocore >= 1.29.0, the url for queue changed so fix the tests to expect the new url.


tox.ini
```diff
@@ -262,6 +263,7 @@ deps =
     external_boto3-boto01: moto<2.0
     external_boto3-py27: rsa<4.7.1
     external_botocore-botocorelatest: botocore
+    external_botocore-botocore129: botocore<1.29
```
umaannamalai, line 266: We can call this one  `external_botocore-botocore128` since it's for `botocore<1.29`.

hmstepanek: Oops. Good catch!

PR #689, opened 2022-11-11
Fix package_version_utils.py logic

lrafeei:
Co-authored-by: Timothy Pansino <TimPansino@users.noreply.github.com>
Co-authored-by: Hannah Stepanek <hmstepanek@users.noreply.github.com>
Co-authored-by: Uma Annamalai <umaannamalai@users.noreply.github.com>

This PR fixes the package_version_utils.py file's logic and adds tests

### DIFF
--- a/newrelic/common/package_version_utils.py
+++ b/newrelic/common/package_version_utils.py
@@ -12,12 +12,52 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# To use:
+# Input: name of library as a string.
+# example:
+#   get_package_version("botocore")
+# Output: version of library as a string
+
 import sys
+
+# Need to account for 4 possible variations of version declaration specified in (rejected) PEP 396
+VERSION_ATTRS = ("__version__", "version", "__version_tuple__", "version_tuple")  # nosec
+NULL_VERSIONS = frozenset((None, "", "0", "0.0", "0.0.0", "0.0.0.0", (0,), (0, 0), (0, 0, 0), (0, 0, 0, 0)))  # nosec
 
 
 def get_package_version(name):
-    # importlib was introduced into the standard library starting in Python3.8.
-    if "importlib" in sys.modules and hasattr(sys.modules["importlib"], "metadata"):
-        return sys.modules["importlib"].metadata.version(name)  # pylint: disable=E1101
-    elif "pkg_resources" in sys.modules:
-        return sys.modules["pkg_resources"].get_distribution(name).version
+    def _get_package_version(name):
+        module = sys.modules.get(name, None)
+        version = None
+        for attr in VERSION_ATTRS:
+            try:
+                version = getattr(module, attr, None)
+                if version not in NULL_VERSIONS:
+                    return version
+            except Exception:
+                pass
+
+        # importlib was introduced into the standard library starting in Python3.8.
+        if "importlib" in sys.modules and hasattr(sys.modules["importlib"], "metadata"):
+            try:
+                version = sys.modules["importlib"].metadata.version(name)  # pylint: disable=E1101
+                if version not in NULL_VERSIONS:
+                    return version
+            except Exception:
+                pass
+
+        if "pkg_resources" in sys.modules:
+            try:
+                version = sys.modules["pkg_resources"].get_distribution(name).version
+                if version not in NULL_VERSIONS:
+                    return version
+            except Exception:
+                pass
+
+    version = _get_package_version(name)
+
+    # Coerce iterables into a string
+    if isinstance(version, (tuple, list)):
+        version = ".".join(str(v) for v in version)
+
+    return version
--- a/tests/agent_unittests/_test_package_version_utils.py
+++ b/tests/agent_unittests/_test_package_version_utils.py
@@ -1,0 +1,14 @@
+def test___version__():
+    pass
+
+
+def test__version():
+    pass
+
+
+def test___version():
+    pass
+
+
+def module():
+    pass
--- a/tests/agent_unittests/test_package_version_utils.py
+++ b/tests/agent_unittests/test_package_version_utils.py
@@ -1,0 +1,72 @@
+# Copyright 2010 New Relic, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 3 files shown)


newrelic/common/package_version_utils.py
```diff
@@ -12,12 +12,52 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# To use:
```
hmstepanek, line 15: This is great! Instead of putting it here though, can we move this into the function as a doc string:
```
def get_package_version(name):
    """Gets the version of the library.
    :param name: The name of library.
    :type name: str
    :return: The version of the library. Returns None if can't determine version.
    :type return: str or None

    Usage::
        >>> get_package_version("botocore")
                "1.1.0"
    """
```

tests/agent_unittests/_test_package_version_utils.py
```diff
@@ -0,0 +1,14 @@
+def test___version__():
```
hmstepanek, line 1: I think this accidentally got committed and we should just delete this file.

tests/agent_unittests/test_package_version_utils.py
```diff
@@ -0,0 +1,72 @@
...
+    yield pytest
+
+
+@pytest.mark.parametrize(
+    "attr,value,expected_value",
+    (
+        ("version", "1.2.3.4", "1.2.3.4"),
+        ("__version__", "1.3.5rc2", "1.3.5rc2"),
+        ("__version_tuple__", (3, 5, 8), "3.5.8"),
+        ("version_tuple", (3, 1, "0b2"), "3.1.0b2"),
```
hmstepanek, line 50: Just to test it works with something other than a tuple:
```suggestion
        ("version_tuple", [3, 1, "0b2"], "3.1.0b2"),
```

hmstepanek (changes requested):
Looks good! Just a couple minor changes.

tests/agent_unittests/_test_package_version_utils.py
lrafeei: I believe this file still needs to exist but just contain empty modules for the existing tests to run

tests/agent_unittests/test_package_version_utils.py
lrafeei: Is this supposed to work if it's not a tuple (for "version_tuple" or "__version_tuple__") or a string (for "version" or "__version__")?

newrelic/common/package_version_utils.py
lrafeei: Good call!  Done

tests/agent_unittests/test_package_version_utils.py
hmstepanek: Only if the version_tuple's are a list/iterable instead of a tuple. We do expect the version's to be strings not iterables.
See implementation:
```
    # Coerce iterables into a string
    if isinstance(version, (tuple, list)):
        version = ".".join(str(v) for v in version)
```

lrafeei: Good point.  Done!

tests/agent_unittests/_test_package_version_utils.py
TimPansino: Can you empty out this file and add the copyright notice to it? It can just be completely empty.

newrelic/common/package_version_utils.py
```diff
@@ -14,10 +14,56 @@
...
+                "1.1.0"
+    """
+
+    def _get_package_version(name):
+        module = sys.modules.get(name, None)
+        version = None
+        for attr in VERSION_ATTRS:
+            try:
+                version = getattr(module, attr, None)
+                version = tuple(version) if isinstance(version, list) else version  # cast potential lists into tuples
```
hmstepanek, line 40: This is a bit nit-picky but I think this is clearer than what's there:
```suggestion
                # Cast any version specified as a list into a tuple.
                version = tuple(version) if isinstance(version, list) else version  
```

newrelic/common/package_version_utils.py
```diff
@@ -14,10 +14,56 @@
...
+                version = sys.modules["pkg_resources"].get_distribution(name).version
+                if version not in NULL_VERSIONS:
+                    return version
+            except Exception:
+                pass
+
+    version = _get_package_version(name)
+
+    # Coerce iterables into a string
+    if isinstance(version, (tuple, list)):
```
hmstepanek, line 66: Since it's already been cast to a tuple now we can simplify this if statement:
```suggestion
    if isinstance(version, tuple):
```

tests/agent_unittests/_test_package_version_utils.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2010 New Relic, Inc.
```
hmstepanek, line 1: This file needs to be deleted entirely and not committed to the repo.

hmstepanek (approved):
A couple suggested fixups. I'm just gonna pre-approve this assuming you'll commit those.

tests/agent_unittests/test_package_version_utils.py
```diff
@@ -0,0 +1,72 @@
...
+@pytest.mark.parametrize(
+    "attr,value,expected_value",
+    (
+        ("version", "1.2.3.4", "1.2.3.4"),
+        ("__version__", "1.3.5rc2", "1.3.5rc2"),
+        ("__version_tuple__", (3, 5, 8), "3.5.8"),
+        ("version_tuple", [3, 1, "0b2"], "3.1.0b2"),
+    ),
+)
+def test_get_package_version(attr, value, expected_value):
```
hmstepanek, line 53: Change to the following and then we can get rid of that empty module:
```
def test_get_package_version(attr, value, expected_value):
    setattr(pytest, attr, value)
    
    version = get_package_version("pytest")

    assert version == expected_value

    delattr(pytest, attr)
```

hmstepanek: I've changed the order of the delattr and the assert. Usually I like to follow the following test format since it's easy to read:
```
# setup

# run test

# asserts

# teardown
```

hmstepanek (commented):
Ok so instead of using that empty module to test we can use pytest since we already monkey patch it anyway. I've provided the code change for the test in order to use the pytest module. Then you can delete the import and the file.

newrelic/common/package_version_utils.py
lrafeei: Done!

tests/agent_unittests/test_package_version_utils.py
lrafeei: Done!

PR #691, opened 2022-11-16
Add tree model function traces

hmstepanek:
# Overview
Add function traces around the following methods in the sklearn tree models:
* `fit`
* `predict`
* `predict_log_proba`
* `predict_proba`
* `score`

_Note: regressor models do not have proba methods._

# Related Github Issue
This is a partial fix for: [Add function traces to model methods](https://issues.newrelic.com/browse/NR-63290).


newrelic/hooks/component_sklearn.py
```diff
@@ -0,0 +1,72 @@
...
+
+from newrelic.api.function_trace import FunctionTraceWrapper
+from newrelic.api.transaction import current_transaction
+from newrelic.common.object_wrapper import function_wrapper, wrap_function_wrapper
+
+
+def wrap_model_method(method_name):
+    @function_wrapper
+    def _wrap_method(wrapped, instance, args, kwargs):
+        # If there is no transaction, do not wrap anything.
```
hmstepanek, line 23: I wasn't sure if this was necessary/how I should test for this case?

newrelic/hooks/component_sklearn.py
```diff
@@ -0,0 +1,64 @@
...
+    with FunctionTrace(name=name, group="MLModel/Sklearn/Named", source=wrapped):
+        return_val = wrapped(*args, **kwargs)
+
+    # Set the _nr_wrapped attribute to denote that this method is no longer wrapped.
+    setattr(instance, "_nr_wrapped_%s" % method_name, False)
+
+    return return_val
+
+
+def _nr_instrument_model(module, model_class):
```
umaannamalai, line 48: In terms of convention, we tend to name the instrument functions so that they match the path to the file where the functions to be instrumented live (also specified in the `process_module_definition`). This helps make it clear which part of the library code is being instrumented by our agent function. In this case, it looks like this is instrumenting methods in the ._classes.py file so we'd name this function `instrument_sklearn_tree_.classes`

newrelic/hooks/component_sklearn.py
```diff
@@ -0,0 +1,64 @@
...
+    methods_to_wrap = ("predict", "fit", "fit_predict", "predict_log_proba", "predict_proba", "transform", "score")
+    for method_name in methods_to_wrap:
+        if hasattr(getattr(module, model_class), method_name):
+            wrap_function_wrapper(module, "%s.%s" % (model_class, method_name), wrap_method)
+
+
+def instrument_sklearn_tree_models(module):
```
umaannamalai, line 55: In terms of convention, we tend to name the instrument functions so that they match the path to the file where the functions to be instrumented live (also specified in the process_module_definition). This helps make it clear which part of the library code is being instrumented by our agent function. In this case, it looks like this is instrumenting methods in the ._classes.py file so we'd name this function instrument_sklearn_tree_.classes

newrelic/hooks/component_sklearn.py
```diff
@@ -0,0 +1,64 @@
...
+
+    method_name = wrapped.__name__
+
+    # If the method has already been wrapped do not wrap it again. This happens
+    # when one model inherits from another and they both implement the method.
+    if getattr(instance, "_nr_wrapped_%s" % method_name, False):
+        return wrapped(*args, **kwargs)
+
+    # Set the _nr_wrapped attribute to denote that this method is being wrapped.
+    setattr(instance, "_nr_wrapped_%s" % method_name, True)
```
TimPansino, line 33: Ensuring we wrap only once is a good idea, but let's move this logic to import time rather than runtime.

We usually use a function named something like `wrap_method_wrapper`  or `wrap_method` as a helper function that will apply a `method_wrapper` around the method. It should handle this logic of checking the `_nr_wrapped` attr. That way we lower the performance impact and only run this code at import.

newrelic/hooks/component_sklearn.py
```diff
@@ -0,0 +1,64 @@
...
+    setattr(instance, "_nr_wrapped_%s" % method_name, False)
+
+    return return_val
+
+
+def _nr_instrument_model(module, model_class):
+    methods_to_wrap = ("predict", "fit", "fit_predict", "predict_log_proba", "predict_proba", "transform", "score")
+    for method_name in methods_to_wrap:
+        if hasattr(getattr(module, model_class), method_name):
+            wrap_function_wrapper(module, "%s.%s" % (model_class, method_name), wrap_method)
```
TimPansino, line 52: Based on the above comment, this `wrap_function_wrapper` call would be replaced with a call to a helper function.

newrelic/hooks/component_sklearn.py
```diff
@@ -0,0 +1,64 @@
...
+def instrument_sklearn_tree_models(module):
+    model_classes = (
+        "DecisionTreeClassifier",
+        "DecisionTreeRegressor",
+        "ExtraTreeClassifier",
+        "ExtraTreeRegressor",
+    )
+    for model_cls in model_classes:
+        if hasattr(module, model_cls):
+            _nr_instrument_model(module, model_cls)
```
TimPansino, line 64: If we're going to reuse this wrapping logic everywhere, can we make this a generic `_instrument_sklearn_module_classes` or something that takes in module and a list of classes, and make our instrumenter functions all just call the generic?

tox.ini
```diff
@@ -203,6 +206,9 @@ deps =
     application_celery: celery<6.0
     application_celery-py{py37,37}: importlib-metadata<5.0
     application_gearman: gearman<3.0.0
+    component_sklearn-scikitlearnlatest: scikit-learn
+    component_sklearn-scikitlearn101: scikit-learn < 1.1
+    component_sklearn-scikitlearn020: scikit-learn < 0.21
```
TimPansino, line 211: Use 2 digits for each tox version to match convention. `0101` and `0020`. Makes it slightly easier to read.

tests/component_sklearn/test_tree_models.py
```diff
@@ -0,0 +1,154 @@
...
+        ],
+        "DecisionTreeRegressor": [
+            ("MLModel/Sklearn/Named/DecisionTreeRegressor.fit", 1),
+            ("MLModel/Sklearn/Named/DecisionTreeRegressor.predict", 2),
+            ("MLModel/Sklearn/Named/DecisionTreeRegressor.score", 1),
+        ],
+    }
+    expected_transaction_name = "test_tree_models:_test"
+    if six.PY3:
+        expected_transaction_name = "test_tree_models:test_model_methods_wrapped_in_function_trace.<locals>._test"
```
TimPansino, line 56: Any reason not to make this a oneliner?

tests/component_sklearn/test_tree_models.py
```diff
@@ -0,0 +1,154 @@
...
+        ],
+        "DecisionTreeRegressor": [
+            ("MLModel/Sklearn/Named/DecisionTreeRegressor.fit", 1),
+            ("MLModel/Sklearn/Named/DecisionTreeRegressor.predict", 4),
+            ("MLModel/Sklearn/Named/DecisionTreeRegressor.score", 2),
+        ],
+    }
+    expected_transaction_name = "test_tree_models:_test"
+    if six.PY3:
+        expected_transaction_name = "test_tree_models:test_multiple_calls_to_model_methods.<locals>._test"
```
TimPansino, line 103: This as well. ^

newrelic/hooks/component_sklearn.py
```diff
@@ -0,0 +1,64 @@
...
+from newrelic.api.transaction import current_transaction
+from newrelic.common.object_wrapper import wrap_function_wrapper
+
+
+def wrap_method(wrapped, instance, args, kwargs):
+    # If there is no transaction, do not wrap anything.
+    if not current_transaction():
+        return wrapped(*args, **kwargs)
+
+    method_name = wrapped.__name__
```
TimPansino, line 25: `newrelic.common.object_names` includes `callable_name` and `object_context` functions that have more robust logic than a simple `__name__` lookup which may be missing or incorrect depending on what weird things a customer might choose to do.

If you want a fully qualified class path without the module included, you'll likely want `object_context(wrapped)[1]`. You can look at the comments on those functions for more info.

newrelic/hooks/component_sklearn.py
```diff
@@ -0,0 +1,64 @@
...
+    # If the method has already been wrapped do not wrap it again. This happens
+    # when one model inherits from another and they both implement the method.
+    if getattr(instance, "_nr_wrapped_%s" % method_name, False):
+        return wrapped(*args, **kwargs)
+
+    # Set the _nr_wrapped attribute to denote that this method is being wrapped.
+    setattr(instance, "_nr_wrapped_%s" % method_name, True)
+
+    # MLModel/Sklearn/Named/<class name>.<method name>
+    func_name = wrapped.__name__
```
TimPansino, line 36: Another that could likely use `object_context`.

newrelic/hooks/component_sklearn.py
```diff
@@ -0,0 +1,64 @@
...
+    model_name = wrapped.__self__.__class__.__name__
```
TimPansino, line 37:
```suggestion
    model_name = object_context(type(instance))[1]
```

You probably want the above. `wrapped.__self__` involves a lookup for something that's already available as `instance`.

tests/component_sklearn/test_tree_models.py
```diff
@@ -0,0 +1,154 @@
...
+    def _test():
+        x_test = [[2.0, 2.0], [2.0, 1.0]]
+        y_test = [1, 1]
+
+        model = run_tree_model()
+
+        model.predict(x_test)
+        model.score(x_test, y_test)
+        # Only classifier models have proba methods.
+        if tree_model_name in ("DecisionTreeClassifier", "ExtraTreeClassifier"):
```
TimPansino, line 121: Replacing with a `hasattr` might make this more portable so it can be reused for other tests.

newrelic/hooks/component_sklearn.py
```diff
@@ -0,0 +1,64 @@
...
+    name = "%s.%s" % (model_name, func_name)
+    with FunctionTrace(name=name, group="MLModel/Sklearn/Named", source=wrapped):
+        return_val = wrapped(*args, **kwargs)
+
+    # Set the _nr_wrapped attribute to denote that this method is no longer wrapped.
+    setattr(instance, "_nr_wrapped_%s" % method_name, False)
```
TimPansino, line 43: Is there a reason we set this to a more specific name than `_nr_wrapped`? Would having a single name not be easier?

TimPansino (changes requested):

tests/component_sklearn/test_tree_models.py
hmstepanek: I suspect this is going to be very specific to the types of models so I'm gonna leave it for now as is. If it turns out to be useful as we evaluate more models we can totally do this though!

newrelic/hooks/component_sklearn.py
hmstepanek: I think we went over this today but it's specific to the name of method that we are wrapping because one method may call another method and we have to be able to differentiate them.
```
def score(self):  # We wrap score.
   self.predict()  # We also wrap predict which is called inside the score method.
```

newrelic/hooks/component_sklearn.py
hmstepanek: I think this is a non-issue now.

newrelic/hooks/component_sklearn.py
hmstepanek: I think this is a non-issue now though good to know for the future!

newrelic/hooks/component_sklearn.py
hmstepanek: I think this is a non-issue now.

tests/component_sklearn/test_tree_models.py
hmstepanek: Ha oops thanks!

tests/component_sklearn/test_tree_models.py
hmstepanek: Good thinking!

tests/mlmodel_sklearn/test_tree_models.py
```diff
@@ -0,0 +1,157 @@
...
+    def _test():
+        x_test = [[2.0, 2.0], [2.0, 1.0]]
+        y_test = [1, 1]
+
+        model = run_tree_model()
+
+        model.predict(x_test)
+        model.score(x_test, y_test)
+        # Only classifier models have proba methods.
+        if tree_model_name in ("DecisionTreeClassifier", "ExtraTreeClassifier"):
```
TimPansino, line 123: This line can also be changed to a hasattr check

TimPansino (changes requested):

TimPansino (approved):

PR #692, opened 2022-11-16
Fix Trace Cache Iteration Crash

TimPansino:

# Overview
Describe the changes present in the pull request

# Related Github Issue

# Testing
exercise your changes.

### DIFF
--- a/newrelic/core/trace_cache.py
+++ b/newrelic/core/trace_cache.py
@@ -197,7 +197,7 @@ class TraceCache(object):
         debug = global_settings().debug
 
         if debug.enable_coroutine_profiling:
-            for thread_id, trace in list(self._cache.items()):
+            for thread_id, trace in self._cache.copy().items():
                 transaction = trace.transaction
                 if transaction and transaction._greenlet is not None:
                     gr = transaction._greenlet()
@@ -359,7 +359,7 @@ class TraceCache(object):
         task = getattr(transaction.root_span, "_task", None)
         loop = get_event_loop(task)
 
-        for trace in list(self._cache.values()):
+        for trace in self._cache.copy().values():
             if trace in seen:
                 continue
 


hmstepanek (changes requested):
I apologize that I didn't see this earlier. I had this one on my todo list but I didn't see you'd already opened a PR for it until just now. I do not believe this fix will work as the crash was during the list() cast prior to the loop running. If the list casting didn't fix this, I do not believe a copy will fix it either. My comment on the issue was a bit rushed and not thought through. This type of fix would only work if the code inside the loop was where the _cache was being modified (the list casting would have also fixed this kind of issue). However, since the crash is still happening, I do not think that's what's happening here (nor do I see any code inside the loop that modifies the size of the _cache). I think this is being modified in a separate thread so the two threads are accessing the _cache simultaneously. I believe we need to add a lock here. I've opened a draft PR that does that here: https://github.com/newrelic/newrelic-python-agent/pull/702/files. We can either open that draft PR or we can modify this fix to be like what's in that draft.

TimPansino:
@hmstepanek 

> If the list casting didn't fix this, I do not believe a copy will fix it either.

The issue with a list is that it's not casting, it's iterating on the dict.items into a list, then iterating the list. The crash has less area to happen but can still happen if the change is made during the short time when the dict.items is iterated over by the list constructor (instead of the longer running for loop).

A shallow `copy` doesn't iterate as far as I'm aware, this is a patch that's been used in other places I've seen to guard against iteration issues.

That being said, I'm not opposed to using locking but I wonder if that causes potential performance issues. It's certainly harder to maintain. I think the simpler solution might be more appropriate considering this is only affecting event loop wait time and debug logic, as opposed to potential impacts to any time that the trace cache is accessed.

TimPansino:
Superseded by #704

PR #694, opened 2022-11-21
Capture scorer results

hmstepanek:
# Overview
Add score results attributes to metric scorers.

# Related Github Issue
Fixes [Capture scorer results](https://issues.newrelic.com/browse/NR-63755).


newrelic/hooks/mlmodel_sklearn.py
```diff
@@ -12,12 +12,81 @@
...
+            return NRBoolType(data)
+
+        # Attempt to subclass the type directly.
+        class NRWrapType(type(data)):
+            _nr_wrapped_model_name = model_name
+
+        return NRWrapType(data)
+    except:  # Squash any exceptions resulting from attempted wrap typing.
+        pass
+    return data
```
TimPansino, line 89: This function is super brittle and dangerous, I'm not comfortable with the implementation here. If all we need is an object that we can set attributes on that can be achieved much more easily with an ObjectProxy. No need to make custom subclasses with custom compatibility functionality required.

I'll push up a commit to swap this out for a call to ObjectProxy and you can take a look at it and see if you agree.

newrelic/hooks/mlmodel_sklearn.py
```diff
@@ -66,6 +139,33 @@ def _instrument_sklearn_models(module, model_classes):
...
+    y_true, y_pred, args, kwargs = _bind_scorer(*args, **kwargs)
+    model_name = "Unknown"
+    if hasattr(y_pred, "_nr_wrapped_model_name"):
+        model_name = y_pred._nr_wrapped_model_name
+    # Attribute values must be int, float, str, or boolean. If it's not one of these
+    # types and an iterable add the values as separate attributes.
+    if not isinstance(score, (str, int, float, bool)):
+        if hasattr(score, "__iter__"):
+            for i, s in enumerate(score):
+                transaction._add_agent_attribute("%s.%s[%s]" % (model_name, wrapped.__name__, i), s)
```
TimPansino, line 163: Branch coverage says this isn't covered at all, we should ensure this is tested.

TimPansino (changes requested):

TimPansino (approved):

PR #697, opened 2022-11-23
Add ensemble model function traces

lrafeei:
Add function traces around the following methods in the sklearn ensemble models:

- fit
- predict
- predict_log_proba
- predict_proba
- score

_Note: regressor models do not have proba methods._


hmstepanek (changes requested):
Couple suggestions here on making the code more readable. This is a hard one so don't feel bad!

hmstepanek (approved):
Looks great! Thank you for doing all that test cleanup. The tests are 100% easier to read now!

PR #700, opened 2022-11-24
Add cluster model function traces

lrafeei:
This PR adds instrumentation to the cluster model for sklearn.

The functions that are instrumented:

- fit
- fit_predict
- predict
- transform


newrelic/config.py
```diff
@@ -2902,6 +2902,120 @@ def _process_module_builtin_defaults():
...
+        "instrument_sklearn_cluster_models",
+    )
+
+    _process_module_definition(
+        "sklearn.cluster.dbscan_",
+        "newrelic.hooks.mlmodel_sklearn",
+        "instrument_sklearn_cluster_models",
+    )
+
+    _process_module_definition(
```
hmstepanek, line 2959: I think we can just delete this one entirely since the only thing in there is a transformer and we aren't instrumenting that.

newrelic/hooks/mlmodel_sklearn.py
```diff
@@ -189,6 +189,43 @@ def instrument_sklearn_ensemble_hist_models(module):
...
+        "MeanShift",
+        "OPTICS",
+    )
+    _instrument_sklearn_models(module, model_classes)
+
+
+def instrument_sklearn_cluster_agglomerative_models(module):
+    model_classes = (
+        "AgglomerativeClustering",
+        "FeatureAgglomeration",
```
hmstepanek, line 206: These exist in older versions:
```suggestion
        "FeatureAgglomeration",
        "Ward",
        "WardAgglomeration",
```

tests/mlmodel_sklearn/test_cluster_models.py
```diff
@@ -0,0 +1,186 @@
...
+            ("Function/MLModel/Sklearn/Named/MeanShift.predict", 1),
+            ("Function/MLModel/Sklearn/Named/MeanShift.fit_predict", 1),
+        ],
+        "MiniBatchKMeans": [
+            ("Function/MLModel/Sklearn/Named/MiniBatchKMeans.fit", 2),
+            ("Function/MLModel/Sklearn/Named/MiniBatchKMeans.predict", 1),
+            ("Function/MLModel/Sklearn/Named/MiniBatchKMeans.fit_predict", 1),
+        ],
+        "SpectralBiclustering": [
+            ("Function/MLModel/Sklearn/Named/SpectralBiclustering.fit", 1),
```
hmstepanek, line 89: If these only have a fit method they probably aren't a true model and we shouldn't be instrumenting them as a customer wouldn't use these directly. Let's double check with @ykriger-newrelic though.

tests/mlmodel_sklearn/test_cluster_models.py
```diff
@@ -0,0 +1,186 @@
...
+        ],
+        "SpectralClustering": [
+            ("Function/MLModel/Sklearn/Named/SpectralClustering.fit", 2),
+            ("Function/MLModel/Sklearn/Named/SpectralClustering.fit_predict", 1),
+        ],
+    }
+    expected_transaction_name = "test_cluster_models:_test"
+    if six.PY3:
+        expected_transaction_name = (
+            "test_cluster_models:test_below_v1_1_model_methods_wrapped_in_function_trace.<locals>._test"
```
hmstepanek, line 102: We need to add those two other models into here: Ward and WardAgglomeration. I'm not sure exactly what version those got removed. I saw them in version 0.15.0.

tests/mlmodel_sklearn/test_cluster_models.py
```diff
@@ -0,0 +1,186 @@
...
+        from sklearn.model_selection import train_test_split
+
+        X, y = load_iris(return_X_y=True)
+        x_train, x_test, y_train, y_test = train_test_split(X, y, stratify=y, random_state=0)
+
+        clf = getattr(sklearn.cluster, cluster_model_name)()
+
+        model = clf.fit(x_train, y_train)
+
+        if hasattr(model, "predict"):
```
hmstepanek, line 171: Assuming we don't want to include classes that don't have a predict method you can remove this hasattr check around the predict.

hmstepanek (changes requested):
Looks good overall, had a question for @ykriger-newrelic on this one. I'm not sure it would really be considered a model without a predict method. I also found a couple models that were missing.

tests/mlmodel_sklearn/test_cluster_models.py
lrafeei: I believe the support window for this is 3 years, so anything before 2019 would not be supported anyway (seems like v0.20 is a good stopping point for us now).  @ak-war can you confirm?

newrelic/hooks/mlmodel_sklearn.py
lrafeei: I believe Ward and WardAgglomeration are models that are outside of our supportability window of 3 years

hmstepanek: Never mind on this change. ^ These are not in version 0.22.0 which I believe is the 3 year old version of this library and outside our support window.

tests/mlmodel_sklearn/test_cluster_models.py
ykriger-newrelic: That's right, we don't have to instrument this model.

newrelic/config.py
lrafeei: There's been a change in the instrumentation criteria in that we instrument modules that have "fit" (and not necessarily "fit" and "predict").  Because of that, we do need to keep this since this one instruments "fit" and "transform"

tests/mlmodel_sklearn/test_cluster_models.py
lrafeei: Since we are instrumenting models that have fit (and not necessarily fit and predict) we do need to keep these.

tests/mlmodel_sklearn/test_cluster_models.py
lrafeei: Since we are instrumenting models that have fit (and not necessarily fit and predict) we might run across models that have only "fit"

hmstepanek (approved):

PR #703, opened 2022-11-29
Add linear model function traces

lrafeei:
Add function traces around the following methods in the sklearn linear_model models:

- fit
- predict
- score


tests/mlmodel_sklearn/test_linear_models.py
```diff
@@ -0,0 +1,342 @@
...
+        "PoissonRegressor",
+        "GammaRegressor",
+        "TweedieRegressor",
+        "QuantileRegressor",
+        "SGDClassifier",
+        "SGDRegressor",
+        "SGDOneClassSVM",
+    ],
+)
+def test_above_v1_1_model_methods_wrapped_in_function_trace(linear_model_name, run_linear_model):
```
hmstepanek, line 247: It looks like there are more than just these models above 1.1.0. We should test everything we can on each version since the models are sometimes in different files and we need to make sure we instrument those files.

hmstepanek (changes requested):
Looks good. I just had one request about testing all the models that exist on the latest version . It looks like we are missing a fair amount there.

tests/mlmodel_sklearn/test_linear_models.py
lrafeei: Sorry if I missed something, but I could not find any more models that were missed above 1.1 (looking in the latest release on main) unless you were referring to the ones in sgd_fast, which are Cython files.

hmstepanek: I don't see RidgeClassifier sklearn/linear_model/_ridge.py mentioned in this test but I do see it on main for example. I see a lot of the model classes in test_below_v1_1 on main. OrthogonalMatchingPursuitCV in sklearn/linear_model/_omp.py:class is another one. I think we just need to copy the majority of those models in below 1.1 into this above 1.1 as well.

lrafeei: Ooh, I see what you mean.  The ones in the "below_ v1_1" test do not have a skipif condition, so those models are in all versions, but the ones in the "above_v1_1" do, so those only test the ones added after v1.1.  Maybe I should change the test name to be less confusing?

TimPansino:
Some tests are failing `FAILED test_ensemble_models.py::test_above_v1_1_model_methods_wrapped_in_function_trace[VotingRegressor]`

tests/mlmodel_sklearn/test_linear_models.py
hmstepanek: Ah I see. I was being really dumb about that. I think the name is fine as is.

hmstepanek (approved):

PR #683, opened 2022-11-04
Fix failing flask_rest tests

hmstepanek:
# Overview
Fix failing flask_rest tests
* Drop support for flask-restplus.
* The latest version of flask-restx dropped support for 2.7 so split the flask_rest tests up into 2 tox envs.

PR #705, opened 2022-12-01
Revert Removal of Flask Restplus Hook

TimPansino:
# Overview

* Flask Restplus hook was removed, adding it back. 
* This module is now unmaintained and outside our support window, and will have no tested support going forward but existing hooks and functionality will remain in place.

# Related Github Issue

Partially Reverts #683


TimPansino:
Wrong name, fixed.

hmstepanek (approved):

PR #706, opened 2022-12-01
Add config setting for sklearn inference event capture.

umaannamalai:
This PR adds a config/ environment variable setting to toggle capture of inference events in `sklearn`.


newrelic/hooks/component_sklearn.py
```diff
@@ -0,0 +1,68 @@
+# Copyright 2010 New Relic, Inc.
```
hmstepanek, line 1: This file was renamed so I believe this was created as the result of a merge conflict. This file should be deleted.

tests/component_sklearn/conftest.py
```diff
@@ -0,0 +1,38 @@
+# Copyright 2010 New Relic, Inc.
```
hmstepanek, line 1: Same thing here-this file should be deleted.

tests/component_sklearn/test_tree_models.py
```diff
@@ -0,0 +1,156 @@
+# Copyright 2010 New Relic, Inc.
```
hmstepanek, line 1: Same thing here-this file should be deleted.

hmstepanek (changes requested):
Looks good. The only thing that needs to be fixed is the merge conflict issue.

hmstepanek:
The tests needed to be kicked off anyway so I went ahead and removed those files.

PR #708, opened 2022-12-07
Fix Type Constructor Classes in Code Level Metrics

### LINKED ISSUE #515: NewRelic agent: Failed to extract source code context from callable
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

**Description**
I added newrelic apm to a tornado based application. The agent is outputting a lot of error messages: `newrelic.api.time_trace ERROR - Failed to extract source code context from callable <address.to.request.handler.here>. Report this issue to newrelic support.`

Is this something in our code?

**Expected Behavior**
No error messages and expected to see transactions in the APM.

**Steps to Reproduce**
1. install `newrelic==7.10.0.175`
2. wrap python main.py execution in `newrelic-admin run-python path/to/code`

**Your Environment**
Python 3.9.2
NewRelic 7.10.0.175
Tornado 6.1.0

TimPansino:
# Overview

* Previously either no data was reported or a crash occurred (depending on agent version) when using a class created using the [3 argument type constructor](https://docs.python.org/3/library/functions.html?highlight=type#type). Fixed and properly returns without line number (which is unavailable).
* Pin tox major version to v3 until CI can be fixed.

# Related Github Issue
Closes #515

### DIFF
--- a/.github/actions/setup-python-matrix/action.yml
+++ b/.github/actions/setup-python-matrix/action.yml
@@ -47,4 +47,4 @@ runs:
       shell: bash
       run: |
         python3.10 -m pip install -U pip
-        python3.10 -m pip install -U wheel setuptools tox virtualenv!=20.0.24
+        python3.10 -m pip install -U wheel setuptools 'tox<4' virtualenv!=20.0.24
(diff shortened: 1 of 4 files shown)


TimPansino:
Problem stemmed from an uncaught OSError, catching all exception types when attempting to find line numbers now.

hmstepanek (approved):
I fixed some lint errors and converted the helper function into a fixture. Otherwise looks good!

PR #710, opened 2022-12-08
CodeCov Integration

TimPansino:
# Overview

* Configure coverage reporting from multiple environments.


tests/testing_support/fixtures.py
```diff
@@ -1483,35 +1483,41 @@ def code_coverage_fixture(source=None):
...
+        coverage_directory = "htmlcov"
+        xml_report = "coverage.xml"
+
     @pytest.fixture(scope="session")
     def _code_coverage_fixture(request):
         if not source:
-            return
-
-        if os.environ.get("GITHUB_ACTIONS") is not None:
+            yield None
```
hmstepanek, line 1503: Do you need this `yield None` here or would the `return` below it be sufficient?

hmstepanek (commented):
I was trying to see the code coverage report locally but I don't see that it's getting generated? It looks like it's working fine in github actions.

tests/testing_support/fixtures.py
TimPansino: Since it's a generator based fixture, without the yield None you'll get a 

`ValueError: code_coverage did not yield a value`

I'll put a comment to this effect.

TimPansino:
> I was trying to see the code coverage report locally but I don't see that it's getting generated? It looks like it's working fine in github actions.

Code coverage reports should be dumped into their tox environment's folder to keep them separate.

hmstepanek (approved):
I see it now! Thank you! Looks good!

PR #711, opened 2022-12-09
Add DevContainer

TimPansino:
# Overview

* Adds configuration to use [newrelic-experimental/pyenv-devcontainer](https://github.com/newrelic-experimental/pyenv-devcontainer) as a devcontainer for VS Code / Github Codespaces.


hmstepanek (changes requested):
Thanks for putting in all the work to create this! Can you add some documentation about how to configure VS code to use this file? If this is in our repo I want everyone to know how to use it going forward including any contributors that aren't our team. There's a brief section in the README called Contributing that might be a good spot to add the steps. It might also be good at the same time to add steps for those who aren't necessarily using VS code but still want to use the dev container.

hmstepanek (approved):
Thanks for adding these docs! In general they look great. The only thing I would request we add to this is a step about that gotcha we ran into with needing to create a GHCR token under the VS code steps. I'm just gonna pre-approve this assuming you'll add that.

PR #713, opened 2022-12-15
Add compose model function traces

lrafeei:
Add function traces around the following methods in the sklearn compose models:

- fit
- predict
- transform


tests/mlmodel_sklearn/test_compose_models.py
```diff
@@ -0,0 +1,90 @@
...
+
+        X = [[0.0, 1.0, 2.0, 2.0], [1.0, 1.0, 0.0, 1.0]]
+        y = None
+        kwargs = {
+            "transformers": [("norm1", Normalizer(norm="l1"), [0, 1]), ("norm2", Normalizer(norm="l1"), slice(2, 4))]
+        }
+        if compose_model_name == "TransformedTargetRegressor":
+            kwargs = {"regressor": LinearRegression()}
+            X = np.arange(4).reshape(-1, 1)
+            y = np.exp(2 * X).ravel()
```
TimPansino, line 78: Can we convert this to if else instead of global and if? Would prefer that for readability.

TimPansino (changes requested):

TimPansino (approved):

PR #715, opened 2022-12-15
Prediction metric stats

hmstepanek:
# Overview
Record custom metrics stats about feature and label values during a prediction.

# Related Github Issue
This fixes https://issues.newrelic.com/browse/NR-63764.


newrelic/hooks/mlmodel_sklearn.py
```diff
@@ -79,13 +90,130 @@ def _nr_wrapper_method(wrapped, instance, args, kwargs):
...
+                ("%s/%s" % (metric_name, "Percentile25"), float(percentile25[index])),
+                ("%s/%s" % (metric_name, "Percentile50"), float(percentile50[index])),
+                ("%s/%s" % (metric_name, "Percentile75"), float(percentile75[index])),
+                ("%s/%s" % (metric_name, "StandardDeviation"), float(standard_deviation[index])),
+                ("%s/%s" % (metric_name, "Min"), float(_min[index])),
+                ("%s/%s" % (metric_name, "Max"), float(_max[index])),
+                ("%s/%s" % (metric_name, "Count"), _count),
+            ]
+        )
+    return (mean, percentile25, percentile50, percentile75, standard_deviation, _min, _max, _count)
```
umaannamalai, line 158: Do we need to return all our stats values here? I see two calls to this function but they don't appear to be assigning the return value to a variable that is used later.

newrelic/hooks/mlmodel_sklearn.py
```diff
@@ -79,13 +90,130 @@ def _nr_wrapper_method(wrapped, instance, args, kwargs):
...
     wrap_function_wrapper(module, "%s.%s" % (_class, method), _nr_wrapper_method)
 
 
+def _calc_prediction_feature_stats(prediction_input, _class, feature_column_names):
+    import numpy as np
+
+    # Drop any feature columns that are not numeric since we can't compute stats
+    # on non-numeric columns.
+    x = np.array(prediction_input)
+    isnumeric_features = np.apply_along_axis(isnumeric, 0, x)
```
umaannamalai, line 108: This is neat.

newrelic/hooks/mlmodel_sklearn.py
hmstepanek: Ooh no we don't! Good catch!

umaannamalai (approved):

PR #736, opened 2023-01-05
Fix log decorating to be JSON compatible

lrafeei:
# Overview
Right now, local log decorating simply adds the New Relic Metadata to the message and does not account for the logs to be potentially in JSON format.
Ex:
{"log_message": "Log Message"} NR-LINKING|{entity.guid}|{hostname}|{trace.id}|{span.id}|{entity.name}|

Currently the New Relic backend can account for situations where the metadata is inserted into the "message" key, like this one:
{"message": "Log Message NR-LINKING|{entity.guid}|{hostname}|{trace.id}|{span.id}|{entity.name}|"}

but this does not account for nested situations such as this:
{"message": {"sub_message": "Hello", "error_message": "World"}}

It seems as though there is a proposal to change how the parsing happens on the NR backend.  Until then, the proposed solution is to include it as a value of the key "NR-LINKING" like this:
{"NR-LINKING": "{entity.guid}|{hostname}|{trace.id}|{span.id}|{entity.name}|"}


tests/logger_logging/test_local_decorating.py
```diff
@@ -55,7 +85,18 @@ def test_local_log_decoration_inside_transaction(logger):
...
     def test():
         exercise_logging(logger)
-        assert logger.caplog.records[0] == get_metadata_string('C', True)
+        assert logger.caplog.records[0] == get_metadata_string("C", True)
+
+    test()
+
+
+@reset_core_stats_engine()
+def test_local_log_decoration_inside_transaction_with_json(logger):
```
hmstepanek, line 94: Do we need another test where it is outside the transaction?

hmstepanek (changes requested):
I had a couple code simplification suggestions and a request about cleaning up/reducing risk in the tests a bit.

tests/logger_logging/test_local_decorating.py
```diff
@@ -31,21 +35,47 @@ def set_trace_ids():
...
-        metadata_string = "".join(('NR-LINKING|', entity_guid, '|', host, '|||Python%20Agent%20Test%20%28logger_logging%29|'))
-    formatted_string = log_message + " " + metadata_string
+        metadata_string = "".join(
+            ("NR-LINKING|", entity_guid, "|", host, "|||Python%20Agent%20Test%20%28logger_logging%29|")
+        )
+
+    # Check to see if log_message is JSON formatted
+    # If so, insert key/value as the following:
+    # {...{"NR-LINKING":"[entity.guid]|[hostname]|[trace.id]|[span.id]|[entity.name]|"}}
+    try:
```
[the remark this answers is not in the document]
lrafeei: Fair.  Done!

tests/logger_logging/test_local_decorating.py
lrafeei: I initially thought it might be redundant but in the grand scheme of all the tests we have, adding this one is a drop in the ocean.  Done!

hmstepanek (approved):
Looks good! Much easier to read now!

PR #741, opened 2023-01-14
Elasticsearch v8 support

lrafeei:
With the release of elasticsearch v8, many new methods were added and the directory structure was rearranged.

Additionally, the connection and transport functionality moved entirely to the [elastic-transport](https://github.com/elastic/elastic-transport-python) library.


newrelic/hooks/datastore_elasticsearch.py
```diff
@@ -108,42 +108,62 @@ def _nr_wrapper_Elasticsearch_method_(wrapped, instance, args, kwargs):
 _elasticsearch_client_methods = (
     ("abort_benchmark", None),
     ("benchmark", _extract_args_index),
-    ("bulk", None),
+    ("bulk", _extract_args_index),
```
hmstepanek, line 111: A lot of these don't seem right to me. I believe simply replacing the `None` with `_extract_args_index` would result in different behavior that before since the `_index_name` called by `_extract_args_index` would not return `None` but `_all`. Also some of the methods have been removed which doesn't seem right either. It seems like we should probably be doing something like the following instead to maintain backwards compatibility:
```
("bulk", _extract_args_index if v8 else None),
```

hmstepanek (changes requested):
I have some concerns about the methods that we are pulling the index out of but otherwise I think the rest looks good.

newrelic/hooks/datastore_elasticsearch.py
lrafeei: I should have explained that in the PR initially, but for some of the ones I changed, I noticed that index was still an argument in previous versions ("bulk" as an example)

newrelic/hooks/datastore_elasticsearch.py
```diff
@@ -108,42 +108,62 @@ def _nr_wrapper_Elasticsearch_method_(wrapped, instance, args, kwargs):
...
     ("list_benchmarks", _extract_args_index),
-    ("mget", None),
+    ("mget", _extract_args_index),
     ("mlt", _extract_args_index),
     ("mpercolate", _extract_args_body_index),
-    ("msearch", None),
-    ("mtermvectors", None),
+    ("msearch", _extract_args_index),
+    ("msearch_template", _extract_args_index),
+    ("mtermvectors", _extract_args_index),
```
hmstepanek, line 142: So I'm seeing function signature changes between version for example this `mtermvectors` was `def mtermvectors(self, index=None, doc_type=None, body=None, params=None):` in version `6.2.0` but was `def mtermvectors(self, body=None, index=None, doc_type=None, params=None, headers=None)` in version `7.11`. I'm also not seeing any tests to make sure we are actually pulling out the index and not something else. I only see tests to make sure the method is being wrapped which is problematic as well. This looks like a bit of a mess and it is a pre-existing mess too. *crying*

TimPansino (approved):

hmstepanek (approved):

PR #745, opened 2023-01-23
Validator transfer p3

lrafeei:
This PR is Part 3 out of 5 of the validator transfer.  The goal is to transfer all the validators into their own files in the validators directory outside of the fixtures.py file.


hmstepanek (approved):
Looks great! Thanks for continuing to plug away at this!

Issue #747, opened 2023-01-26
Add instrumentation for redis.asyncio in redis-py


GabrielCappelli:
### Is your feature request related to a problem? Please describe.
No timings are reported when using `redis.asyncio` client in redis-py >= 4.2.0-rc1

### Feature Description
Add instrumentation for the async client of py-redis that was added on 4.2.0-rc1

### Describe Alternatives
-

### Additional context
`aioredis` has been moved to `redis-py` starting on version `4.2.0-rc1`. Only redis-py will be maintained going forward.

### Priority
Please help us better understand this feature request by choosing a priority from the following options:
[Nice to Have, Really Want, Must Have, Blocker]
**Really Want**

TimPansino:
Added with v8.6.0

PR #744, opened 2023-01-23
Support `redis.asyncio`

pauk-slon:
_Before contributing, please read our [contributing guidelines](https://github.com/newrelic/newrelic-python-agent/blob/main/CONTRIBUTING.rst) and [code of conduct](https://github.com/newrelic/.github/blob/master/CODE_OF_CONDUCT.md)._

# Overview
`aioredis` is included to redis-py since version 4.2.0rc1 as `redis.asyncio` (https://github.com/redis/redis-py/pull/1899). This PR adapts the current `newrelic.hooks.datastore_aioredis` for supporting `redis.asyncio`.

PR #750, opened 2023-01-26
Redis Asyncio Testing

TimPansino:
# Overview

* Adapt `aioredis` tests to the moved version of the library located at `redis.asyncio` in the `redis-py` lib.
* Add tox envs for `redis.asyncio` separate from `aioredis` or existing `redis-py` envs to avoid as much confusing overlap as possible.
* Add `get_package_version_tuple` to replace common logic of splitting strings to tuples with safer alternative.

# Related Github Issue

Testing for #744
Related to #747

### DIFF
--- a/newrelic/common/package_version_utils.py
+++ b/newrelic/common/package_version_utils.py
@@ -20,7 +20,7 @@ NULL_VERSIONS = frozenset((None, "", "0", "0.0", "0.0.0", "0.0.0.0", (0,), (0, 0
 
 
 def get_package_version(name):
-    """Gets the version of the library.
+    """Gets the version string of the library.
     :param name: The name of library.
     :type name: str
     :return: The version of the library. Returns None if can't determine version.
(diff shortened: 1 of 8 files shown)


newrelic/common/package_version_utils.py
```diff
@@ -31,40 +31,68 @@ def get_package_version(name):
...
     version = _get_package_version(name)
 
     # Coerce iterables into a string
     if isinstance(version, tuple):
         version = ".".join(str(v) for v in version)
 
     return version
+
+
+def get_package_version_tuple(name):
```
hmstepanek, line 43: This is perfect! I've been thinking about doing about this for a while but it was one of those low hanging fruit that I didn't have time for. Thanks for doing this! I wonder if it would be more immediately helpful to open a separate PR for this directly into main so we can take advantage of this immediately rather than it existing only on a feature branch though?

newrelic/hooks/datastore_aioredis.py
```diff
@@ -21,17 +21,10 @@
...
-def get_aioredis_version():
-    try:
-        import aioredis as aioredis_legacy
-    except ModuleNotFoundError:
-        return None
-    try:
-        return tuple(int(x) for x in getattr(aioredis_legacy, "__version__").split("."))
-    except Exception:
-        return 0, 0, 0
+get_aioredis_version = lambda: get_package_version_tuple("aioredis")
```
hmstepanek, line 27: I only see that this is used in one place. Instead of making this a lambda can we just replace its usage on line 66 with the following?
```
aioredis_version = get_package_version_tuple("aioredis")
```

hmstepanek (changes requested):
Had a couple suggestions/requests otherwise looks good!

newrelic/common/package_version_utils.py
TimPansino: I'd like to merge this in today and get it into the release so why don't we just ram this through rather than trying to separate?

newrelic/hooks/datastore_aioredis.py
TimPansino: Will do

hmstepanek (approved):

PR #751, opened 2023-01-26
Add sklearn label events

umaannamalai:
This PR adds implementation and testing for sklearn label events.


newrelic/hooks/mlmodel_sklearn.py
```diff
@@ -80,12 +80,52 @@ def _nr_wrapper_method(wrapped, instance, args, kwargs):
...
 
 
+def create_label_event(transaction, _class, instance, return_val):
+    inference_id = uuid.uuid4()
+    model_name = getattr(instance, "_nr_wrapped_name", _class)
+    model_version = getattr(instance, "_nr_wrapped_version", "0.0.0")
+    label_names = getattr(instance, "_nr_wrapped_label_names", None)
+    settings = transaction.settings if transaction.settings is not None else global_settings()
+
+    if return_val is not None:
```
hmstepanek, line 97: You can just shorten this to:
```suggestion
    if return_val:
```

newrelic/hooks/mlmodel_sklearn.py
```diff
@@ -80,12 +80,52 @@ def _nr_wrapper_method(wrapped, instance, args, kwargs):
...
             wrap_predict(transaction, _class, wrapped, instance, args, kwargs)
+            create_label_event(transaction, _class, instance, return_val)
             return PredictReturnTypeProxy(return_val, model_name=_class, training_step=training_step)
         return return_val
 
     wrap_function_wrapper(module, "%s.%s" % (_class, method), _nr_wrapper_method)
 
 
+def create_label_event(transaction, _class, instance, return_val):
+    inference_id = uuid.uuid4()
```
hmstepanek, line 91: This uuid needs to be the same as the feature event uuid since they are part of the same prediction.

newrelic/hooks/mlmodel_sklearn.py
```diff
@@ -80,12 +80,52 @@ def _nr_wrapper_method(wrapped, instance, args, kwargs):
...
+    model_name = getattr(instance, "_nr_wrapped_name", _class)
+    model_version = getattr(instance, "_nr_wrapped_version", "0.0.0")
+    label_names = getattr(instance, "_nr_wrapped_label_names", None)
+    settings = transaction.settings if transaction.settings is not None else global_settings()
+
+    if return_val is not None:
+        for index, value in enumerate(return_val):
+            python_value_type = str(type(value))
+            value_type = str(categorize_data_type(python_value_type))
+            label_names_list = _get_label_names(label_names, return_val)
```
hmstepanek, line 101: We should move this call outside of the loop so it's not re-computing the names for each value.

tests/mlmodel_sklearn/test_feature_events.py
```diff
@@ -87,6 +109,9 @@ def _test():
     _test()
 
 
+label_type = "bool" if sys.version_info < (3, 8) else "numerical"
```
hmstepanek, line 112: This is interesting.

tests/mlmodel_sklearn/test_feature_events.py
```diff
@@ -18,6 +18,7 @@
 import pandas
```
hmstepanek, line 18: Perhaps we should rename this test file to `test_inference_events.py` since it tests both label and feature events now.

hmstepanek (changes requested):
Just a couple small requests otherwise looks good!

newrelic/hooks/mlmodel_sklearn.py
umaannamalai: Good call out! This was how I initially had this implemented but since `return_val` is an `ndarray`, numpy throws the following error: `The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`. To get around this, I had to explicitly call out the `None` case.

hmstepanek: Oh right! I've run into this one before-I totally forgot about it. Ndarrays and their special handling... :/

newrelic/hooks/mlmodel_sklearn.py
```diff
@@ -79,13 +79,53 @@ def _nr_wrapper_method(wrapped, instance, args, kwargs):
         # _nr_wrapped attrs that will attach model info to the data.
         if method in ("predict", "fit_predict"):
             training_step = getattr(instance, "_nr_wrapped_training_step", "Unknown")
-            wrap_predict(transaction, _class, wrapped, instance, args, kwargs)
+            inference_id = uuid.uuid4()
+            wrap_predict(transaction, _class, inference_id, instance, args, kwargs)
```
hmstepanek, line 83: I wonder if we should rename `wrap_predict` to `create_feature_event`? Just so the code is more understandable. This model wrapping code is pretty bloated so anything we can do to make it more readable would be a big help I think.

hmstepanek (approved):
Looks good. Had one minor point of feedback but it's not a huge deal either way you want to go. I'm pre-approving this one.

hmstepanek (changes requested):
This is so tricky and I'm so sorry! I've provided an example which will hopefully help explain what we're after here a little bit clearer along with some code that I used on my branch to do it.

hmstepanek (approved):
The changes look good here since last I looked at this. Looks like there's a bunch of tests failing. I presume that's the same issue I was running into on the other branch which we are already aware of so just marking this as approved and we can address it on the other PR.

hmstepanek:
Closing this as these changes are now included in https://github.com/newrelic/newrelic-python-agent/pull/715.

PR #752, opened 2023-01-26
Fix tests in starlette v0.23.1

lrafeei:
With the release of [Starlette v0.23.1](https://github.com/encode/starlette/pull/1940), the check to see if "more_body" exists is removed

### DIFF
--- a/tests/framework_starlette/test_bg_tasks.py
+++ b/tests/framework_starlette/test_bg_tasks.py
@@ -87,6 +87,8 @@ def test_basehttp_style_middleware(target_application, route):
         response = app.get("/" + route)
         assert response.status == 200
 
+    # The bug was fixed in version 0.21.0 but re-occured in 0.23.1.
+    # The bug was also not present on 0.20.1 to 0.23.1 if using Python3.7.
     BUG_COMPLETELY_FIXED = (0, 21, 0) <= starlette_version < (0, 23, 1) or (
         (0, 20, 1) <= starlette_version < (0, 23, 1) and sys.version_info[:2] > (3, 7)
     )
--- a/tests/framework_starlette/test_bg_tasks.py
+++ b/tests/framework_starlette/test_bg_tasks.py
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys
-
 import pytest
 from starlette import __version__
 from testing_support.validators.validate_transaction_count import (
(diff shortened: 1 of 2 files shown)


tests/framework_starlette/test_bg_tasks.py
```diff
@@ -87,10 +87,10 @@ def _test():
         response = app.get("/" + route)
         assert response.status == 200
 
-    BUG_COMPLETELY_FIXED = (starlette_version >= (0, 21, 0)) or (
-        starlette_version >= (0, 20, 1) and sys.version_info[:2] > (3, 7)
+    BUG_COMPLETELY_FIXED = (0, 21, 0) <= starlette_version < (0, 23, 1) or (
```
hmstepanek, line 90: I'm so confused and it's not your fault. lol I think adding a comment here might be helpful:
```suggestion
    # The bug was fixed in version 0.21.0 but re-occured in 0.23.1.
    # The bug was also not present on 0.20.1 to 0.23.1 if using Python3.7.
    BUG_COMPLETELY_FIXED = (0, 21, 0) <= starlette_version < (0, 23, 1) or (
```

tests/framework_starlette/test_bg_tasks.py
```diff
@@ -87,10 +87,10 @@ def _test():
...
+        (0, 20, 1) <= starlette_version < (0, 23, 1) and sys.version_info[:2] > (3, 7)
     )
-    BUG_PARTIALLY_FIXED = (0, 20, 1) <= starlette_version < (0, 21, 0) and sys.version_info[:2] <= (3, 7)
+    BUG_PARTIALLY_FIXED = (0, 20, 1) <= starlette_version < (0, 21, 0) or starlette_version >= (0, 23, 1)
```
hmstepanek, line 93: This doesn't seem to be backwards compatible with the previous logic. Is there a good reason for that or is there a mistake here?

hmstepanek (changes requested):
Thanks for doing this! This logic just get's more complicated all the time. lol I had one concern about the logic of the partial bug and I suggested adding an explanation comment since the logic is more complicated and difficult to follow. Let me know what you think!

tests/framework_starlette/test_bg_tasks.py
lrafeei: I realized the equal-to-or-greater-than Python 3.7 check is not necessary for our tests anymore since we removed testing for Python 3.6.

tests/framework_starlette/test_bg_tasks.py
lrafeei: Done!

PR #754, opened 2023-01-27
Report feature event w/o value

hmstepanek:
# Overview
Report the feature event w/o the raw value if `machine_learning.inference_event_value.enabled` is False.

### DIFF
--- a/tests/mlmodel_sklearn/test_feature_events.py
+++ b/tests/mlmodel_sklearn/test_feature_events.py
@@ -18,6 +18,7 @@ import numpy as np
 import pandas
 from _validate_custom_events import validate_custom_events
 from testing_support.fixtures import (
+    override_application_settings,
     reset_core_stats_engine,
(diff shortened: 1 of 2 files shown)


tests/mlmodel_sklearn/test_feature_events.py
```diff
@@ -412,3 +413,47 @@ def _test():
...
+        }
+    },
+]
+
+
+@reset_core_stats_engine()
+@override_application_settings({"machine_learning.inference_event_value.enabled": False})
+def test_does_not_include_value_when_inference_event_value_enabled_is_false():
+    @validate_custom_events(numpy_str_recorded_custom_events_no_value)
+    @validate_custom_event_count(count=2)
```
umaannamalai, line 444: Just noting that I will update this test once the label event PR is merged since this test will then need to account for the label events in the overall expected count.

umaannamalai (approved):

PR #758, opened 2023-02-02
Infinite Tracing Compression

lrafeei:
This PR has GZIP compression implemented for Infinite Tracing

### DIFF
--- a/newrelic/core/agent_streaming.py
+++ b/newrelic/core/agent_streaming.py
@@ -15,8 +15,6 @@
 import logging
 import threading
 
-from newrelic.core.config import global_settings_dump
-
 try:
     import grpc
 
@@ -59,8 +57,7 @@ class StreamingRpc(object):
         self.notify = self.condition()
         self.record_metric = record_metric
         self.closed = False
-        if compression is None:
-            compression = global_settings_dump()["infinite_tracing.compression"]
+        # If this is not set, None is still a falsy value.
         self.compression_setting = grpc.Compression.Gzip if compression else grpc.Compression.NoCompression
 
         self.create_channel()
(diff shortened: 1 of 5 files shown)


newrelic/core/agent_streaming.py
```diff
@@ -44,7 +44,7 @@ class StreamingRpc(object):
     )
     OPTIONS = [("grpc.enable_retries", 0)]
 
-    def __init__(self, endpoint, stream_buffer, metadata, record_metric, ssl=True):
+    def __init__(self, endpoint, stream_buffer, metadata, record_metric, ssl=True, compression=True):
```
TimPansino, line 47: Let's default the StreamingRPC to False, let the agent settings handle the default value for actual applications. This way testing behavior doesn't change when manually creating and testing the StreamingRPC.

TimPansino (changes requested):
Let's add a test for this to check that the correct flag is set on the grpc channel object when pulled from the settings, and let's add a test that asserts the spans are sent and deserialized correctly with compression.

To do the first one, the tests in `test_infinite_tracing` that use `override_generic_settings` could be used as a guide. Connect to the data collector, pull the rpc off the application's active session, and pull the channel off the RPC to check the compression attribute (that I assume it has).

To do the second one, how about we parametrize the `test_rpc_serialization_and_deserialization` test from my PR (#759) with compression set to True and False, and apply the flag to the StreamingRpc that's manually created in it. That way it'll be tested with batching enabled and disabled and properly deserialized.

newrelic/core/agent_streaming.py
lrafeei: I can do that--Any thoughts on setting it to `None` instead and then using `global_settings_dump()` to set the value if nothing has been passed in?

TimPansino (approved):

PR #759, opened 2023-02-03
Infinite Tracing Batching

TimPansino:
# Overview
* Adds batching implementation to infinite tracing.
* Updates pb2 files to include support for RecordSpanBatch.
* Adds testing for StreamBufferIterator batching.
* Adds testing for infinite tracing supportability metrics.
  * (`Supportability/InfiniteTracing/Span/Sent`, `Supportability/InfiniteTracing/Span/Seen`)
* Adds regression testing for lost data on reconnect with infinite tracing.


newrelic/common/streaming_utils.py
```diff
@@ -100,12 +113,30 @@ def __next__(self):
...
+                if self.batching:
+                    stream_buffer_len = len(self.stream_buffer)
+                    if stream_buffer_len > self.MAX_BATCH_SIZE:
+                        # Ensure batch size is never more than 100 to prevent issues with serializing large numbers
+                        # of spans causing their age to exceed 10 seconds. That would cause them to be rejected
+                        # by the trace observer.
+                        batch = [self.stream_buffer._queue.popleft() for _ in range(self.MAX_BATCH_SIZE)]
+                        return SpanBatch(spans=batch)
+                    elif stream_buffer_len:
+                        # For small span batches empty stream buffer into list and clear queue.
```
hmstepanek, line 125: This mentions needing to use a lock. Where is the lock?

hmstepanek (approved):
I had one question/concern about lock mentioned in the comment cause I didn't see a lock being used. Otherwise looks good. I'm just gonna mark it as approved and if it turns out the lock is an issue we can fix it and I'll re-review.

newrelic/common/streaming_utils.py
TimPansino: The very first line of the function is the lock. 

`with self._notify:`

PR #765, opened 2023-02-10
Refactor Code Coverage

TimPansino:
# Overview

* Reorder code coverage fixture to always start before agent is initialized. Should provide much better coverage numbers.


hmstepanek (approved):
Looks good.

PR #779, opened 2023-03-14
Reduce number of spans in django framework

hmstepanek:
# Overview
This removes the following middlewares from being instrumented in an effort to reduce the number of spans created when instrumenting django:
* CsrfViewMiddleware
* XFrameOptionsMiddleware
* MessageMiddleware
* CsrfViewMiddleware
* CommonMiddleware
* SecurityMiddleware

# Related Github Issue
[Spike DT Improvements](https://issues.newrelic.com/browse/NR-91358)

### DIFF
--- a/newrelic/hooks/framework_django.py
+++ b/newrelic/hooks/framework_django.py
@@ -12,48 +12,71 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import functools
+import logging
 import sys
 import threading
-import logging
-import functools
-
-from newrelic.packages import six
 
 from newrelic.api.application import register_application
 from newrelic.api.background_task import BackgroundTaskWrapper
 from newrelic.api.error_trace import wrap_error_trace
-from newrelic.api.function_trace import (FunctionTrace, wrap_function_trace,
-        FunctionTraceWrapper)
+from newrelic.api.function_trace import (
+    FunctionTrace,
+    FunctionTraceWrapper,
+    wrap_function_trace,
+)
 from newrelic.api.html_insertion import insert_html_snippet
-from newrelic.api.transaction import current_transaction
 from newrelic.api.time_trace import notice_error
+from newrelic.api.transaction import current_transaction
 from newrelic.api.transaction_name import wrap_transaction_name
 from newrelic.api.wsgi_application import WSGIApplicationWrapper
-
-from newrelic.common.object_wrapper import (FunctionWrapper, wrap_in_function,
-        wrap_post_function, wrap_function_wrapper, function_wrapper)
+from newrelic.common.coroutine import is_asyncio_coroutine, is_coroutine_function
 from newrelic.common.object_names import callable_name
+from newrelic.common.object_wrapper import (
+    FunctionWrapper,
+    function_wrapper,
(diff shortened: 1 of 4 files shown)


newrelic/hooks/framework_django.py
```diff
@@ -12,48 +12,71 @@
...
 if six.PY3:
     from newrelic.hooks.framework_django_py3 import (
-        _nr_wrapper_BaseHandler_get_response_async_,
         _nr_wrap_converted_middleware_async_,
+        _nr_wrapper_BaseHandler_get_response_async_,
     )
 
+# These middleware are not useful to wrap as they don't do anything particularly
+# interesting that could cause performance issues.
+MIDDLEWARE_DENY_WRAP = [
```
umaannamalai, line 54: In situations like this in the agent, we would typically make this into a `frozenset()` for performance reasons.

tests/framework_django/test_application.py
```diff
@@ -12,297 +12,259 @@
...
-            'CsrfViewMiddleware.process_response'), 1),
-    (('Function/django.contrib.sessions.middleware:'
-            'SessionMiddleware.process_response'), 1),
-    (('Function/django.middleware.common:'
-            'CommonMiddleware.process_response'), 1),
-    (('Function/django.middleware.gzip:'
-            'GZipMiddleware.process_response'), 1),
-    (('Function/newrelic.hooks.framework_django:'
-            'browser_timing_insertion'), 1),
+    (("Function/django.middleware.common:" "CommonMiddleware.process_request"), None),
```
umaannamalai, line 49: Is there an advantage to leaving the metrics in for the ignored middlewares with a count of `None` vs dropping them from the scoped metric list?

newrelic/hooks/framework_django.py
```diff
@@ -12,48 +12,71 @@
...
+    "django.middleware.csrf:CsrfViewMiddleware",
+    "django.middleware.clickjacking:XFrameOptionsMiddleware",
+    "django.contrib.messages.middleware:MessageMiddleware",
+    "django.middleware.csrf:CsrfViewMiddleware",
+    "django.middleware.common:CommonMiddleware",
+    "django.middleware.security:SecurityMiddleware",
```
umaannamalai, line 60: Based on the warning listed for security middleware [here,](https://docs.djangoproject.com/en/4.1/ref/middleware/#module-django.middleware.security) could it be worth it to keep the `SecurityMiddleware` spans around to support potential future vulnerability detection?

tests/framework_django/test_application.py
hmstepanek: I did this specifically to make sure they were dropped because if you simply don't look for them they could still be there but the test wouldn't catch it.

newrelic/hooks/framework_django.py
hmstepanek: That's good point. Since this middleware just sets security related headers, for vuln detection I imagine we would want to capture those headers. I don't think we would specifically need to hook into this middleware to capture the security related headers though. You should be able to capture these headers on the response request independent of wrapping this particular middleware.

umaannamalai (approved):
Looks great!

PR #780, opened 2023-03-16
Reduce number of spans in Starlette

hmstepanek:
# Overview
This removes the following middleware from being instrumented in an effort to reduce the number of spans created when instrumenting starlette:
* `ServerErrorMiddleware.__call__`

# Related Github Issue
[Spike DT Improvements](https://issues.newrelic.com/browse/NR-91358)


hmstepanek:
Both of the middleware we were planning on no longer wrapping are used to create the transaction name so I'm closing this PR for now.

Issue #748, opened 2023-01-25
[Python] - Lambda Timeout - GraphQL Introspection Query 


Mainian:
## Description

Not sure if this issue belongs here or in [newrelic/newrelic-python-agent](https://github.com/newrelic/newrelic-python-agent)

After processing a [Graphql Introspection query](https://github.com/graphql-python/graphql-core/blob/main/docs/usage/introspection.rst), the Lambda Function times out and telemetry data is not forwarded to New Relic. 

## Steps to Reproduce

With a Lambda Function running Graphql Server, make a graphql introspection query. The function will return a response, but the Lambda will not be available to process another request until it's timeout duration is reached.

Example graphql request:

```
{
  __schema {
    types {
      name
    }
  }
}
```

## Expected Behaviour

Lambda function does not timeout and the New Relic Agent forwards telemetry data.

## Relevant Logs / Console output

Sample log file:

```
2023-01-27T11:59:01.120-05:00 | START RequestId: 17026805-857a-4683-936e-f4b7fede3ea5 Version: 60
2023-01-27T11:59:01.618-05:00 | [NR_EXT] Sent 1/1 New Relic function log batches successfully in 45.202ms (41ms to transmit 9.2kB).

...

2023-01-27T12:00:00.973-05:00 | END RequestId: 17026805-857a-4683-936e-f4b7fede3ea5
2023-01-27T12:00:00.973-05:00 | REPORT RequestId: 17026805-857a-4683-936e-f4b7fede3ea5 Duration: 59851.68 ms Billed Duration: 59852 ms Memory Size: 1536 MB Max Memory Used: 219 MB Init Duration: 3642.79 ms
```

## Your Environment

AWS Lambda - Python 3.8
New Relic Lambda Layer 

## Additional context

The Python New Relic agent [explicitly ignores](https://github.com/newrelic/newrelic-python-agent/blob/0db5fee1e5d44b0791dc517ac9f5d88d1240a340/newrelic/hooks/framework_graphql.py#L137) GraphQL introspection queries transactions.

Un-ignoring the transaction does not cause the Lambda Function to timeout and telemetry data is forwarded:

```
        transaction = newrelic.agent.current_transaction(active_only=False)
        transaction.ignore_transaction = False
```

Mainian:
Issue #280 and it's [pull request](https://github.com/newrelic/newrelic-python-agent/pull/290) introduced this issue into the Lambda Layer. 

@TimPansino 
[...]

PR #783, opened 2023-03-20
Add GraphQL Introspection Setting

TimPansino:
# Overview

* Add a setting to enable capturing of GraphQL introspection queries. False by default, retaining previous behavior.
* Adjust introspection query detection to only ignore queries that **ONLY** contain introspection, allowing mixed queries to be traced.

# Related Github Issue

Relevant to #748


hmstepanek (approved):
Looks good. Thanks for putting this together so quickly!

PR #787, opened 2023-03-21
Guard GraphQL Settings Lookup

### LINKED ISSUE #786: NoneType object has no attribute settings
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

**Description**
[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT share sensitive information, whether personal, proprietary, or otherwise! )

This broke our production build. Not every, but lots of requests failed due to newrelic framework_graphql

AttributeError: 'NoneType' object has no attribute 'settings'
  File "promise/promise.py", line 87, in try_catch
    return (handler(*args, **kwargs), None)
  File "graphql/execution/executor.py", line 127, in promise_executor
    return execute_operation(exe_context, exe_context.operation, root_value)
  File "newrelic/hooks/framework_graphql.py", line 108, in wrap_execute_operation
    settings = transaction.settings

**Expected Behavior**
[NOTE]: # ( Tell us what you expected to happen. )

It doesn't crash :)

**[Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results**
[NOTE]: # ( Provide any other relevant log data. )
[TIP]:  # ( Scrub logs and diagnostic information for sensitive information )

**Steps to Reproduce**
[NOTE]: # ( Please be as specific as possible. )
[TIP]:  # ( Link a sample application that demonstrates the issue. )
Any graphql query

**Your Environment**
[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of New Relic software and any relevant configurations. )

newrelic 8.7.1 (This did NOT happen on 8.7.0)
graphene 2.1.6

**Additional context**
[TIP]:  # ( Add any other context about the problem here. For example, relevant community posts or support tickets. )

I know We're on an old version of graphene,  but still, one would hope patch version upgrades don't have breaking changes 🤣

TimPansino:
# Overview

* A bug was introduced in GraphQL instrumentation when not running within a transaction in v8.7.1.
* Adds a settings lookup guard to address this crash.

# Related Github Issue

Closes #786

### DIFF
--- a/newrelic/hooks/framework_graphql.py
+++ b/newrelic/hooks/framework_graphql.py
@@ -105,7 +105,6 @@ def bind_operation_v2(exe_context, operation, root_value):
 def wrap_execute_operation(wrapped, instance, args, kwargs):
     transaction = current_transaction()
     trace = current_trace()
-    settings = transaction.settings
 
     if not transaction:
         return wrapped(*args, **kwargs)
@@ -136,7 +135,8 @@ def wrap_execute_operation(wrapped, instance, args, kwargs):
     if operation.selection_set is not None:
         fields = operation.selection_set.selections
         # Ignore transactions for introspection queries
-        if not settings.instrumentation.graphql.capture_introspection_queries:
+        capture_introspection_queries = transaction and transaction.settings and transaction.settings.instrumentation and transaction.settings.instrumentation.graphql and transaction.settings.instrumentation.graphql.capture_introspection_queries
+        if not capture_introspection_queries:
             # If all selected fields are introspection fields
             if all(get_node_value(field, "name") in GRAPHQL_INTROSPECTION_FIELDS for field in fields):
                 ignore_transaction()
--- a/newrelic/hooks/framework_graphql.py
+++ b/newrelic/hooks/framework_graphql.py
@@ -135,14 +135,7 @@ def wrap_execute_operation(wrapped, instance, args, kwargs):
     if operation.selection_set is not None:
         fields = operation.selection_set.selections
         # Ignore transactions for introspection queries
-        capture_introspection_queries = (
-            transaction
-            and transaction.settings
-            and transaction.settings.instrumentation
-            and transaction.settings.instrumentation.graphql
-            and transaction.settings.instrumentation.graphql.capture_introspection_queries
-        )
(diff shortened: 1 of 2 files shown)


newrelic/hooks/framework_graphql.py
```diff
@@ -136,7 +135,14 @@ def wrap_execute_operation(wrapped, instance, args, kwargs):
     if operation.selection_set is not None:
         fields = operation.selection_set.selections
         # Ignore transactions for introspection queries
-        if not settings.instrumentation.graphql.capture_introspection_queries:
+        capture_introspection_queries = (
```
hmstepanek, line 138: Can't this just be left as it was before since the transaction we know must exist now if it got this far? Looking at other places in the code we don't seem to guard against the settings path not existing like this so wondering if this is really needed.
```
if not transaction.settings.instrumentation.graphql.capture_introspection_queries:
```

newrelic/hooks/framework_graphql.py
```diff
@@ -105,7 +105,6 @@ def bind_operation_v2(exe_context, operation, root_value):
 def wrap_execute_operation(wrapped, instance, args, kwargs):
     transaction = current_transaction()
     trace = current_trace()
```
hmstepanek, line 107: Do we need a no harm test for this case where we are not in a transaction?

hmstepanek (changes requested):
I had a couple suggestions here. I understand wanting to get this out ASAP so I'm cool if the writing of the test would be too much work and we just wanna skip it for now.

newrelic/hooks/framework_graphql.py
TimPansino: We do sometimes and sometimes not, it's a mixed bag of style across the agent. I'm not sure it's any safer as long as the top level settings exists so maybe we can leave it at that.

newrelic/hooks/framework_graphql.py
TimPansino: Added one.

hmstepanek (approved):

PR #789, opened 2023-03-28
User Tracking for Errors Inbox

umaannamalai:
This PR adds implementation and testing for the new `set_user_id` API to be used for user tracking with the errors inbox.


tests/agent_features/test_attributes_in_action.py
```diff
@@ -938,10 +939,27 @@ def test_none_type_routing_key_agent_attribute():
...
+@validate_attributes("agent", [], _forgone_agent_attributes)
+def test_enduser_id_attribute_api_outside_txn():
+    """Validate enduser.id attribute set by API ends up on transaction events, error events, and traced errors."""
+    set_user_id("1234")
+
+    try:
+        raise ValueError()
+    except Exception:
+        application_instance = application()
+        application_instance.notice_error()
```
TimPansino, line 964: Is there a point to running this test outside a transaction with these validators? If we don't expect any data, should this just be a no harm test that only looks for a crash?

newrelic/api/transaction.py
```diff
@@ -1813,6 +1813,13 @@ def add_custom_parameters(items):
     return add_custom_attributes(items)
 
 
+def set_user_id(user_id):
+    transaction = current_transaction()
+    if not user_id or not transaction:
+        return
+
+    transaction._add_agent_attribute("enduser.id", user_id)
```
TimPansino, line 1821: Do we need to process this with the attribute system to ensure the attribute gets properly truncated? Can we add a truncation test to check? Maybe add that to the parametrization matrix.

TimPansino (changes requested):

TimPansino (approved):
🚀

PR #798, opened 2023-04-11
Fix Loguru Instrumentation for v0.7.0

TimPansino:
# Overview

* Signatures changed in loguru v0.7.0 breaking local decorating and call stack adjustments. Fixed with auto-signature implementation.

### DIFF
--- a/newrelic/common/signature.py
+++ b/newrelic/common/signature.py
@@ -17,9 +17,9 @@ from newrelic.packages import six
 if six.PY3:
     from inspect import Signature
 
-    def bind_args(func, args, kwargs, unwrap=False):
+    def bind_args(func, args, kwargs):
         """Bind arguments and apply defaults to missing arugments for a callable."""
-        bound_args = Signature.from_callable(func, follow_wrapped=unwrap).bind(*args, **kwargs)
+        bound_args = Signature.from_callable(func).bind(*args, **kwargs)
         bound_args.apply_defaults()
         return bound_args.arguments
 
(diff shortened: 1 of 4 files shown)


newrelic/common/signature.py
```diff
@@ -0,0 +1,44 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from newrelic.packages import six
+
+if six.PY3:
+    from inspect import Signature
+
+    def bind_args(func, args, kwargs, unwrap=False):
```
hmstepanek, line 20: If I'm remembering correctly, the follow_wrapped didn't work for us with the decorators in elasticsearch. While I realize not all decorators manipulate parameters the way es did, I think it would be safer for future developers of the agent if we just don't support wrapping at all for these auto sig functions. That way future devs won't be tempted to use these functions in a potentially unsafe manor. It might also be good to capture that in a comment somewhere in this file. Perhaps in a doc string of the function?

If you feel strongly that we should keep the support for setting follow_wrapped, we absolutely should add a warning about this in the doc string for future devs. I'd argue that since we don't have a current use case for utilizing follow_wrapped though, that it would be better to simply just not support it right now and have the discussion later if we find a case where we would like to use it.

hmstepanek (changes requested):

lrafeei (approved):

lrafeei (approved):

newrelic/common/signature.py
```diff
@@ -0,0 +1,31 @@
...
```
hmstepanek, line 20: It looks like unwrap arg is still here even though it was removed in the other implementation and the doc string.

hmstepanek (approved):
Seems like we should remove the unwrap arg from the function def also. Just gonna pre-approve this.

PR #802, opened 2023-04-25
Add new ML event type

hmstepanek:
# Overview
Add new ML Event data type.


newrelic/core/stats_engine.py
```diff
@@ -1489,6 +1519,13 @@ def reset_custom_events(self):
         else:
             self._custom_events = SampledDataSet()
 
+    def reset_ml_events(self):
+        if self.__settings is not None:
+            # TODO fix this with the actual setting
+            self._ml_events = SampledDataSet(8333)
```
lrafeei, line 1525: Should this be 100000 or 8333?  Follow up question: if this 8333 is correct, should line 59 in newrelic.core.config also be 8333?

hmstepanek: It should be 8333 since it will be on the fast harvest but this should be replaced in the next PR with the actual setting instead of a hardcoded number anyway. That computation will happen behind the scenes. Users will set the value to 100,000 and it will get converted into the 8333 value internally based on the fast harvest cycle.

lrafeei: Gotcha!

lrafeei (approved):

PR #803, opened 2023-04-27
Pin virtualenv, fix pip arg deprecation & disable kafka tests

hmstepanek:
# Overview
Pin virtualenv to the last version before they stopped supporting Python 2.7 (f strings are now used, which are only supported in Python 3)
Pin pip because install-options was deprecated (we use this to install pycurl with openssl).

### DIFF
--- a/.github/actions/setup-python-matrix/action.yml
+++ b/.github/actions/setup-python-matrix/action.yml
@@ -47,4 +47,4 @@ runs:
       shell: bash
       run: |
         python3.10 -m pip install -U pip
-        python3.10 -m pip install -U wheel setuptools 'tox<4' virtualenv!=20.0.24
+        python3.10 -m pip install -U wheel setuptools 'tox<4' 'virtualenv<20.21.1'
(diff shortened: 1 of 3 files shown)


.github/actions/setup-python-matrix/action.yml
```diff
@@ -47,4 +47,4 @@ runs:
       shell: bash
       run: |
         python3.10 -m pip install -U pip
-        python3.10 -m pip install -U wheel setuptools 'tox<4' virtualenv!=20.0.24
+        python3.10 -m pip install -U wheel setuptools 'tox<4' 'virtualenv<20.21.1'
```
umaannamalai, line 50: It looks like v20.21.1 still supported Python 2.7 (https://github.com/pypa/virtualenv/blob/20.21.1/.github/workflows/check.yml#L66) so we can update this to be virtualenv<[20.22.0](https://github.com/pypa/virtualenv/releases/tag/20.22.0).

umaannamalai (approved):

PR #809, opened 2023-05-02
Instrument RedisCluster

hmstepanek:
# Overview
RedisCluster was added in ~2021 and is a new client that we need to instrument. Because we chose to instrument at a higher level, when RedisCluster was added it inherited commands from lower level classes below our instrumentation and we missed it.

### DIFF
--- a/newrelic/config.py
+++ b/newrelic/config.py
@@ -2854,10 +2854,6 @@ def _process_module_builtin_defaults():
     )
 
     _process_module_definition(
-        "redis.commands.cluster", "newrelic.hooks.datastore_redis", "instrument_redis_commands_cluster"
-    )
-
-    _process_module_definition(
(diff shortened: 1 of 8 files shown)


newrelic/config.py
```diff
@@ -2837,6 +2837,14 @@ def _process_module_builtin_defaults():
     )
     _process_module_definition("redis.client", "newrelic.hooks.datastore_redis", "instrument_redis_client")
 
+    _process_module_definition(
+        "redis.commands.cluster", "newrelic.hooks.datastore_redis", "instrument_redis_commands_cluster"
+    )
+
+    _process_module_definition(
```
inmar-jwheeler, line 2844: Why call this twice?

lrafeei (approved):
🚀

PR #810, opened 2023-05-02
Omit some frameworks from coverage analysis

lrafeei:
This PR adds files to omit for coverage analysis.

- newrelic/hooks/adapter_meinheld.py: This instrumentation is done on C code.  In 1 month, this falls outside our support window.
- newrelic/hooks/adapter_flup.py: This is 2 years outside our support window.
- newrelic/hooks/component_piston.py: The instrumentation for this has not been updated.  Either we drop this or we have to reinstrument this library (which is mostly not written in Python)
- newrelic/hooks/datastore_pyelasticsearch.py: This is 2 years outside our support window.
- newrelic/hooks/external_pywapi.py: This is 10 years outside our support window.
- newrelic/hooks/external_dropbox.py: The instrumentation for this has not been updated.  Either we drop this or we have to reinstrument this library.
- newrelic/hooks/external_facepy.py: This is a month outside our support window.
- newrelic/hooks/external_xmlrpclib.py: Now called xmlrpc for Python 3, this specific instrumentation is on a framework that is not within our support window
- newrelic/hooks/framework_pylons.py: This is 2 years outside our support window.
- newrelic/hooks/framework_web2py.py: This is a month outside our support window.
- newrelic/hooks/middleware_weberror.py: This is 4 years outside our support window.

If there is a subsequent release of any of these libraries, we will circle back and provide adequate testing for these frameworks.


hmstepanek (approved):
I just removed the commented out omit line but everything else looks good!

PR #811, opened 2023-05-02
Add machine learning and ml_event config options

hmstepanek:
# Overview
Add the following configuration settings for mlops:
* `machine_learning.enabled`
* `machine_learning.inference_events.value.enabled`
* `event_harvest_config.harvest_limits.ml_event_data`
* `ml_insights_events.enabled`


umaannamalai (approved):
Looks good! We can iterate on these settings in the future based on whether we use OTLP events more broadly going forward.

PR #813, opened 2023-05-04
Run coverage around pytest

hmstepanek:
# Overview
Run code coverage around pytest instead of within pytest. Running codecoverage within pytest means we miss import time instrumentation coverage.


lrafeei (approved):
Just for future reference, we are removing memcache tests because this framework has been in maintenance mode since 2017.  Pymemcache is the recommended library to use for this.

PR #815, opened 2023-05-11
Dimensional Metrics (WIP)

TimPansino:
# Overview

* Add OTLP dimensional metrics APIs.
* Does NOT implement serialization or sending yet. Currently dumps raw payloads to a global list for testing purposes.


hmstepanek (changes requested):
Overall looks really good! This was a huge undertaking!

I was a little worried about using a frozenset for the metric identity because I was afraid the tags might get really long and impact performance. That being said I think all the places we use it to compare with == we are comparing with the hash of the frozenset so I think that's fine. I just wanted to note it here in case I missed some comparison in the code where this might be problematic.

I noted a couple areas in the code where I had general questions/don't know this part of our code well enough to quite follow what's going on so perhaps we can discuss these areas on Monday before I leave for vacation.

hmstepanek (approved):

PR #817, opened 2023-05-25
Cache _get_package_version to avoid heavy CPU load

SlavaSkvortsov:
# Overview
Changed introduced in [this](https://github.com/newrelic/newrelic-python-agent/pull/750) PR was a reason why our service ran out of CPU after the NR update. 
image

If you have [redis-py](https://github.com/redis/redis-py) installed, but don't have [aioredis](https://github.com/aio-libs/aioredis-py), NewRelic tries to find a version of aioredis anyway every time you call your Redis client. Instead of a quick check based on importing the library, NR now uses pkg_resources to find a lib. And pkg_resources is slow :( 
Because we use Redis a lot, it heavily affected the performance. 

The easiest solution is to add a global cache for package versions - they won't change in runtime anyway 

# Related Github Issue

# Testing
exercise your changes.

### DIFF
--- a/newrelic/common/package_version_utils.py
+++ b/newrelic/common/package_version_utils.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import sys
+from functools import lru_cache
 
 # Need to account for 4 possible variations of version declaration specified in (rejected) PEP 396
 VERSION_ATTRS = ("__version__", "version", "__version_tuple__", "version_tuple")  # nosec
@@ -67,6 +68,7 @@ def get_package_version_tuple(name):
     return version
 
 
+@lru_cache
 def _get_package_version(name):
     module = sys.modules.get(name, None)
     version = None


hmstepanek (changes requested):
Thank you for the bringing this to our attention and submitting a fix! Since we still support Python2.7 and lru_cache is only available in Python3, we need a solution for Python2.7. It looks like the tests are failing for Python3.7 as well though likely that could be fixed with some configuration changes. It looks like there are a couple options if we want to just pull in a third party for Python2.7:
* [functools32](https://pypi.org/project/functools32/)
* [repoze.lru](https://pypi.org/project/repoze.lru/)
Another option is implementing our own caching. I've only briefly started looking into this and our team will likely want to decide a direction together on this. I just wanted to provide some initial feedback here so you don't feel like we're ignoring you!

SlavaSkvortsov:
Hey @hmstepanek! 

Want to let you know that we spot more issues with this code. It performs poorly with [rq](https://python-rq.org/) library, since every workhorse there is a separate process, having lru_cache does not help. And getting the versions of the packages is something that happens on the initialization of NewRelic. An average job looks like this:
image
first 400 milliseconds is the initialization of NewRelic (you can see that the function is called `cached_get_package_version` coz we monkeypatch the library to avoid the issue described in the original issue) and only after that the actual job start. When you have a lot of small jobs, spending 400 milliseconds of CPU for every single one of them is significant. 

We don't have a solution for this issue yet. 

Here is some code to monkeypatch the lib to solve the issue described in the issue above in case someone needs it :) 
```python3
def _monkeypatch_newrelic() -> None:
    try:
        from newrelic.common import package_version_utils
    except ImportError:  # pragma: nocover
        return  # pragma: nocover

    if not hasattr(package_version_utils, '_get_package_version'):
        return  # pragma: nocover

    original_get_package_version = package_version_utils._get_package_version

    @cache
    def cached_get_package_version(package_name: str) -> str:
        return original_get_package_version(package_name)

    package_version_utils._get_package_version = cached_get_package_version
```

About the solution compatible with python2.7 - I don't have any experience with that, so I can't recommend anything. Thanks for looking into the problem and feel free to change the PR! 

Thank you!

TimPansino:
Closing in favor of #946 which incorporates and extends these changes.

(This branch got really out of date and I couldn't update it here.)

PR #819, opened 2023-06-01
Changes for NR CSEC module support

AnupamJuniwal:
## Overview
The changes include all the necessary modifications to APM codebase to invoke and control NR CSEC module.

## Related Github Issue
- none

### Automated test cases added
- none

### Manual test cases run
- Verified the APM agent is loading CSEC agent correctly
- Verified the APM agent reads the required config by CSEC agent correctly.
- Verified the integration to be running with CSEC agent as expected.

### DIFF
--- a/newrelic/config.py
+++ b/newrelic/config.py
@@ -329,6 +329,13 @@ def _process_configuration(section):
     _process_setting(section, "ca_bundle_path", "get", None)
     _process_setting(section, "audit_log_file", "get", None)
     _process_setting(section, "monitor_mode", "getboolean", None)
+    _process_setting(section, "security.agent.enabled", "getboolean", None)
+    _process_setting(section, "security.enabled", "getboolean", None)
+    _process_setting(section, "security.mode", "get", None)
+    _process_setting(section, "security.validator_service_url", "get", None)
+    _process_setting(section, "security.detection.rci.enabled", "getboolean", None)
+    _process_setting(section, "security.detection.rxss.enabled", "getboolean", None)
+    _process_setting(section, "security.detection.deserialization.enabled", "getboolean", None)
     _process_setting(section, "developer_mode", "getboolean", None)
     _process_setting(section, "high_security", "getboolean", None)
     _process_setting(section, "capture_params", "getboolean", None)
@@ -3169,6 +3176,22 @@ def _setup_agent_console():
         newrelic.core.agent.Agent.run_on_startup(_startup_agent_console)
 
 
+def _setup_security_module():
+    """Initiates k2 security module and adds a
+    callback to agent startup to propagate NR config
+    """
+    try:
+        if not _settings.security.agent.enabled:
+            return
+        from newrelic_security.api.agent import Agent as SecurityAgent
+        # initialize security agent
+        security_agent = SecurityAgent()
+         # create a callback to reinitialise the security module
+        newrelic.core.agent.Agent.run_on_startup(security_agent.refresh_agent)
+    except Exception as k2error:
+        _logger.error("K2 Startup failed with error %s", k2error)
+
+
 def initialize(
     config_file=None,
     environment=None,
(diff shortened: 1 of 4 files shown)


newrelic/config.py
```diff
@@ -3169,6 +3176,23 @@ def _setup_agent_console():
...
+            return
+        from newrelic_security.api.agent import Agent as SecurityAgent
+
+        # initialize security agent
+        security_agent = SecurityAgent()
+        # create a callback to reinitialise the security module
+        newrelic.core.agent.Agent.run_on_startup(security_agent.refresh_agent)
+    except Exception as k2error:
+        _logger.error("K2 Startup failed with error %s", k2error)
+
```
TimPansino, line 3194:
```suggestion
def _setup_security_module():
    def _security_agent_registration_callback():
        """Callback that initializes k2 security module and propagates NR config."""
        try:
            from newrelic.api.application import application_instance
            application = application_instance(activate=False)
            if not application or not application.settings or not application.settings.security.agent.enabled:
                return
            
            from newrelic_security.api.agent import Agent as SecurityAgent
            
            # initialize security agent
            security_agent = SecurityAgent()
            security_agent.refresh_agent()

        except Exception as k2error:
            _logger.error("K2 Startup failed with error %s", k2error)

    # create a callback to only initialize the security module after high security settings have been applied
    newrelic.core.agent.Agent.run_on_registration(_security_agent_registration_callback)
```

TimPansino: For compliance I think we can't run any of this until high security mode has a chance to disable it. We'll need to wait until after connect and registration, and also use application instance settings instead of the global settings (which don't include server side fixups like high security).

I think this callback should do the trick, but that depends on if

1. Your agent is a singleton and can be called like this multiple times with no issues.
2. Your agent can handle being called this late into the import process.

TimPansino: We should DEFINITELY add tests for this, somehow.

TimPansino (changes requested):

newrelic/config.py
```diff
@@ -3169,6 +3176,23 @@ def _setup_agent_console():
         newrelic.core.agent.Agent.run_on_startup(_startup_agent_console)
 
 
+def _setup_security_module():
+    """Initiates k2 security module and adds a
+    callback to agent startup to propagate NR config
+    """
+    try:
+        if not _settings.security.agent.enabled:
```
umaannamalai, line 3184: Is there error handling already in the security agent to handle `security.mode `or the `security.validator_service_url` settings not being configured by the user?

newrelic/config.py
```diff
@@ -3169,6 +3176,23 @@ def _setup_agent_console():
...
+            return
+        from newrelic_security.api.agent import Agent as SecurityAgent
+
+        # initialize security agent
+        security_agent = SecurityAgent()
+        # create a callback to reinitialise the security module
+        newrelic.core.agent.Agent.run_on_startup(security_agent.refresh_agent)
+    except Exception as k2error:
+        _logger.error("K2 Startup failed with error %s", k2error)
```
umaannamalai, line 3193: Can we change the language here to use `Security Agent` instead of `K2` to be consistent with other agents?

newrelic/config.py
umaannamalai: The above code suggestion will also need the application instance passed to the call to `run_on_registration()`

newrelic/core/config.py
```diff
@@ -840,6 +869,16 @@ def default_host(license_key):
     "NEW_RELIC_APPLICATION_LOGGING_LOCAL_DECORATING_ENABLED", default=False
 )
 
+_settings.security.agent.enabled = _environ_as_bool("NEW_RELIC_SECURITY_AGENT_ENABLED", False)
+_settings.security.enabled = _environ_as_bool("NEW_RELIC_SECURITY_ENABLED", False)
+_settings.security.mode = os.environ.get("NEW_RELIC_SECURITY_MODE", "IAST")
+_settings.security.validator_service_url = os.environ.get("NEW_RELIC_SECURITY_VALIDATOR_SERVICE_URL", None)
```
umaannamalai, line 875: We may want to provide `wss://csec.nr-data.net` as the default here.

AnupamJuniwal:
Stale. No longer required.

PR #821, opened 2023-06-02
Add OTLP protocol class & protos

hmstepanek:
# Overview
Add an OTLP protocol class and protos to use to upload data to OTLP collector.


newrelic/common/otlp_utils.py
```diff
@@ -0,0 +1,53 @@
...
+            return KeyValue(key=key, value=AnyValue(int_value=value))
+        elif isinstance(value, float):
+            return KeyValue(key=key, value=AnyValue(double_value=value))
+        elif isinstance(value, str):
+            return KeyValue(key=key, value=AnyValue(string_value=value))
+        # Technically AnyValue accepts array, kvlist, and bytes however, since
+        # those are not valid custom attribute types according to our api spec,
+        # we will not bother to support them here either.
+        else:
+            _logger.warn("Unsupported ML event attribute value type %s: %s." % (key, value))
```
TimPansino, line 40: Should this also be run through the attribute system? Or is this step after that's already been run?

newrelic/core/agent_protocol.py
```diff
@@ -516,3 +516,94 @@ def connect(
...
+        if settings.audit_log_file:
+            audit_log_fp = open(settings.audit_log_file, "a")
+        else:
+            audit_log_fp = None
+
+        otlp_host = self.HOST_MAP.get(host or settings.host, None)
+        if not otlp_host:
+            default = self.HOST_MAP["collector.newrelic.com"]
+            _logger.warn("Unable to find corresponding OTLP host using default %s" % default)
+            otlp_host = default
```
TimPansino, line 542: I feel like this does warrant a setting so customers could use a sidecar collector if they want.

newrelic/core/agent_protocol.py
```diff
@@ -516,3 +516,94 @@ def connect(
...
+
+        self.client = client_cls(
+            host=otlp_host,
+            port=4318,
```
TimPansino, line 546: If we do add a setting this port should probably not be hard coded anymore.

newrelic/core/agent_protocol.py
```diff
@@ -516,3 +516,94 @@ def connect(
...
+            proxy_host=settings.proxy_host,
+            proxy_port=settings.proxy_port,
+            proxy_user=settings.proxy_user,
+            proxy_pass=settings.proxy_pass,
+            timeout=settings.agent_limits.data_collector_timeout,
+            ca_bundle_path=settings.ca_bundle_path,
+            disable_certificate_validation=settings.debug.disable_certificate_validation,
+            compression_threshold=settings.agent_limits.data_compression_threshold,
+            compression_level=settings.agent_limits.data_compression_level,
+            compression_method=settings.compressed_content_encoding,
```
TimPansino, line 557: Do we use these still?

newrelic/core/agent_protocol.py
```diff
@@ -516,3 +516,94 @@ def connect(
...
+            max_payload_size_in_bytes=1000000,
+            audit_log_fp=audit_log_fp,
+        )
+
+        self._params = {
+            "protocol_version": self.VERSION,
+            "license_key": settings.license_key,
+            "marshal_format": "json",
+        }
```
TimPansino, line 566: Are these still valid / required?

newrelic/common/otlp_utils.py
TimPansino, line 40: This line specifically calls out events, but should be generic for OTLP.

newrelic/core/agent_protocol.py
hmstepanek: Yes, the default client class used by the protocol class is still ApplicationModeClient. ApplicationModeClient inherits from HttpClient. The send_request method of ApplicationModeClient is called inside the protocol class's send method. Inside the ApplicationModeClient's send_request method there is possible compression if the payload is over the compression threshold: https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/common/agent_http.py#L416.

newrelic/core/agent_protocol.py
hmstepanek: No, I don't think we actually need to pass any params in this case. I've removed these and added the license key as a header.

TimPansino (approved):

PR #822, opened 2023-06-05
Hook up ml event to OTLP

hmstepanek:
# Overview
Use otlp protocol class and protos or json fallback to create ml_event payload and create events in otlp.


TimPansino (approved):

TimPansino:
🚀

PR #826, opened 2023-06-09
OTLP Serialization for Dimensional Metrics

TimPansino:
# Overview
* Add dimensional metric serialization to OTLP via protobuf or json.


tests/agent_features/test_dimensional_metrics.py
```diff
@@ -70,37 +83,88 @@ def _test():
...
 
 @pytest.mark.parametrize("tags,expected", _test_tags_examples)
 @reset_core_stats_engine()
 def test_record_dimensional_metrics_outside_transaction(tags, expected):
-    @validate_dimensional_metrics_outside_transaction([("Metric/1", expected, 1), ("Metric/2", expected, 1)])
+    @validate_dimensional_metrics_outside_transaction([("Metric.1", expected, 1), ("Metric.2", expected, 1)])
     def _test():
         app = application_instance()
-        record_dimensional_metrics([("Metric/1", 1, tags), ("Metric/2", 1, tags)], application=app)
+        record_dimensional_metrics([("Metric.1", 1, tags), ("Metric.2", 1, tags)], application=app)
```
hmstepanek, line 106: Was the `/` to `.` conversion intentional here?

tests/agent_unittests/test_utilization_settings.py
```diff
@@ -118,22 +118,6 @@ def reset(wrapped, instance, args, kwargs):
     return reset
 
 
-@reset_agent_config(INI_FILE_WITHOUT_UTIL_CONF, ENV_WITHOUT_UTIL_CONF)
```
hmstepanek, line 121: I think this test was intended to be the host: None test. But I think we should keep these tests as they do test different things.

tox.ini
```diff
@@ -203,6 +204,8 @@ deps =
     adapter_waitress-waitress02: waitress<2.1
     adapter_waitress-waitresslatest: waitress
     agent_features: beautifulsoup4
+    agent_features-{py37,py38,py39,py310,py311,pypy37}: protobuf
```
hmstepanek, line 207: We should probably have some tests that run without protobuf as well to make sure that code path works as well.

hmstepanek (changes requested):
A couple suggestions mainly around testing. Overall looks good.

tests/agent_features/test_dimensional_metrics.py
TimPansino: Yeah I was having trouble querying these metrics in the UI with a slash. Not sure why but NRQL doesn't seem to accept it.

tox.ini
TimPansino: I was trying to think of how to make that work. Possibly have a debug setting that forces JSON serialization?

hmstepanek: Yeah you might be able to do that. It's a little tricky cause it's at import time inside of the nr core but yes perhaps just a switch would be good enough. The disadvantage of doing that is you are mocking a little bit so you would miss validating the exception logic around protobuf imports failing. I was thinking maybe just run a couple test envs with protobuf and the rest without. Something like:
```
python-agent_features-{py27,py311}-with_protobuf-without_extensions,
.
.
.
agent_features-{py311}-with_protobuf: protobuf
```

tests/agent_unittests/test_utilization_settings.py
hmstepanek: That's fine. We can move them into the other file then. I think I just saw a place where it was validating env vars and stuck it in there so my bad.

tests/agent_features/test_configuration.py
```diff
@@ -581,6 +583,7 @@ def test_translate_deprecated_ignored_params_with_new_setting():
         ("otlp_port", 0),
     ),
 )
+@override_generic_settings(global_settings(), {"host": None})
```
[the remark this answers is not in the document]
hmstepanek: This hasn't been addressed either. Since this isn't really related to the current PR and metrics at all, I think we should just drop this change from this PR. There's plenty of other changes going on here without adding this in.

hmstepanek (changes requested):
Had one small request about asserting the `service.provider` value inside the payload, otherwise I think the payload tests look good now. I also added a request to just drop the changes to the config tests for now since none of those issues were addressed since the last review and they seem pretty unrelated to this PR anyway-I rather not worry about them in this PR.

hmstepanek (approved):

PR #835, opened 2023-06-20
Remove Python 2.7 and pypy2 testing

lrafeei:
Python 2.7 [support was dropped](https://github.com/actions/setup-python/issues/672) on June 19th.  This solution is a TEMPORARY one that disables python 2.7 and pypy2 testing until a more permanent fix is implemented.


hmstepanek (approved):
Looks good to me.

PR #836, opened 2023-06-21
Containerized CI Pipeline

TimPansino:
# Overview

* Build and publish docker image for CI pipeline to continue supporting Python 2.7.

## Testing

`.github/containers` contains a Makefile that can be used to validate the install logic for Python 2 & 3 in the container. To build and test for python in the image, run:

```bash
make test
```

To run a local copy of the CI image with the repository mounted inside, run:

```bash
make run
```


.github/workflows/build-ci-image.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Generate Docker Metadata (Tags and Labels)
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: ghcr.io/${{ github.repository }}
+          flavor: |
+            prefix=
+            suffix=
+            latest=false
```
hmstepanek, line 46: Is this latest=false correct here? Seems like it should be true.

.github/workflows/build-ci-image.yml
```diff
@@ -0,0 +1,68 @@
...
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and Publish Image
+        uses: docker/build-push-action@v3
+        with:
+          push: ${{ github.event_name != 'pull_request' }}
```
hmstepanek, line 64: I thought we decided that we _were_ going to pushing it on each pull request? Did I miss-understand?

tox.ini
```diff
@@ -54,108 +54,123 @@ envlist =
...
     python-framework_starlette-{py37,py38,py39,py310,py311,pypy37}-starlettelatest,
     python-framework_strawberry-{py37,py38,py39,py310,py311}-strawberrylatest,
-    python-logger_logging-{py37,py38,py39,py310,py311,pypy37},
+    python-logger_logging-{py27,py37,py38,py39,py310,py311,pypy,pypy37},
     python-logger_loguru-{py37,py38,py39,py310,py311,pypy37}-logurulatest,
     python-logger_loguru-py39-loguru{06,05,04,03},
-    libcurl-framework_tornado-{py37,py38,py39,py310,py311,pypy37}-tornado0600,
-    libcurl-framework_tornado-{py38,py39,py310,py311}-tornadomaster,
-    rabbitmq-messagebroker_pika-{py37,py38,py39,pypy37}-pika0.13,
+    python-framework_tornado-{py37,py38,py39,py310,py311,pypy37}-tornado0600,
```
hmstepanek, line 156: I was a little confused about this change at first but I guess we're just standardizing this to be python-framework_ like the others.

hmstepanek (commented):
Overall looks really good. I had a couple questions. I'm not sure if they are actual issues or if I just didn't fully understand the plan but take a look and let me know.

.github/workflows/build-ci-image.yml
TimPansino: Yeah that disables the auto latest tagging which has some logic we're avoiding

tox.ini
TimPansino: Yeah that's the idea. No need to install packages on the fly when they can be baked in.

.github/workflows/build-ci-image.yml
hmstepanek: Ah ok! Thanks for the explanation!

hmstepanek (approved):

PR #837, opened 2023-06-21
Fix local scoped package reporting

hmstepanek:
# Overview
* Fixes a bug where locally scoped packages where not being reported on connect.
* Fixes a bug where importlib_metadata version was not being captured correctly.

# Testing
* Checkout main.
* `pip install -e <newrelic>`
* `pip install --user <modules>`
* Run a simple web app and via the plugins in the UI under the environment tab. The packages installed under user scope should be missing.
* Checkout this branch and re-run the test, now the packages installed under user scope should be listed in the UI.


newrelic/packages/isort/stdlibs/py36.py
```diff
@@ -0,0 +1,224 @@
+"""
```
umaannamalai, line 1: Do we need the py36 stdlib if the agent only supports 3.7+?

hmstepanek: No we don't but removing it requires modifying the __init__.py file from the latest on isort so I thought it was simpler to just leave it in for copy-paste purposes. I don't have any strong opinions on it though.

umaannamalai (approved):

PR #840, opened 2023-06-22
Remove pymssql datastore

lrafeei:
This PR removes the pymssql datastore from the list of instrumented frameworks.  This is due to the fact that this instrumentation was not actually done (the code for pymssql is written using *.pyc code).


TimPansino:
Are we sure this instrumentation doesn't work at all?

TimPansino:
Well that was a pain to install, but instrumentation for both `pymssql==2.2.7` and `pymssql==2.1.5` work perfectly for me in a linux container. No reason to drop this instrumentation.

Considering this is inside our support window, we should add tests for MSSQL.

TimPansino:
Closing in favor of adding testing

PR #856, opened 2023-06-27
Fix OTLP Count Metric Serialization

TimPansino:
# Overview

* Fix duplicate encoding of count metrics as summary metrics in OTLP.
  * Includes regression test.


hmstepanek (approved):
Super tricky bug! I just asked for a comment to explain why we are using type instead of isinstance for future devs who see this code and also to ignore the pylint rule on these lines. I'm just gonna pre-approve this.

PR #857, opened 2023-06-27
Add Transaction API Tests

lrafeei:
This PR adds more tests to cover some testing coverage gaps in newrelic/api/transactions.py


newrelic/api/transaction.py
```diff
@@ -1192,11 +1184,10 @@ def _accept_distributed_trace_payload(self, payload, transport_type="HTTP"):
...
                 return False
 
-            if "pr" in data:
-                try:
-                    data["pr"] = float(data["pr"])
-                except:
-                    data["pr"] = None
+            try:
+                data["pr"] = float(data["pr"])
+            except:
```
TimPansino, line 1189:
```suggestion
            except Exception:
```

If we're touching this line can we change this to avoid the blanket except statement? Causes problems for things like PDB.

TimPansino (approved):

PR #859, opened 2023-06-28
Ignore Django instrumentation from older versions

lrafeei:
This PR allows for coverage reporting to skip instrumentation that is used only in versions of Django that are no longer supported.  A subsequent PR will update instrumentation for newer versions of Django.


newrelic/hooks/framework_django.py
```diff
@@ -256,9 +270,14 @@ def wrapper(wrapped, instance, args, kwargs):
         yield wrapper(wrapped)
 
 
-def wrap_view_middleware(middleware):
+# Because this is not being used in any version of Django that is
+# within New Relic's support window, so no tests will be added
```
hmstepanek, line 274:
```suggestion
# within New Relic's support window, no tests will be added
```

newrelic/hooks/framework_django.py
```diff
@@ -600,10 +606,9 @@ def _wrapped(path):
             with FunctionTrace(name=name, label=path, source=wrapped):
                 result = wrapped(path)
 
-                if type(result) is tuple:
+                if isinstance(result, tuple):
```
hmstepanek, line 609: Sometimes there are good reasons for using the type() is instead of isinstance check, I suspect this is likely one of those cases so we should probably change this back to type(result) is tuple. Sometimes an isinstance check will return true while the type is check will return false (in inheritance cases). I wouldn't risk breaking the logic here which is why I say, while I don't know for sure without a deeper investigation, it's safer to leave it as it was.
```suggestion
                if type(result) is tuple:
```

hmstepanek (approved):
A couple small requests but otherwise looks good. I'm just gonna pre-approve this assuming you'll make the changes. Let me know if you have any questions about that isinstance check-it's kinda tricky sometimes!

hmstepanek (approved):

PR #860, opened 2023-06-29
Increase test coverage for newrelic/config.py

hmstepanek:
# Overview
Improve test coverage for newrelic/config.py


lrafeei (approved):

tests/agent_features/test_configuration.py
```diff
@@ -583,3 +597,380 @@ def test_default_values(name, expected_value):
...
+        initialize(config_file=f.name)
+
+    if expect_error:
+        assert "CONFIGURATION ERROR" in logger.caplog.records
+    else:
+        assert "CONFIGURATION ERROR" not in logger.caplog.records
+
+
+@pytest.fixture
+def caplog_handler():
```
umaannamalai, line 951: I wonder if we should make this it's own fixture in testing support since we use it in the logging tests as well.

PR #863, opened 2023-06-30
Fix Kafka CI

TimPansino:
# Overview

* Add `librdkafka` package from source to container for `confluent-kafka`.
* Reenable kafka testing on CI.
* Remove special case packages from kafka CI process.
* Adjust kafka db settings to force IPv4
* Small cleanup to pip scripting to match devcontainer.


tox.ini
```diff
@@ -160,7 +160,7 @@ envlist =
     python-framework_tornado-{py38,py39,py310,py311}-tornadomaster,
     rabbitmq-messagebroker_pika-{py27,py37,py38,py39,pypy27,pypy38}-pika0.13,
     rabbitmq-messagebroker_pika-{py37,py38,py39,py310,py311,pypy38}-pikalatest,
-    kafka-messagebroker_confluentkafka-{py27,py37,py38,py39,py310,py311}-confluentkafkalatest,
+    kafka-messagebroker_confluentkafka-{py37,py38,py39,py310,py311}-confluentkafkalatest,
```
hmstepanek, line 163: Why are we dropping 2.7 testing here? It looks like it's still supported on the latest version.

hmstepanek (changes requested):
Looks good-had one small question about 2.7 tests and a reminder about changing the tag to latest.

.github/workflows/tests.yml
```diff
@@ -119,7 +119,7 @@ jobs:
 
     runs-on: ubuntu-20.04
     container:
-      image: ghcr.io/newrelic/newrelic-python-agent-ci:sha-b0ffe8bdbb28ba0579377076ad680054da8fbc28
+      image: ghcr.io/newrelic/newrelic-python-agent-ci:sha-7f686305c0ddf52e731eb8ee91668ff73aeb7a03
```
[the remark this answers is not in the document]
TimPansino: Correct

tox.ini
TimPansino: From my testing the library is broken on Python 2 now, it throws import errors on the latest versions.

hmstepanek: Oh ouch ok. Sounds good then!

hmstepanek (approved):
Assuming the tests all pass I'm pre-approving this. Just make sure we change the docker image sha over to latest before merging.

PR #867, opened 2023-07-11
Only send 1 event per prediction & fix tests

hmstepanek:
# Overview
1. This changes from sending 1 event per feature/label value to 1 event per prediction (Also referred to as a change in "schema").
    * Replace feature_name & feature_value and label_name & label_value to `feature.<name> = <value>` and `label.<name> = <value>`.
    * Remove feature_type and label_type.
    * Add a new schema attribute.
1. Remove pypy38 from test suite and remove compile dependencies from Dockerfile. Scipy wasn't able to compile consistently regardless of installing the necessary dependencies and adding these extra dependencies broke the kafka compile.

### DIFF
--- a/newrelic/hooks/mlmodel_sklearn.py
+++ b/newrelic/hooks/mlmodel_sklearn.py
@@ -96,8 +96,7 @@ def _wrap_method_trace(module, class_, method, name=None, group=None):
         if method in ("predict", "fit_predict"):
             training_step = getattr(instance, "_nr_wrapped_training_step", "Unknown")
             inference_id = uuid.uuid4()
-            create_feature_event(transaction, class_, inference_id, instance, args, kwargs)
-            create_label_event(transaction, class_, inference_id, instance, return_val)
+            create_prediction_event(transaction, class_, inference_id, instance, args, kwargs, return_val)
             return PredictReturnTypeProxy(return_val, model_name=class_, training_step=training_step)
         return return_val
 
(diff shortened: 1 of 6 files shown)


TimPansino:
To be clear, we're removing PyPy3.8 testing because we can't get scipy to compile, yes? And these dependencies that are being dropped were added for that?

TimPansino:
Ah I didn't read the second half of the PR description, I see you already covered that

newrelic/hooks/mlmodel_sklearn.py
```diff
@@ -283,15 +235,38 @@ def bind_predict(X, *args, **kwargs):
...
     model_name = getattr(instance, "_nr_wrapped_name", class_)
     model_version = getattr(instance, "_nr_wrapped_version", "0.0.0")
     user_provided_feature_names = getattr(instance, "_nr_wrapped_feature_names", None)
+    label_names = getattr(instance, "_nr_wrapped_label_names", None)
     settings = transaction.settings if transaction.settings is not None else global_settings()
 
+    labels = []
+    if return_val is not None:
+        if not hasattr(return_val, "__iter__"):
+            labels = np.array([return_val])
```
TimPansino, line 251: Have we tested that this works for string return vals? Seems like it might split something like `"abc"` to `["a", "b", "c"]`.

hmstepanek:
It has been validated in the UI. I'm checking on that string related question now...

newrelic/hooks/mlmodel_sklearn.py
hmstepanek: Very good point! For the label columns, the values are only numeric (these values come from the output of the model prediction) so we should not need to handle this scenario.

TimPansino (approved):

PR #874, opened 2023-07-18
Merge main

hmstepanek:
# Overview
Merge main into dev branch.


TimPansino (approved):
Merges look fine, sounds like kafka is borked but that's not really a concern here

Issue #862, opened 2023-06-30
Access __version__ is deprecated


AngellusMortis:
**Description**

```console
/usr/local/lib/python3.10/site-packages/newrelic/common/package_version_utils.py:75: DeprecationWarning: Accessing attr.__version__ is deprecated and will be removed in a future release. Use importlib.metadata directly to query for attrs's 
packaging metadata.
```

**Expected Behavior**

Fix deprecation warning.

**Steps to Reproduce**

1. Import `newrelic` (may want to enable [devmode](https://docs.python.org/3/library/devmode.html))

**Your Environment**

* Python 3.10.11
* newrelic==8.8.1

TimPansino:
We could probably suppress deprecation warnings caused by this function. I'll add that as a feature request.

AngellusMortis:
Suppressing a deprecation is not really a "fix". It will eventually be removed. `importlib-metadata` is part of 3.10 stdlib and it is [backported on PyPi to Python 3.7+](https://pypi.org/project/importlib-metadata/). Migrating to using importlib-metadata as the deprecated suggest seems like the best solution so this does not break in a future version of Python.

lrafeei:
Just to clarify, this is something that will have to be resolved during our "Add support for Python 3.12 before it is released" task in the near future.

PR #878, opened 2023-07-22
Use importlib.metadata first to avoid deprecation warnings

renanivo:
# Overview
Tries first to get the package version using `importlib.metadata`. This should be enough to avoid the `DepracationWarning` discussed in the related issue.

# Related Github Issue
#862


renanivo:
I've included another commit to retrieve the package distribution name when a module is provided. Without that, the reported warning can still happen, when the module (`attr`) is provided instead of the package name (`attrs`).

lrafeei:
Thanks for the contribution!  I was going to include something like this in our Python 3.12 support work, but there is no reason we can't commit this sooner.  I just reshuffled some tests and added a conditional that will allow for the agent to continue supporting versions of Python below 3.10.

lrafeei (approved):

renanivo:
Thank you, @lrafeei!

PR #881, opened 2023-07-31
Redis asyncio testing

lrafeei:
This PR adds functionality to use asyncio as part of Redis

### DIFF
--- a/tests/datastore_redis/test_asyncio.py
+++ b/tests/datastore_redis/test_asyncio.py
@@ -92,7 +92,7 @@ def test_async_pubsub(client, loop):  # noqa
 
     async def _test_pubsub():
         async with client.pubsub() as pubsub:
-            await pubsub.subscribe("channel:1", "channel:2")
+            await pubsub.psubscribe("channel:*")
 
             future = asyncio.create_task(reader(pubsub))
 
--- a/tests/datastore_redis/test_asyncio.py
+++ b/tests/datastore_redis/test_asyncio.py
@@ -1,0 +1,105 @@
+# Copyright 2010 New Relic, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+
+import pytest
+from testing_support.db_settings import redis_settings
+from testing_support.fixture.event_loop import event_loop as loop  # noqa: F401
+from testing_support.util import instance_hostname
+from testing_support.validators.validate_transaction_metrics import (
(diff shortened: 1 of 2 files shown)


tests/datastore_redis/test_asyncio.py
```diff
@@ -0,0 +1,105 @@
...
+from newrelic.common.package_version_utils import get_package_version_tuple
+
+# Settings
+
+DB_SETTINGS = redis_settings()[0]
+REDIS_VERSION = get_package_version_tuple("redis")
+
+# Metrics
+
+_base_scoped_metrics = (("Datastore/operation/Redis/publish", 3),)
```
hmstepanek, line 35:
```suggestion
_scoped_metrics = [("Datastore/operation/Redis/publish", 3)]
```

tests/datastore_redis/test_asyncio.py
```diff
@@ -0,0 +1,105 @@
...
+from testing_support.fixture.event_loop import event_loop as loop  # noqa: F401
+from testing_support.util import instance_hostname
+from testing_support.validators.validate_transaction_metrics import (
+    validate_transaction_metrics,
+)
+
+from newrelic.api.background_task import background_task
+from newrelic.common.package_version_utils import get_package_version_tuple
+
+# Settings
```
hmstepanek, line 28: I'd suggest simplifying lines 28-57 to the following:
```
DB_SETTINGS = redis_settings()[0]
REDIS_VERSION = get_package_version_tuple("redis")

_enable_scoped_metrics = [("Datastore/operation/Redis/publish", 3)]

_enable_rollup_metrics = [
    ("Datastore/all", 3),
    ("Datastore/allOther", 3),
    ("Datastore/Redis/all", 3),
    ("Datastore/Redis/allOther", 3),
    ("Datastore/operation/Redis/publish", 3),
    ("Datastore/instance/Redis/%s/%s" % (instance_hostname(DB_SETTINGS["host"]), DB_SETTINGS["port"]), 3)
]

```

tests/datastore_redis/test_asyncio.py
```diff
@@ -0,0 +1,105 @@
...
+
+@pytest.mark.skipif(REDIS_VERSION < (4, 2), reason="This functionality exists in Redis 4.2+")
+@validate_transaction_metrics(
+    "test_asyncio:test_async_pubsub",
+    scoped_metrics=_enable_scoped_metrics,
+    rollup_metrics=_enable_rollup_metrics,
+    background_task=True,
+)
+@background_task()
+def test_async_pubsub(client, loop):  # noqa
```
hmstepanek, line 85: I have concerns about the way this test is written. It will loop forever until message["data"] == "NOPE" and there is an assumption here that it will grab all of the data but we aren't actually asserting that in the test. For example, it could just receive the final "NOPE" value first and we would be none the wiser. Instead I'd suggest asserting we received all of the data at the end of the test so something like:
```
        received_messages = []
        message = {"data": "Hello"}
        while message and message["data"] != b"NOPE":
            message = await client.get_message(ignore_subscribe_messages=True)
            if message:
                received_messages.append(message)
        assert received_messages == [{"data": "Hello"}, {"data": "World"}, {"data": "NOPE"}]
```

hmstepanek (changes requested):
Had a couple minor suggestions but otherwise looks good.

hmstepanek (approved):
Looks good!

PR #882, opened 2023-07-31
Firstore Async Instrumentation

TimPansino:
# Overview

* Adds instrumentation for Firestore Async APIs.

Depends on: #886


tests/datastore_firestore/test_async_transaction.py
```diff
@@ -0,0 +1,128 @@
...
+        with pytest.raises(RuntimeError):
+            await _exercise(async_client.transaction())
+        assert len([_ async for _ in async_collection.list_documents()]) == 3
+    return _exercise_async_transaction_rollback
+
+
+def test_firestore_async_transaction_commit(loop, exercise_async_transaction_commit, async_collection):
+    _test_scoped_metrics = [
+        ("Datastore/operation/Firestore/commit", 1),
+        # ("Datastore/operation/Firestore/get_all", 2),
```
umaannamalai, line 81: Commented out code left in here

TimPansino: Ah yep, leaving that there because it's what the tests SHOULD be doing but AsyncTransaction is broken. PR is open to Firestore to fix their library, we should see when this starts to work and can add this metric back in.

umaannamalai (approved):

PR #887, opened 2023-08-02
Firestore Instance Info

TimPansino:
# Overview

* Add instance info logic to firestore tracing.
* Add tests for all firestore API classes and methods to validate instance info is set.

Depends on: #882

### DIFF
--- a/newrelic/hooks/datastore_firestore.py
+++ b/newrelic/hooks/datastore_firestore.py
@@ -103,6 +103,9 @@ def instrument_google_cloud_firestore_v1_async_client(module):
                     operation=method,
                     product="Firestore",
                     target=None,
+                    host=_conn_str_to_host(_get_target),
+                    port_path_or_id=_conn_str_to_port(_get_target),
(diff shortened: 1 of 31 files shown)


newrelic/hooks/datastore_firestore.py
```diff
@@ -17,8 +17,49 @@
...
+
+    return closure
+
+
+def _conn_str_to_port(getter):
+    """Safely transform a getter that can retrieve a connection string into the resulting port."""
+
+    def closure(obj, *args, **kwargs):
+        try:
+            return int(getter(obj, *args, **kwargs).split(":")[1])
```
umaannamalai, line 37: Should this be casted to an int? Technically, in the datastore_trace, we defined the type as a string. That being said, it would then have to be casted in the tests to be an int since FIRESTORE_PORT is an integer.

TimPansino: I think we can make this a string

umaannamalai (approved):

PR #894, opened 2023-08-14
Fix Normalization Rules

TimPansino:
# Overview

* Normalization rules currently have no default values, and simply crash on init if an argument is missing. Fix this by adding sensible defaults to all values.
* Currently metrics are **NOT** ignored when using a MetricNormalizationRule with ignore set to True. Fix this with a harvest time check.
* Clean up fixtures in `cross_agent` test suite to be independent of current directory.


hmstepanek (approved):
LGTM. Gonna ignore kafka...

PR #899, opened 2023-08-16
Datastores: Replace __version__ with get_package_version

lrafeei:
This PR replaces __version__ with get_package_version in datastores.


TimPansino (approved):

TimPansino:
Copied from other PR: 

> Can we also add a `assert version is not None` to all the tests where we use get_package_version? That would be useful to avoid issues where we don't realize the version isn't propagating.

PR #900, opened 2023-08-16
GraphQL: Replace __version__ with get_package_version

lrafeei:
This PR replaces version with get_package_version in GraphQL frameworks


TimPansino:
I think this might be already covered in the develop branch for graphql that should merge soon. Can we confirm that first before accepting this PR?

PR #901, opened 2023-08-16
Message Brokers:  Replace __version__ with get_package_version

lrafeei:
This PR replaces version with get_package_version in message brokers


TimPansino (commented):
Can we also add a `assert version is not None` to all the tests where we use get_package_version? That would be useful to avoid issues where we don't realize the version isn't propagating.

PR #902, opened 2023-08-16
Lambdas and Boto: Replace __version__ with get_package_version

lrafeei:
This PR replaces version with get_package_version in lambdas/boto


hmstepanek (approved):
One thing to just be aware of is calling this function instead of using `module.__version__` directly takes ~3x as long. Mind you the difference is 10 microseconds compared to 30 microseconds. I think this is very negligible but just something to keep in mind with these types of changes.

PR #904, opened 2023-08-16
Update testing matrix for supported packages.

umaannamalai:
This PR removes unsupported package versions from the tox testing matrix.


TimPansino (approved):

lrafeei:
Should we add no-coverage flags in the parts of the code that support instrumentation for unsupported frameworks (in some cases, functions that have been removed/deprecated and no longer used in supported versions)?  This way, coverage numbers do not fall, and we do not actually remove existing code

lrafeei (approved):

TimPansino (approved):

PR #905, opened 2023-08-17
Fix database instance metric bug

hmstepanek:
# Overview
This is a fix for the database not showing up in the service map for certain database types.

`[enable_datastore_instance_feature` was added in 2016 when the database instance feature was first developed. It
appears to be a method of gating capturing datastore instance information internally within the agent at the time
that it was implemented. However, it is not needed now and database instrumentations
that don't call this are actually broken in that the metric that is used to create the service map (namely `Datastore/instance/MySQL/<host>/<port>`) does not get created due to not calling this enable feature function. This results in a UI issue where the database does not show up in the service map due to this missing metric.

Note several db's such as memcache and pyodb do not contain instance info.


lrafeei (approved):
🚀

PR #908, opened 2023-08-23
GraphQL Async Instrumentation Support

umaannamalai:
This PR drops support for` graphql-core` 2 and adds support for async features.


tests/datastore_solrpy/test_solr.py
```diff
@@ -13,6 +13,11 @@
 # limitations under the License.
 
 from solr import SolrConnection
+<<<<<<< HEAD
```
hmstepanek, line 16: There's an unhandled merge conflict here still.

hmstepanek (changes requested):
It looks like there's just that one unhandled merge conflict. Otherwise it looks good.

hmstepanek (approved):

PR #910, opened 2023-08-25
Make inferences a separate event

lrafeei:
This PR makes the following changes as requested from MLops team:

- Each inference is now a separate event.  Attach an inference_id to each event.
- Allow addition of metadata to events
- Do not automatically convert feature values to string.


newrelic/core/attribute.py
```diff
@@ -282,14 +279,40 @@ def process_user_attribute(name, value, max_length=MAX_ATTRIBUTE_LENGTH, ending=
...
+    #   isinstance([var with type <class 'numpy.str_'>], <class 'str'>) == True
+    #   type([var with type <class 'numpy.str_'>]) != <class 'str'>
+    #
+
+    # This is to capture any potential numpy types that could
+    # otherwise correspond to a standard python type and
+    # converts them to the appropriate standard python type.
+    # If this is not the case, the function will attempt to
+    # convert the value into a string type. (e.g. UUID types)
+    if not type(value) in valid_value_types:
```
hmstepanek, line 310: I'm concerned about the performance impact this might have on the rest of the agent code for customers that add a lot of custom attributes. I wonder if it might be better to do this just inside the mlops code when setting the attribute?

newrelic/hooks/mlmodel_sklearn.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, wrapped, model_name, training_step):
         self._nr_training_step = training_step
 
 
-def _wrap_method_trace(module, class_, method, name=None, group=None):
+def _wrap_method_trace(module, class_, method, name=None, group=None, metadata=None, label_mapping=None):
```
hmstepanek:
```suggestion
def _wrap_method_trace(module, class_, method, name=None, group=None):
```

tests/mlmodel_sklearn/test_ml_model.py
```diff
@@ -156,16 +156,62 @@ def _test():
...
+        x_train = [[0, 0], [1, 1]]
+        y_train = [0, 1]
+        x_test = [[1.0, 2.0]]
+
+        model = CustomTestModel().fit(x_train, y_train)
+        wrap_mlmodel(
+            model,
+            name="MyCustomModel",
+            version="1.2.3",
+            metadata={"metadata1": "WHAT?", "metadata2": "YEAH!"},
```
hmstepanek, line 193: I know I'm a killing the amusement here but:
```
metadata={"metadata1": "value1", "metadata2": "value2"},
```

hmstepanek (changes requested):
I've suggested a lot of changes here which might be kinda hard to follow this way so we might wanna just pair on this tomorrow if you would find that easier. Overall pretty good. The major issue is the passing of the metadata and mapping here needs to be done via setting an attribute in the wrapper and then referencing that in the instrumentation instead of trying to pass it through the call stack. It looks like you originally start on the right path in your commit history but then deviated from it in a follow up commit but I think it's a relatively quick fix to adjust it. I've left a note in our channel but I'm considering dropping the mapping addition based on what we discussed in our planning meeting earlier today.

hmstepanek (approved):

PR #921, opened 2023-09-14
Add sklearn instrumentation and ML model feature support

umaannamalai:
This PR adds new instrumentation to support monitoring of ML models created with `sklearn`.  It also introduces two new, publics API. The first API is named `record_ml_event` and allows custom ML events with information about the prediction to be recorded. The second API is named `wrap_mlmodel` and allows users to pass additional information (including name and version) about their model directly as a decorator.


TimPansino (approved):
🚀

PR #923, opened 2023-09-18
Disable Sending of ML Events by Default

umaannamalai:
This PR sets the default value of `ml_insights_events.enabled` to `False` so it can be intentionally enabled by users who would like to send ML events to NR.


TimPansino (approved):

tests/agent_features/test_ml_events.py
```diff
@@ -121,27 +125,31 @@ def _test():
...
 
+@override_application_settings({"ml_insights_events.enabled": True})
 @reset_core_stats_engine()
 @validate_ml_event_count(count=0)
 @background_task()
 def test_record_ml_event_inside_transaction_params_not_a_dict():
     record_ml_event("ParamsListEvent", ["not", "a", "dict"])
 
 
+@override_application_settings({"ml_insights_events.enabled": True})
```
TimPansino, line 152: I don't think all of the tests in this folder shouldn't need a decorator for each function. We can just add these settings to the defaults and default them all to on.

newrelic/core/config.py
```diff
@@ -717,7 +717,7 @@ def default_otlp_host(host):
 _settings.transaction_events.attributes.include = []
 
 _settings.custom_insights_events.enabled = True
-_settings.ml_insights_events.enabled = True
+_settings.ml_insights_events.enabled = False
```
TimPansino, line 720: Is this supposed to stay false too? Or just the global override setting?

TimPansino (changes requested):

newrelic/core/config.py
umaannamalai: This one was supposed to stay False since we check this as a standalone setting in calls to `record_ml_event` and not in conjunction with machine_learning.enabled.

TimPansino (approved):

PR #934, opened 2023-10-10
Add OpenAI sync chat completion instrumentation

umaannamalai:
This PR adds instrumentation and testing for OpenAI's chat completions endpoint using synchronous `create` calls.


newrelic/hooks/mlmodel_openai.py
```diff
@@ -0,0 +1,145 @@
...
+    settings = transaction.settings if transaction.settings is not None else global_settings()
+
+    chat_completion_summary_dict = {
+        "id": chat_completion_id,
+        "appName": settings.app_name,
+        "conversation_id": conversation_id,
+        "span_id": span_id,
+        "trace_id": trace_id,
+        "transaction_id": transaction._transaction_id,
+        "request_id": request_id,
```
hmstepanek, line 60: This wasn't in the open-ai CDD though we need it for feedback. Scope wise, it makes sense to attach it here. But it's only used for identifying a message in the feedback CDD so I wonder if we need it here or just on the LlmMessage event? I suppose it doesn't hurt to have both, it's a pretty small amount of data.

Alternatively, we could make the chat completion id the request id just like we made the message id the response id + the index? Maybe we should ask the group about this?

newrelic/hooks/mlmodel_openai.py
```diff
@@ -0,0 +1,145 @@
...
+    message_list = list(kwargs.get("messages", [])) + [response.choices[0].message]
+
+    create_chat_completion_message_event(transaction, settings.app_name, message_list, chat_completion_id, span_id, trace_id, response_model, response_id, request_id)
+
+    return response
+
+
+def check_rate_limit_header(response_headers, header_name, is_int):
+    if not response_headers:
+        return ""
```
hmstepanek, line 95: If we return an empty string, this attribute is not stored on the event. This seems fine to me, just confirming we are in agreement that that's the behavior we want.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -0,0 +1,145 @@
...
+
+    if header_name in response_headers:
+        header_value = response_headers.get(header_name)
+        if is_int:
```
hmstepanek, line 99: Should probably have a try except around this int casting.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def create_chat_completion_message_event(transaction, app_name, message_list, chat_completion_id, span_id, trace_id, response_model, response_id, request_id):
+    if not transaction:
+        return
+
+    for index, message in enumerate(message_list):
+        chat_completion_message_dict = {
+            "id": "%s-%s" % (response_id, index),
+            "appName": app_name,
```
hmstepanek, line 113: I think we need to include conversation_id here since an event is uniquely defined by (conversation id, request id, message id).

newrelic/hooks/mlmodel_openai.py
```diff
@@ -0,0 +1,145 @@
...
+            "request_id": request_id,
+            "span_id": span_id,
+            "trace_id": trace_id,
+            "transaction_id": transaction._transaction_id,
+            "content": message.get("content", "")[:MAX_LOG_MESSAGE_LENGTH],
+            "role": message.get("role"),
```
hmstepanek, line 119: Should probably set a default for this.
```suggestion
            "role": message.get("role", ""),
```

tests/mlmodel_openai/test_chat_completion.py
```diff
@@ -13,19 +13,184 @@
...
+    add_custom_attribute("conversation_id", "my-awesome-id")
+    openai.ChatCompletion.create(
+        model="gpt-3.5-turbo",
+        messages=_test_openai_chat_completion_sync_messages,
+        temperature=0.7,
+        max_tokens=100
+    )
+
+
+disabled_ml_settings = {
```
hmstepanek, line 178: Is there something in the CDD about putting this behind an on-off switch like machine_learning.enabled? If there's not I'd say let's just leave that out of the implementation and tests for now so this would just be:
disabled_ml_event_settings = {  # Should not send events
    "ml_insights_events.enabled": False
}

If there is, then we should only record an event if `ml_insights_events.enabled=True`. And we should test these settings independently so have a test for:
```
disabled_ml_settings = {  # Should not send events
    "machine_learning.enabled": True,
    "ml_insights_events.enabled": False
}
disabled_ml_settings = {  # Should not send events
    "machine_learning.enabled": False,
    "ml_insights_events.enabled": True
}
disabled_ml_settings = {  # Should send events.
    "machine_learning.enabled": True,
    "ml_insights_events.enabled": True
}
```

tests/mlmodel_openai/test_chat_completion.py
```diff
@@ -13,19 +13,184 @@
...
+    set_trace_info()
+    add_custom_attribute("conversation_id", "my-awesome-id")
+    openai.ChatCompletion.create(
+        model="gpt-3.5-turbo",
+        messages=_test_openai_chat_completion_sync_messages,
+        temperature=0.7,
+        max_tokens=100
+    )
+
+
```
hmstepanek, line 164: We should have a test in here for when conversation_id isn't set on the transaction.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -0,0 +1,145 @@
...
+
+    ft_name = callable_name(wrapped)
+    with FunctionTrace(ft_name) as ft:
+        response = wrapped(*args, **kwargs)
+
+    if not response:
+        return
+
+    custom_attrs_dict = transaction._custom_params
+    conversation_id = custom_attrs_dict.get("conversation_id", str(uuid.uuid4()))
```
hmstepanek, line 40: We decided this in a recent meeting and I've updated the CDD's accordingly since then. If the customer doesn't set the conversation id on the transaction then it should be defaulted to NULL. I think I might clarify this to say it won't be present if the user doesn't set it so this code should be:
```suggestion
    conversation_id = custom_attrs_dict.get("conversation_id", "")
```

tests/mlmodel_openai/test_chat_completion.py
```diff
@@ -13,19 +13,184 @@
...
+disabled_ml_settings = {
+    "machine_learning.enabled": False,
+    "ml_insights_events.enabled": False
+}
+
+
```
hmstepanek, line 183: Given the timeline, I'm not going to block this PR over it but, we should probably add some tests in here for validating payloads and headers where attributes are unexpectedly not present in the response.

hmstepanek (changes requested):
Overall looks really good! Had some minor requests.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -0,0 +1,145 @@
...
+        "response.headers.llmVersion": response_headers.get("openai-version", ""),
+        "response.headers.ratelimitLimitRequests": check_rate_limit_header(response_headers, "x-ratelimit-limit-requests", True),
+        "response.headers.ratelimitLimitTokens": check_rate_limit_header(response_headers, "x-ratelimit-limit-tokens", True),
+        "response.headers.ratelimitResetTokens": check_rate_limit_header(response_headers, "x-ratelimit-reset-tokens", False),
+        "response.headers.ratelimitResetRequests": check_rate_limit_header(response_headers, "x-ratelimit-reset-requests", False),
+        "response.headers.ratelimitRemainingTokens": check_rate_limit_header(response_headers, "x-ratelimit-remaining-tokens", True),
+        "response.headers.ratelimitRemainingRequests": check_rate_limit_header(response_headers, "x-ratelimit-remaining-requests", True),
+        "vendor": "openAI",
+        "ingest_source": "Python",
+        "number_of_messages": len(kwargs.get("messages", [])) + len(response.choices),
```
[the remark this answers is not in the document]
umaannamalai: Yes, fixed!

newrelic/hooks/mlmodel_openai.py
umaannamalai: Checking with the AI team on this

umaannamalai: This behavior is confirmed to be fine 👍🏽

tests/mlmodel_openai/test_chat_completion.py
umaannamalai: We don't have any mention of gating this behind `machine_learning.enabled` (at least currently). I will remove that from the tests... we can add in the setting combo tests if we decide to bring this in later.

newrelic/hooks/mlmodel_openai.py
hmstepanek: @umaannamalai Was just doing a quick once-over. It looks like this issue is still outstanding ^ but otherwise everything has been addressed (minus the request_id question but I'm fine merging it as is and we can change it later if we need to).

hmstepanek (changes requested):
Just doing a re-review. It looks good to me now. I think let's address the .get(<attr>, "") later as the implementation is gonna be a little tricky to get right given all the chained lookups and indexing of the attr values. That way we can get this out while we hash out the rest of the subtler details later behind the scenes.

newrelic/hooks/mlmodel_openai.py
umaannamalai: Good point! I checked about this and I think we're going to go ahead and keep the request ID on the summary event and leave the chat completion ID as a UUID for our first iteration.

hmstepanek (approved):
Looks good!

PR #937, opened 2023-10-11
Bedrock Testing Infrastructure

TimPansino:
# Overview

* Add testing infrastructure for AWS Bedrock to tox.
* Add base files for test suite
  * Includes a simple no harm test for 4 text based models, and recorded responses.
* Add record/replay server for Bedrock
* Move OpenAI record/replay server to test suite folder and out of testing_support
  * The files were not general enough to exactly reuse, had to implement new logic for parsing prompts and headers from objects and payloads.
* Change recording for OpenAI/Bedrock to use json printer
  * pprint looks nice, but was getting in the way with splitting strings on long lines.
  * This now records `None` as `null` which is annoying but can be fixed with a find/replace.

### DIFF
--- a/.github/actions/update-rpm-config/action.yml
+++ b/.github/actions/update-rpm-config/action.yml
@@ -1,0 +1,109 @@
+name: "update-rpm-config"
+description: "Set current version of agent in rpm config using API."
+inputs:
+  agent-language:
+    description: "Language agent to configure (eg. python)"
+    required: true
+    default: "python"
+  target-system:
+    description: "Target System: prod|staging|all"
+    required: true
+    default: "all"
+  agent-version:
+    description: "3-4 digit agent version number (eg. 1.2.3) with optional leading v (ignored)"
+    required: true
+  dry-run:
+    description: "Dry Run"
+    required: true
+    default: "false"
+  production-api-key:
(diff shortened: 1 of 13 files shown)


hmstepanek (changes requested):
In general this looks good but I'm not seeing all of the models/responses in the mocked list. We need all of the following models/responses to be mocked:
* Claude
* Command
* Jurassic
* Llama 2
* Stable Diffusion
* Titan

Could you add in the mocked responses for the missing models?
I'm also ok with merging this and then doing it in a follow up PR if you would prefer.

TimPansino:
> In general this looks good but I'm not seeing all of the models/responses in the mocked list. We need all of the following models/responses to be mocked:
> 
> * Claude
> * Command
> * Jurassic
> * Llama 2
> * Stable Diffusion
> * Titan
> 
> Could you add in the mocked responses for the missing models? I'm also ok with merging this and then doing it in a follow up PR if you would prefer.

After some discussion, we'll stick to the text generation model set presented here for the initial PR. Follow up PRs to add support for various other models will require the inclusion of mocking for their various return types if not added here.

hmstepanek (approved):

PR #938, opened 2023-10-11
Add OpenAI sync embedding instrumentation

umaannamalai:
This PR adds instrumentation for OpenAI sync [embeddings](https://platform.openai.com/docs/guides/embeddings/what-are-embeddings).


newrelic/hooks/mlmodel_openai.py
```diff
@@ -12,15 +11,83 @@
...
 from newrelic.common.object_names import callable_name
-from newrelic.core.attribute import MAX_LOG_MESSAGE_LENGTH
+from newrelic.common.object_wrapper import wrap_function_wrapper
+from newrelic.core.config import global_settings
+
+
+def wrap_embedding_create(wrapped, instance, args, kwargs):
+    transaction = current_transaction()
+    if not transaction:
+        return
```
hmstepanek, line 30: I fixed this in the async PR but this should be:
```suggestion
        return wrapped(*args, **kwargs)
```

newrelic/hooks/mlmodel_openai.py
```diff
@@ -12,15 +11,83 @@
...
+
+    ft_name = callable_name(wrapped)
+    with FunctionTrace(ft_name) as ft:
+        response = wrapped(*args, **kwargs)
+
+    if not response:
+        return
```
hmstepanek, line 37: Again, fixed this in the async PR but...it's possible (though unlikely) that response isn't just None but {} or something else that is Falsey and not None. Just to be safe I'd suggest doing the following instead of an empty return:
```suggestion
        return response
```

hmstepanek (approved):
There's some default settings that I've kinda cleaned up in my PR but we can just take care of it in that one. This looks good as is IMO.

PR #940, opened 2023-10-17
Attach ml_event to APM entity by default

hmstepanek:
# Overview
By default, attach ml events to the APM entity. Only if the event name is InferenceEvent do not attach it to the APM entity.

### DIFF
--- a/tests/agent_features/test_ml_events.py
+++ b/tests/agent_features/test_ml_events.py
@@ -58,10 +58,41 @@ def core_app(collector_agent_registration):
 
 
 @validate_ml_event_payload(
-    [{"foo": "bar", "real_agent_id": "1234567", "event.domain": "newrelic.ml_events", "event.name": "InferenceEvent"}]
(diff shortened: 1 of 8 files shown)


tests/agent_features/test_ml_events.py
```diff
@@ -58,20 +58,91 @@ def core_app(collector_agent_registration):
...
+                "foo": "bar",
+                "real_agent_id": "1234567",
+                "event.domain": "newrelic.ml_events",
+                "event.name": "InferenceEvent",
+            }
+        ]
+    }
+)
+@reset_core_stats_engine()
+def test_ml_event_payload_noninference_event_inside_transaction(core_app):
```
umaannamalai, line 95: It looks like this test may need to be renamed since it is testing an inference event.

umaannamalai (commented):
🚀

umaannamalai (approved):

PR #941, opened 2023-10-18
OpenAI ErrorTrace attributes

lrafeei:
This PR adds certain attributes to OpenAI's ErrorTrace


newrelic/hooks/mlmodel_openai.py
```diff
@@ -24,14 +24,53 @@
...
+        "request.model": request_args.get("model") or request_args.get("engine", "None"),
+        "temperature": request_args.get("temperature", "None"),
+        "max_tokens": request_args.get("max_tokens", "None"),
+        "vendor": "openAI",
+        "ingest_source": "Python",
+        "organization": getattr(exception, "organization", "None"),
+        "number_of_messages": len(request_args.get("messages", [])),
+        "status_code": status_code,
+        "error_message": getattr(exception, "_message", "None"),
+        "error_type": exception.__class__.__name__ or "None",
```
hmstepanek, line 54: Our current notice_error instrumentation already pulls the class off of the exception internally so this attribute is a repeat. The difference is that our instrumentation includes the module path in the class name but I think that's fine here.
```suggestion
```

newrelic/hooks/mlmodel_openai.py
```diff
@@ -24,14 +24,53 @@
...
         return wrapped(*args, **kwargs)
 
     ft_name = callable_name(wrapped)
     with FunctionTrace(ft_name) as ft:
-        response = wrapped(*args, **kwargs)
+        try:
+            response = wrapped(*args, **kwargs)
+        except Exception as exc:
+            error_attributes = openai_error_attributes(exc, kwargs)
+            transaction.notice_error(attributes=error_attributes)
```
hmstepanek, line 72: We need notice_error to pick up some additional attributes that aren't picked up in transaction's notice_error. Specifically [these](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/api/time_trace.py#L356-L358):
```
        # Add error details as agent attributes to span event.
        self._add_agent_attribute("error.class", fullname)
        self._add_agent_attribute("error.message", message)
        self._add_agent_attribute("error.expected", is_expected)
```
Luckily function_trace inherits from time_trace and picks these up for us.
```suggestion
            ft.notice_error(attributes=error_attributes)
```

Along with this it looks like we need to make a change to the notice_error there to optionally take in a message argument to set the message.

hmstepanek (changes requested):
This is the right idea but there's a lot here that needs to be changed, some of it is subtle name changes and some of it is more substantial. Let's schedule a pairing session for tomorrow. Feel free to take a look before hand and then we can just pair on it as I think that would be easier.

tests/mlmodel_openai/test_error.py
```diff
@@ -0,0 +1,196 @@
...
+@background_task()
+def test_wrong_api_key_error(monkeypatch):
+    with pytest.raises(openai.error.AuthenticationError):
+        monkeypatch.setattr(openai, "api_key", "DEADBEEF")  # openai.api_key = "DEADBEEF"
+        openai.ChatCompletion.create(
+            model="gpt-3.5-turbo",
+            messages=({"role": "user", "content": "Invalid API key."},),
+            temperature=0.7,
+            max_tokens=100,
+        )
```
hmstepanek, line 196: While we instrumented both embedding and chatcompletion it looks like we only have error tests for chatcompletion. We need to make some for embedding as well.

hmstepanek (changes requested):

hmstepanek (commented):
In addition to wrapping sync for errors we also need to catch errors in async for these methods. We can do this is a separate PR though. I think there's prob enough here as it is is.

tests/mlmodel_openai/test_error.py
```diff
@@ -0,0 +1,196 @@
...
+            "vendor": "openAI",
+            "ingest_source": "Python",
+            "response.number_of_messages": 1,
+            "error.code": "model_not_found",
+        },
+    },
+)
+@validate_span_events(
+    exact_agents={
+        "error.message": "The model `does-not-exist` does not exist",
```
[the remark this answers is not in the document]
lrafeei: Unfortunately not.  The `error.message` is in one span, and the `http.statusCode` is in another span.  Putting it together will make the validator want to look for both attributes in one span.

tests/mlmodel_openai/test_error.py
```diff
@@ -0,0 +1,196 @@
...
+def test_authentication_error(monkeypatch):
+    with pytest.raises(openai.error.AuthenticationError):
+        monkeypatch.setattr(openai, "api_key", None)  # openai.api_key = None
+        openai.ChatCompletion.create(
+            model="gpt-3.5-turbo",
+            messages=_test_openai_chat_completion_sync_messages,
+            temperature=0.7,
+            max_tokens=100,
+        )
+
```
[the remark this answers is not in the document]
lrafeei: For ChatCompletion, `temperature` and `max_token` are not required arguments, so it will not generate an error.  (Seems like `model`/`engine` and `message` are the only required arguments)

tests/mlmodel_openai/test_error.py
lrafeei: Done!

newrelic/hooks/mlmodel_openai.py
```diff
@@ -27,6 +27,28 @@
...
+
+    error_attributes = {
+        "api_key_last_four_digits": api_key_last_four_digits,
+        "request.model": request_args.get("model") or request_args.get("engine") or "",
+        "request.temperature": request_args.get("temperature", ""),
+        "request.max_tokens": request_args.get("max_tokens", ""),
+        "vendor": "openAI",
+        "ingest_source": "Python",
+        "response.organization": getattr(exception, "organization", ""),
+        "response.number_of_messages": number_of_messages if number_of_messages > 0 else "",
```
hmstepanek, line 43: I'd just leave this simplified as the following. It does actually get reported if it's 0 but I think we want that for the UI in this case.
```
"response.number_of_messages": number_of_messages,
```

hmstepanek (changes requested):
The main thing here is the inconsistency in agent attributes but I think if you do the following that should resolve all those inconsistencies:
```
        try:
            response = wrapped(*args, **kwargs)
        except Exception as exc:
            error_attributes = openai_error_attributes(exc, kwargs)
            status_code=error_attributes.pop("status_code")
            # Usually the status_code is only added to an external error trace's attrs.
            ft._add_agent_attribute("http.statusCode", status_code)
            ft.notice_error(
                status_code=status_code,
                message=error_attributes.pop("error.message"),
                attributes=error_attributes,
            )
            raise
```

hmstepanek (approved):
Thank you for sticking with this change. It was not as easy as I anticipated! This looks really good now!

I pushed a couple commits:
1. To get rid of the config settings overrides inside the test files. We don't need to override the ignore errors based on status codes anymore and the rest of the settings are set in conftest already so the overrides are no longer needed.
2. To add a test for the _nr_message change in stats engine.

PR #942, opened 2023-10-18
Add openai feedback support

hmstepanek:
# Overview
Add get_ai_message_ids api function to retrieve message ids to be used in the customer's ai feedback endpoint.


umaannamalai:
This is looking good, will we have a separate PR for the new `record_ai_feedback_event` API?

tests/mlmodel_openai/test_get_ai_message_ids.py
```diff
@@ -0,0 +1,127 @@
...
+
+_test_openai_chat_completion_messages_1 = (
+    {"role": "system", "content": "You are a scientist."},
+    {"role": "user", "content": "What is 212 degrees Fahrenheit converted to Celsius?"},
+)
+_test_openai_chat_completion_messages_2 = (
+    {"role": "system", "content": "You are a mathematician."},
+    {"role": "user", "content": "What is 1 plus 2?"},
+)
+expected_message_ids_1 = [
```
umaannamalai, line 30: It may be worth testing some cases here where one or some combo of these values are `None`. At a minimum, it's likely conversation ID will be empty in many cases

hmstepanek: I've added some tests for conversation id being unset. I was originally gonna add more tests into this anticipating bedrock and the response_id being unset but I thought it was better to wait on those until we actually get to bedrock.

umaannamalai (approved):

PR #943, opened 2023-10-18
Add truncation for ML events.

umaannamalai:
This PR truncates attributes on ML events to a default of 4096 characters instead of the standard 255.


newrelic/core/attribute.py
```diff
@@ -89,6 +89,7 @@
 
 MAX_NUM_USER_ATTRIBUTES = 128
 MAX_ATTRIBUTE_LENGTH = 255
+MAX_ML_ATTRIBUTE_LENGTH = 4096
```
hmstepanek, line 92: When I ran this through to the UI it showed an off-by-1 error. We just need to change this to 4095.
```suggestion
MAX_ML_ATTRIBUTE_LENGTH = 4095
```

hmstepanek (changes requested):
Just a quick fix on this one and it should be good to go. We should really fix the limit on the number of attributes here as well for ml_event but I'm fine doing that in a follow up PR.

hmstepanek (approved):

hmstepanek (changes requested):
I had one suggestion for cleaning up the code a bit. It looks like we have the custom event user attribute limit in the code wrong but that's a separate issue.

hmstepanek (approved):

PR #945, opened 2023-10-18
Add framework metric for OpenAI.

umaannamalai:
This PR adds framework metrics for OpenAI to support entity tagging in the NR UI.


newrelic/hooks/mlmodel_openai.py
```diff
@@ -29,6 +30,8 @@ def wrap_embedding_create(wrapped, instance, args, kwargs):
     if not transaction:
         return wrapped(*args, **kwargs)
 
+    transaction.add_ml_model_info("OpenAI", get_package_version("openai"))
```
hmstepanek, line 33: We need to move the `get_package_version("openai")` to the top of the module so that it doesn't get computed every time for performance reasons.

hmstepanek (changes requested):
Looks good other than one small fixup for performance.

newrelic/hooks/mlmodel_openai.py
umaannamalai: Good catch!

hmstepanek (approved):

PR #948, opened 2023-10-19
Automatic RPM System Updates

TimPansino:
# Overview

* Automatically update RPM system configuration with new agent versions.

# Testing

* Tests replaced with dry-run of this configuration.
  * Successful run: https://github.com/newrelic/newrelic-python-agent/actions/runs/6616665000/job/17971392198


.github/actions/update-rpm-config/action.yml
```diff
@@ -0,0 +1,109 @@
...
+    description: "API key for New Relic Production"
+    required: false
+  staging-api-key:
+    description: "API key for New Relic Staging"
+    required: false
+
+runs:
+  using: "composite"
+  steps:
+    - name: Trim potential leading v from agent version
```
hmstepanek, line 29: Why not just pass the version without the preceding v?

hmstepanek (approved):
Just had one question otherwise looks good. I'm just gonna pre-approve it and let you decide whether it make sense to change it or not.

.github/actions/update-rpm-config/action.yml
TimPansino: It's easier to put the normalization logic inside the "function" (action) than to rely on the format of the input being exactly correct. Especially if you just pass in `${{github.ref_name}}` which comes with a `v` prefix for us, it prevents you from having to run a step of logic in the workflow to trim that before using it as input.

PR #950, opened 2023-10-23
Mock openai error responses

hmstepanek:
# Overview
Add ability to record and mock error responses from open ai server.

### DIFF
--- a/tests/mlmodel_openai/test_error.py
+++ b/tests/mlmodel_openai/test_error.py
@@ -44,7 +44,8 @@ def test_invalid_request_error_model_does_not_exist():
 
 
 @background_task()
-def test_authentication_error_invalid_api_key():
+def test_authentication_error_invalid_api_key(monkeypatch):
+    monkeypatch.setattr(openai, "api_key", "InvalidKey")
     with pytest.raises(openai.error.AuthenticationError):
         openai.ChatCompletion.create(
             model="gpt-3.5-turbo",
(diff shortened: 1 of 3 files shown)


tests/mlmodel_openai/test_error.py
```diff
@@ -0,0 +1,57 @@
...
+    with pytest.raises(openai.error.AuthenticationError):
+        openai.ChatCompletion.create(
+            model="gpt-3.5-turbo",
+            messages=(
+                {"role": "system", "content": "Invalid API key."},
+                {"role": "user", "content": "What is 212 degrees Fahrenheit converted to Celsius?"},
+            ),
+            temperature=0.7,
+            max_tokens=100,
+        )
```
TimPansino, line 57: Does this test not require you to set the API key to nothing when recording responses? Or does this input somehow  cause it to throw that error?

hmstepanek: Yes, it's just based on the input. I chose to not modify the existing logic in the mock server so it's still based on the content, even though yes, technically in the real world you would need to set an invalid license key (arguably `NOT-A-REAL-SECRET` is already not a real license key).

tests/mlmodel_openai/test_error.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@background_task()
+def test_authentication_error_invalid_api_key():
+    with pytest.raises(openai.error.AuthenticationError):
```
TimPansino, line 48:
```suggestion
def test_authentication_error_invalid_api_key(monkeypatch):
    monkeypatch.setattr(openai, "api_key", "InvalidKey")
    with pytest.raises(openai.error.AuthenticationError):
```

TimPansino: This change would probably be sufficient to cause the appropriate error even when using real credentials.

TimPansino (approved):

PR #951, opened 2023-10-25
Add streaming support for openai

hmstepanek:
# Overview
Add streaming support for openai instrumentation


umaannamalai (approved):
After cross-checking this PR with the OpenAI v0 streaming implementation, I think this looks good and covers the edge cases I was concerned about. There are just a few minor attribute naming things that don't need to be addressed in this specific PR so looks good 🚀

PR #957, opened 2023-11-03
AWS Bedrock Embedding Instrumentation

TimPansino:
# Overview

* Add instrumentation for AWS Bedrock's titan embedding models.


tests/mlmodel_bedrock/test_chat_completion.py
```diff
@@ -98,11 +89,18 @@ def expected_events_no_convo_id(model_id):
 
 
 @reset_core_stats_engine()
-def test_bedrock_chat_completion_in_txn(exercise_model, expected_events):
+def test_bedrock_chat_completion_in_txn_with_convo_id(set_trace_info, exercise_model, expected_events):
     @validate_ml_events(expected_events)
     # One summary event, one user message, and one response message from the assistant
     @validate_ml_event_count(count=3)
-    @background_task()
+    # @validate_transaction_metrics(
```
hmstepanek, line 96: I assume you're leaving these here as a TODO once you merge the branches together.

hmstepanek (approved):
Looks good. Before we release these I'd like to do some no harm tests with all of these bedrock and open ai implementations in case one of these models doesn't have a particular attribute or something but we can do that later.

Stuff like: `response_body["inputTextTokenCount"]` feels a bit unsafe to me. I think we tried to mostly do this already with openai but there might be some remaining issues in there since we don't have actual tests implemented for it yet.

PR #958, opened 2023-11-03
Add bedrock feedback support

lrafeei:
# Overview
Modify get_ai_message_ids API function to retrieve message IDs to be used in the customer's AI feedback endpoint.


tests/external_botocore/test_get_ai_message_ids.py
```diff
@@ -0,0 +1,70 @@
...
+@reset_core_stats_engine()
+def test_get_ai_message_ids_bedrock_chat_completion_in_txn(exercise_model):  # noqa: F811
+    @background_task()
+    def _test():
+        set_trace_info()
+        add_custom_attribute("conversation_id", "my-awesome-id")
+        exercise_model(prompt=_test_bedrock_chat_completion_prompt, temperature=0.7, max_tokens=100)
+        message_ids = [m for m in get_ai_message_ids()]
+
+        for message_id_info in message_ids:
```
hmstepanek, line 64: This one test, tests all the models which is great but the assertions will need to be catered towards the model under tests. I'd suggest creating another expected dict in the _test_chat_completion.py with the expected ai_message_ids for each model. I'd also suggest moving these tests into the test_chat_completion.py as I think it will be easier in the long run since all the fixtures and expected values are in there that are getting re-used here.

There are 3 primary test cases we need to cover here:

In the case where bedrock has a response_id (the Jurassic model), you should be able to make this a hardcoded check as all of these id's will be repeatable within the transaction.

In the case where bedrock does not provide a response_id (the titan model). Maybe make the value None in the expected dictionary and then in cases where it's None in this test's assertion check for presence like you are doing here instead of asserting the value. This is similar to what we do in some of our validators.

The case where there is no conversation_id set (aka conversation_id == "").

tests/external_botocore/test_get_ai_message_ids.py
```diff
@@ -0,0 +1,70 @@
...
+            # Check to see if this equals the conversation ID we added
+            assert message_id_info["conversation_id"] == "my-awesome-id"
+            # Check to see if this exists
+            assert message_id_info["message_id"]
+
+    _test()
```
hmstepanek, line 70: Make sure we are popping the value off of the transaction here when get_ai_message_ids is called so we are reducing the memory footprint so add the following assertion:
```suggestion
        assert transaction._nr_message_ids = {}

    _test()
```

hmstepanek (changes requested):
I had a couple changes. Primarily, we are missing the request_id in the tuple of ids returned when get_ai_message_ids is called.

I had some additional feedback about the tests to make sure we are covering everything.

newrelic/api/ml_model.py
```diff
@@ -33,3 +35,24 @@ def wrap_mlmodel(model, name=None, version=None, feature_names=None, label_names
...
+def get_ai_message_ids(response_id=None):
+    transaction = current_transaction()
+    if transaction:
+        nr_message_ids = getattr(transaction, "_nr_message_ids", {})
+        message_id_info = (
+            nr_message_ids.pop("bedrock_key", ()) if not response_id else nr_message_ids.pop(response_id, ())
+        )
+
+        if not message_id_info:
+            response_id_warning = "" if not response_id else " for %s" % response_id
```
hmstepanek, line 49:
```suggestion
            response_id_warning = "." if not response_id else " for %s." % response_id
```

newrelic/hooks/external_botocore.py
```diff
@@ -331,6 +346,12 @@ def handle_chat_completion_event(client, transaction, extractor, model, response
         conversation_id=conversation_id,
         response_id=response_id,
     )
+    if not hasattr(transaction, "_nr_message_ids"):
+        transaction._nr_message_ids = {}
+    if not response_id:
```
hmstepanek, line 351: We never index by the response_id in bedrock so this should just be:
```suggestion
```

newrelic/hooks/external_botocore.py
```diff
@@ -331,6 +346,12 @@ def handle_chat_completion_event(client, transaction, extractor, model, response
...
+        transaction._nr_message_ids["bedrock_key"] = message_ids
+    else:
```
hmstepanek, line 353:
```suggestion
```

newrelic/hooks/external_botocore.py
```diff
@@ -331,6 +346,12 @@ def handle_chat_completion_event(client, transaction, extractor, model, response
         conversation_id=conversation_id,
         response_id=response_id,
     )
+    if not hasattr(transaction, "_nr_message_ids"):
+        transaction._nr_message_ids = {}
+    if not response_id:
+        transaction._nr_message_ids["bedrock_key"] = message_ids
```
hmstepanek, line 352:
```suggestion
transaction._nr_message_ids["bedrock_key"] = message_ids
```

newrelic/hooks/external_botocore.py
```diff
@@ -331,6 +346,12 @@ def handle_chat_completion_event(client, transaction, extractor, model, response
...
+    else:
+        transaction._nr_message_ids[response_id] = message_ids
```
hmstepanek, line 354:
```suggestion
```

tests/external_botocore/_test_bedrock_chat_completion.py
```diff
@@ -1,10 +1,83 @@
...
+            },
+            {
+                "conversation_id": "my-awesome-id",
+                "request_id": "660d4de9-6804-460e-8556-4ab2a019d1e3",
+                "message_id": None,  # UUID that varies with each run
+            },
+        ]
+    },
+    "ai21.j2-mid-v1": {
+        "1234": [
```
hmstepanek, line 38:
```suggestion
        "bedrock_key": [
```

tests/external_botocore/_test_bedrock_chat_completion.py
```diff
@@ -1,10 +1,83 @@
...
+            },
+            {
+                "conversation_id": "my-awesome-id",
+                "request_id": "f354b9a7-9eac-4f50-a8d7-7d5d23566176",
+                "message_id": None,  # UUID that varies with each run
+            },
+        ]
+    },
+    "cohere.command-text-v14": {
+        "a9cc8ce6-50b6-40b6-bf77-cf24561d8de7": [
```
hmstepanek, line 66:
```suggestion
        "bedrock-key": [
```

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -154,3 +160,82 @@ def test_bedrock_chat_completion_outside_txn(set_trace_info, exercise_model):
...
+        expected_message_ids = [value for value in expected_ai_message_ids.values()][0]
+        if id == "bedrock_key":
+            message_ids = [m for m in get_ai_message_ids()]
+            for index, message_id_info in enumerate(message_ids):
+                expected_message_id_info = expected_message_ids[index]
+                assert message_id_info["conversation_id"] == expected_message_id_info["conversation_id"]
+                assert message_id_info["request_id"] == expected_message_id_info["request_id"]
+                assert message_id_info["message_id"]
+
+        else:
```
hmstepanek, line 201: Again, for bedrock you don't need this else statement because the customer is never required to pass in the response_id to get the message ids.

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -154,3 +160,82 @@ def test_bedrock_chat_completion_outside_txn(set_trace_info, exercise_model):
...
+):  # noqa: F811
+    @background_task()
+    def _test():
+        set_trace_info()
+        add_custom_attribute("conversation_id", "my-awesome-id")
+        exercise_model(prompt=_test_bedrock_chat_completion_prompt, temperature=0.7, max_tokens=100)
+
+        id = [key for key in expected_ai_message_ids.keys()][0]
+        expected_message_ids = [value for value in expected_ai_message_ids.values()][0]
+        if id == "bedrock_key":
```
hmstepanek, line 193:
```suggestion
```

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -154,3 +160,82 @@ def test_bedrock_chat_completion_outside_txn(set_trace_info, exercise_model):
...
+            message_ids = [m for m in get_ai_message_ids()]
+            for index, message_id_info in enumerate(message_ids):
+                expected_message_id_info = expected_message_ids[index]
+                assert message_id_info["request_id"] == expected_message_id_info["request_id"]
+                assert message_id_info["message_id"]
```
hmstepanek, line 230: The conversation id is always returned from the get_ai_message_ids call, even if it isn't set (aka empty string).
```suggestion
                assert message_id_info["message_id"]
                assert message_id_info["conversation_id"] == ""
```

hmstepanek, line 230:
```suggestion
                if expected_message_id_info["message_id"]: 
                    assert message_id_info["message_id"] == expected_message_id_info["message_id"]
                else: # If the expected message_id is None, just check presence rather than value.
                    assert message_id_info["message_id"]
```

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -154,3 +160,82 @@ def test_bedrock_chat_completion_outside_txn(set_trace_info, exercise_model):
...
+
+        id = [key for key in expected_ai_message_ids.keys()][0]
+        expected_message_ids = [value for value in expected_ai_message_ids.values()][0]
+        if id == "bedrock_key":
+            message_ids = [m for m in get_ai_message_ids()]
+            for index, message_id_info in enumerate(message_ids):
+                expected_message_id_info = expected_message_ids[index]
+                assert message_id_info["conversation_id"] == expected_message_id_info["conversation_id"]
+                assert message_id_info["request_id"] == expected_message_id_info["request_id"]
+                assert message_id_info["message_id"]
```
hmstepanek, line 199:
```suggestion
                if expected_message_id_info["message_id"]: 
                    assert message_id_info["message_id"] == expected_message_id_info["message_id"]
                else: # If the expected message_id is None, just check presence rather than value.
                    assert message_id_info["message_id"]
```

hmstepanek (changes requested):
In Bedrock the customer is never required to provide the response_id to get the message ids so the requested changes are centered around that. The response_id is required for OpenAI since OpenAI has async capabilities and we need to be able to distinguish one call's message ids from another's (since you could have multiple async calls before calling `get_ai_message_ids`). This is not the case in Bedrock since it doesn't support async so the customer doesn't have to pass an identifier in for us to know which bedrock call they want message ids for.


We recently decided to change the name of this public api function to `get_llm_message_ids` so while you're at it if you wouldn't mind updating the name in your PR, Uma is taking care of the rest of the name changes in her PR to openai.

Otherwise everything else looks good.

hmstepanek (approved):
Looks great!

PR #960, opened 2023-11-06
Add bedrock claude model support

hmstepanek:
Add bedrock claude model support.


TimPansino (approved):
Contents look good, merge conflicts still present though

PR #961, opened 2023-11-06
Add ingest source to openai events

hmstepanek:
* Since we are moving back to using custom events and instrumentation.provider is an ml event attribute, add an ingest_source attribute to the events in prep for that move.
* Remove duplicate test file.
* Cherry pick starlette test fix from main.


tests/mlmodel_openai/test_error.py
TimPansino, line 1: Is this file supposed to be deleted?

hmstepanek: Yes. It's a duplicate of the other two chat_completion_error and embedding_error test files.

TimPansino (approved):

PR #962, opened 2023-11-06
Pin openai tests to below 1.0

hmstepanek:
Openai released 1.0 which broke our tests as it moves instrumentation points around and changes some error classes. For now, pin openai tests to below 1.0 until we open a PR to instrument the new version.


TimPansino (approved):

umaannamalai (approved):

umaannamalai:
Merging - Starlette tests to be fixed in followup PR.

PR #964, opened 2023-11-07
Add record_llm_feedback_event API

umaannamalai:
This PR adds implementation for `record_llm_feedback_event`. This API is used to create machine learning events that capture detail about a chat completion message and its associated user feedback/ rating. It can also be used in conjunction with `get_llm_message_ids`


newrelic/api/ml_model.py
```diff
@@ -52,3 +53,34 @@ def get_ai_message_ids(response_id=None):
         return [{"conversation_id": conversation_id, "request_id": request_id, "message_id": _id} for _id in ids]
     warnings.warn("No message ids found. get_ai_message_ids must be called within the scope of a transaction.")
     return []
+
+
+def record_ai_feedback(
```
hmstepanek, line 58:
```suggestion
def record_ai_feedback_event(
```

umaannamalai: I saw it both ways in the CDD but it makes sense to converge on `record_ai_feedback_event` for consistency with other API names :)

newrelic/api/ml_model.py
```diff
@@ -52,3 +53,35 @@ def get_ai_message_ids(response_id=None):
...
+    if not transaction:
+        warnings.warn(
+            "No message feedback events will be recorded. record_ai_feedback must be called within the "
+            "scope of a transaction."
+        )
+        return
+
+    feedback_message_id = str(uuid.uuid4())
+    try:
+        rating = str(rating)
```
hmstepanek, line 71: This doesn't seem necessary to me but I could be missing something. Why are we casting this to a string?

newrelic/api/ml_model.py
```diff
@@ -52,3 +53,35 @@ def get_ai_message_ids(response_id=None):
...
+    except Exception:
+        pass
+
+    feedback_message_event = {
+        "id": feedback_message_id,
+        "message_id": message_id,
+        "rating": rating,
+        "metadata": metadata or "",
```
hmstepanek, line 79: I'd expect metadata to be dictionary of extra user attributes to attach to the event so it seems like it would be better to attach the metadata so that it appears at the same level as the rest of the attributes:
```
metadata = metadata or {}
feedback_message_event.update(metadata)
```

tests/agent_features/test_record_ai_feedback.py
```diff
@@ -0,0 +1,100 @@
...
+    @background_task()
+    def _test():
+        record_ai_feedback_event(
+            rating=1,
+            message_id="message_id",
+            category="informative",
+            request_id="request_id",
+            conversation_id="conversation_id",
+            message="message",
+            metadata="metadata",
```
hmstepanek:
```suggestion
            metadata={"key": "value"},
```

tests/agent_features/test_record_ai_feedback.py
```diff
@@ -0,0 +1,100 @@
...
+    _test()
+
+
+ai_feedback_required_args_recorded_events = [
+    (
+        {"type": "LlmFeedbackMessage"},
+        {
+            "id": None,
+            "category": "",
+            "rating": "1",
```
hmstepanek: I'm recommending this cause it's more realistic usage.
```suggestion
            "rating": "Good",
```

tests/agent_features/test_record_ai_feedback.py
```diff
@@ -0,0 +1,100 @@
...
+    ),
+]
+
+
+@reset_core_stats_engine()
+def test_record_ai_feedback_event_required_args_supplied():
+    @validate_ml_events(ai_feedback_required_args_recorded_events)
+    @background_task()
+    def _test():
+        record_ai_feedback_event(message_id="message_id", rating=1)
```
hmstepanek:
```suggestion
        record_ai_feedback_event(message_id="message_id", rating="Good")
```

tests/agent_features/test_record_ai_feedback.py
```diff
@@ -0,0 +1,100 @@
...
+        {
+            "id": None,
+            "category": "",
+            "rating": "1",
+            "message_id": "message_id",
+            "request_id": "",
+            "conversation_id": "",
+            "ingest_source": "Python",
+            "message": "",
+            "metadata": "",
```
hmstepanek:
```suggestion
```

tests/agent_features/test_record_ai_feedback.py
```diff
@@ -0,0 +1,100 @@
...
+        },
+    ),
+]
+
+
+@reset_core_stats_engine()
+def test_record_ai_feedback_event_required_args_supplied():
+    @validate_ml_events(ai_feedback_required_args_recorded_events)
+    @background_task()
+    def _test():
```
hmstepanek: I don't see a good reason to have this `def _test()` in here. I think we could move these validators outside the test and it would simplify things. Same with the other tests.

tests/agent_features/test_record_ai_feedback.py
```diff
@@ -0,0 +1,100 @@
...
+# limitations under the License.
+
+from testing_support.fixtures import reset_core_stats_engine
+from testing_support.validators.validate_ml_event_count import validate_ml_event_count
+from testing_support.validators.validate_ml_events import validate_ml_events
+
+from newrelic.api.background_task import background_task
+from newrelic.api.ml_model import record_ai_feedback_event
+
+ai_feedback_all_args_recorded_events = [
```
hmstepanek: Instead of defining these outside the validator, just put them inside the validators. I know we do this sometimes. In general, I'm not a huge fan of this pattern as it makes the tests harder to read IMO as you have to jump between the expected and the test. Sometimes it makes sense when it's re-used between tests or when it's super long but here there's no re-use and it's relatively short so I'd just move it into the validator as it will simplify the code and improve readability.

hmstepanek (changes requested):
Besides changing the name I had a couple minor requests.

newrelic/api/ml_model.py
umaannamalai: The thought was that the rating could be a string or int (numeric rating) so this would cast the rating to always be the same type. Given that it's only being casted for the sake of adding it to the feedback event and not touched further, we can keep the original type in tact and remove this.

tests/agent_features/test_record_ai_feedback.py
umaannamalai: Moved these into the tests. I kept the `_test()` function for the tests that checked the ML event attributes so it could be passed to the validator within the test itself.

hmstepanek (approved):
Looks great!

umaannamalai:
Merging - the test failures are for botocore which are being addressed in another PR.

PR #966, opened 2023-11-08
Bedrock Error Tracing

TimPansino:
# Overview

* Adds support for error tracing in AWS Bedrock
* Merges in commits from main to fix failing tests


hmstepanek (approved):
I think this implementation is fine. Assuming all exceptions have the same data available in them I don't see a good reason to test with different error conditions.

PR #969, opened 2023-11-09
Fix CI Image Permissions for Non-Root Users

TimPansino:
# Overview

* Move pyenv install location to shared directory for use by non-root users.
* Fix the Devcontainer's write permissions issues surrounding root ownership of `pyenv` directory.


TimPansino:
This apparently does not work with GitHub actions, according to https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions

lrafeei (approved):

PR #970, opened 2023-11-09
Package Version Performance Regression

### LINKED ISSUE #918: cpu blocked for ~10s before first request on complex Flask app with gevent on 8.10.1 release
Hi,

I have a pretty complex flask app locally that hangs for about ~10s ONLY before the first request when using 8.10.1 release (doesn't happen on 8.10.0). Still happens on 9.0.0 though. The CPU usage is 100% for those seconds. All the other requests work normally after.

How do you want me to profile it? Using ubuntu 22.04 & Python 3.11 and many libraries.

### LINKED ISSUE #927: Increased CPU usage since 8.11.0 / 8.10.1 release
After we upgraded from 8.10.0 to 8.11.0, we noticed an increase in CPU usage:
image
image

Yesterday, we went back to 8.10.0 on TST and it seemed to reduce CPU usage:
image

I'm hoping to get some pointers for profiling this issue, i can spend some time. So far we seem to be the only ones experiencing this, although #918 might be related.

From https://docs.newrelic.com/docs/release-notes/agent-release-notes/python-release-notes/python-agent-81100/ it's not clear which change may have caused this. It might also have been in 8.10.1, which we skipped. We also upgraded to 9.0.0, which shows the same increased CPU usage. 

**Your Environment**

We use newrelic in an e-commerce application based on Django-Oscar and Django REST Framework. 

We are hosting this on Amazon ECS, graphs are from its "Health and service" dashboard.

We have TST, ACC and PRD environments, and we see the issue on all three envs. The TST env is logging at DEBUG level, so its logs might contain useful info, if we knew what to look for. Searching on "newrelic" only gives messages like `newrelic.config DEBUG - register module ('rest_framework.views', 'newrelic.hooks.component_djangorestframework', 'instrument_rest_framework_views')`, which happens after the application starts up.

TimPansino:
# Overview

* Fixes performance issues with `get_package_version` by reverting #878.
* Avoid reintroducing DeprecationWarnings by suppressing warnings when accessing package version attributes.
  * Accessing these first for speed is intended behavior whether or not the version attribute is marked as deprecated. The agent is robust enough to handle the removal of these attributes, which makes the DeprecationWarnings irrelevant.

# Related Github Issues
* Closes #927
* Closes #918

### DIFF
--- a/tests/agent_unittests/test_package_version_utils.py
+++ b/tests/agent_unittests/test_package_version_utils.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import sys
+import warnings
 
 import pytest
 from testing_support.validators.validate_function_called import validate_function_called
@@ -66,10 +67,10 @@ def cleared_package_version_cache():
         ("version_tuple", [3, 1, "0b2"], "3.1.0b2"),
     ),
 )
-def test_get_package_version(attr, value, expected_value):
+def test_get_package_version(monkeypatch, attr, value, expected_value):
     # There is no file/module here, so we monkeypatch
     # pytest instead for our purposes
-    setattr(pytest, attr, value)
+    monkeypatch.setattr(pytest, attr, value, raising=False)
     version = get_package_version("pytest")
     assert version == expected_value
     delattr(pytest, attr)
@@ -77,18 +78,15 @@ def test_get_package_version(attr, value, expected_value):
 
 # This test only works on Python 3.7
 @SKIP_IF_IMPORTLIB_METADATA
-def test_skips_version_callables():
+def test_skips_version_callables(monkeypatch):
     # There is no file/module here, so we monkeypatch
     # pytest instead for our purposes
-    setattr(pytest, "version", lambda x: "1.2.3.4")
-    setattr(pytest, "version_tuple", [3, 1, "0b2"])
+    monkeypatch.setattr(pytest, "version", lambda x: "1.2.3.4", raising=False)
+    monkeypatch.setattr(pytest, "version_tuple", [3, 1, "0b2"], raising=False)
 
     version = get_package_version("pytest")
 
     assert version == "3.1.0b2"
-
-    delattr(pytest, "version")
-    delattr(pytest, "version_tuple")
 
 
 # This test only works on Python 3.7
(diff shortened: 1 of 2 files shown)


tests/agent_unittests/test_package_version_utils.py
```diff
@@ -66,30 +67,27 @@ def cleared_package_version_cache():
...
 )
-def test_get_package_version(attr, value, expected_value):
+def test_get_package_version(monkeypatch, attr, value, expected_value):
     # There is no file/module here, so we monkeypatch
     # pytest instead for our purposes
-    setattr(pytest, attr, value)
+    monkeypatch.setattr(pytest, attr, value, raising=False)
     version = get_package_version("pytest")
     assert version == expected_value
     delattr(pytest, attr)
```
hmstepanek, line 76:
```suggestion
```

tests/agent_unittests/test_package_version_utils.py
```diff
@@ -102,10 +100,10 @@ def test_skips_version_callables():
...
 )
-def test_get_package_version_tuple(attr, value, expected_value):
+def test_get_package_version_tuple(monkeypatch, attr, value, expected_value):
     # There is no file/module here, so we monkeypatch
     # pytest instead for our purposes
-    setattr(pytest, attr, value)
+    monkeypatch.setattr(pytest, attr, value, raising=False)
     version = get_package_version_tuple("pytest")
     assert version == expected_value
     delattr(pytest, attr)
```
hmstepanek, line 109:
```suggestion
```

tests/agent_unittests/test_package_version_utils.py
```diff
@@ -132,10 +130,26 @@ def test_pkg_resources_metadata():
...
+        return "3.2.1"
+    else:
+        raise NotImplementedError()
+
+
+def test_deprecation_warning_suppression(monkeypatch, recwarn):
+    # Add fake module to be deleted later
+    monkeypatch.setattr(pytest, "__getattr__", _getattr_deprecation_warning, raising=False)
+
+    assert get_package_version("pytest") == "3.2.1"
```
hmstepanek, line 145:
```suggestion
    assert get_package_version("pytest") == "3.2.1"
    
```

hmstepanek (changes requested):

umaannamalai (approved):

hmstepanek (approved):

PR #975, opened 2023-11-14
Add max_attribute_value setting

hmstepanek:
# Overview
Add config setting that increases the custom event limit's attribute value size.


lrafeei (commented):
Can we also cherry pick this commit, [a976e75](https://github.com/newrelic/newrelic-python-agent/pull/966/commits/a976e75888d44c46f83eb7be0cffc4f987d51741), into this PR so that the miscellaneous botocore tests pass?

newrelic/core/custom_event.py
```diff
@@ -55,25 +58,22 @@ def process_event_type(name):
         check_event_type_valid_chars(name)
 
     except NameIsNotStringException:
-        _logger.debug('Event type must be a string. Dropping '
-                'event: %r', name)
+        _logger.debug("Event type must be a string. Dropping " "event: %r", name)
```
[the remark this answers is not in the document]
hmstepanek: Yes! It looks like black reformatted these all onto the same line but didn't merge the quotes together.

newrelic/config.py
```diff
@@ -443,6 +443,7 @@ def _process_configuration(section):
     )
     _process_setting(section, "custom_insights_events.enabled", "getboolean", None)
     _process_setting(section, "custom_insights_events.max_samples_stored", "getint", None)
+    _process_setting(section, "custom_insights_events.max_attribute_value", "getint", 255)
```
umaannamalai, line 446: Should we modify the setting name to clarify that this is impacting the attribute length specifically? Maybe `custom_insights_events.max_attribute_length`?

lrafeei (approved):

newrelic/config.py
hmstepanek: Prob a discussion to be had on the spec: https://source.datanerd.us/agents/agent-specs/pull/648. We do need to be specific that this is the value not the key though.

umaannamalai: Yeah maybe `custom_insights_events.max_attribute_value_length`? We can move the discussion to the spec.

PR #977, opened 2023-11-15
Set transaction_id = guid not _transaction_id

hmstepanek:
_transaction_id is the identity of the transaction object and is used in `__exit__` as an early exit condition. guid is actually the transactionId that we see in the UI.


tests/external_botocore/conftest.py
```diff
@@ -154,6 +154,7 @@ def bind__do_get_response(request, operation_model, context):
 def set_trace_info():
     def _set_trace_info():
         txn = current_transaction()
+        txn.guid = "transaction-id"
```
umaannamalai, line 157: Should we move this into the `if txn` block?

umaannamalai (approved):

PR #980, opened 2023-11-15
Docker CGroups v2 Utilization Support

TimPansino:
# Overview

* Add support for sending utilization data for docker with cgroups v2.

### DIFF
--- a/newrelic/common/utilization.py
+++ b/newrelic/common/utilization.py
@@ -265,13 +265,16 @@ class PCFUtilization(CommonUtilization):
 class DockerUtilization(CommonUtilization):
     VENDOR_NAME = 'docker'
     EXPECTED_KEYS = ('id',)
-    METADATA_FILE = '/proc/self/cgroup'
-    DOCKER_RE = re.compile(r'([0-9a-f]{64,})')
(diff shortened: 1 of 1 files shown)


newrelic/common/utilization.py
```diff
@@ -265,21 +265,42 @@ def get_values(cls, response):
...
                     cgroup = stripped.split(':')
                     if len(cgroup) != 3:
                         continue
                     subsystems = cgroup[1].split(',')
                     if 'cpu' in subsystems:
-                        return cgroup[2]
+                        contents = cgroup[2].split('/')[-1]
+                        match = cls.METADATA_RE_CGROUPS_V1.search(contents)
+                        if match:
+                            return match.group(1)
```
hmstepanek, line 290: It looks like this was grabbing `match.group(0)` before. Should this still be `match.group(0)`?

hmstepanek (commented):
It seems like there might be a bug in the v1. I've noted it below.

newrelic/common/utilization.py
TimPansino: Yeah those were equivalent but unnecessary, so I put it back to 0.

hmstepanek (approved):

PR #981, opened 2023-11-15
Guard metadata logic

hmstepanek:
# Overview
Guard against the user overriding builtin attributes with the metadata.


newrelic/api/ml_model.py
```diff
@@ -67,18 +67,19 @@ def record_llm_feedback_event(
...
-        "rating": rating,
-        "conversation_id": conversation_id or "",
-        "request_id": request_id or "",
-        "category": category or "",
-        "message": message or "",
-        "ingest_source": "Python",
-    }
-    feedback_message_event.update(metadata)
+    feedback_message_event = {}
+    feedback_message_event.update(metadata or {})
```
TimPansino, line 71:
```suggestion
    feedback_message_event = metadata.copy() if metadata else {}
```

This is probably faster, and slightly more readable.

TimPansino (approved):

PR #982, opened 2023-11-15
Add package_capturing.enabled setting

hmstepanek:
# Overview
Add `package_capturing.enabled` config setting that allows the customer to disable capturing of the package and version info when they need to (aka when they have a large number of packages and the capturing causes a large CPU spike on startup).

# Related Github Issue
This is a fix for https://github.com/newrelic/newrelic-python-agent/issues/918.


lrafeei (approved):
🚀

TimPansino:
As my random bike-shed contribution, I was thinking this setting should go under maybe an environment tab? Maybe like `environment.package_version_reporting` or something? Not sure if there's precedent, but just to match expectations with the UI's environment tab.

umaannamalai (approved):

TimPansino:
The actual setting name is `package_reporting.enabled` for future reference. The PR title was out of date.

PR #987, opened 2023-11-21
Refactor OpenAI Error Tracing

umaannamalai:
This PR refactors error tracing for OpenAI with the following changes:

- Only pass error attrs to `notice_error()`

- Attach `completion_id`/ `embedding_id` to error trace to distinguish AI errors

- Send chat completion and embedding events on error as well

- Add `error: True` to chat completion summary/ embedding events

- Add `is_response` attr to chat completion messages that are LLM responses


newrelic/hooks/mlmodel_openai.py
```diff
@@ -54,46 +53,75 @@ def wrap_embedding_create(wrapped, instance, args, kwargs):
...
+                "error.param": getattr(exc, "param", ""),
+            }
+            exc._nr_message = notice_error_attributes.pop("error.message")
             ft.notice_error(
-                attributes=error_attributes,
+                attributes=notice_error_attributes,
             )
+            # Gather attributes to add to embedding summary event in error context
+            error_summary_attributes = {
+                "response.organization": getattr(exc, "organization", ""),
```
umaannamalai, line 105: Note to reviewer to chime in: It feels strange to call this `response.organization` even though we didn't get in from the response in the error case. We could change this to just be `organization` for all summary events.

tests/mlmodel_openai/test_embeddings_error.py
```diff
@@ -242,23 +321,52 @@ def test_embeddings_authentication_error_async(loop, monkeypatch):
...
+)
+@validate_span_events(
+   exact_agents={
+       "error.message": "Incorrect API key provided: DEADBEEF. You can find your API key at https://platform.openai.com/account/api-keys.",
+   }
+)
+@validate_custom_events(embedding_invalid_key_error_events)
+@validate_custom_event_count(count=1)
+@background_task()
+def test_embeddings_wrong_api_key_error_async(loop, monkeypatch, set_trace_info):
```
hmstepanek, line 366: It looks like this is a duplicated test?

tests/mlmodel_openai/test_embeddings_error.py
```diff
@@ -65,27 +108,46 @@ def test_embeddings_invalid_request_error_no_model():
...
+            "embedding_id": None,  # UUID that varies with each run
+            "appName": "Python Agent Test (mlmodel_openai)",
+            "transaction_id": "transaction-id",
+            "span_id": "span-id",
+            "trace_id": "trace-id",
+            "input": "Invalid API key.",
+            "api_key_last_four_digits": "",
+            "duration": None,  # Response time varies each test run
+            "request.model": "text-embedding-ada-002", # No model in this test case
+            "response.organization": None,
```
hmstepanek, line 143: Same thing here. We should probably be asserting this is empty string not None.

newrelic/hooks/mlmodel_openai.py
hmstepanek: It is still coming from a response in a way it's just the error response but I see your point. If we change it though, the UI will have to change with it so for now I'd just leave it. It's definitely worth having the discussion on these names before GA though since it will be really hard to change them after that.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -54,46 +53,75 @@ def wrap_embedding_create(wrapped, instance, args, kwargs):
...
+    trace_id = available_metadata.get("trace.id", "")
+
+    settings = transaction.settings if transaction.settings is not None else global_settings()
+
+    # Shared attributes that will be included on all embedding events (error or not)
+    base_embedding_dict = {
+        "embedding_id": embedding_id,
+        "appName": settings.app_name,
+        "api_key_last_four_digits": api_key_last_four_digits,
+        "span_id": span_id,
```
hmstepanek, line 79: Is there a bug in this code right now where we are capturing the wrong span id? Shouldn't we be capturing the span id inside the function trace? I just looked at the SDK code and they attach the function trace's span id.

hmstepanek: I'm not sure how this would work in practice but what I was imaging for this design and what I would propose is that this code wouldn't change this much from the previous version and that our .get(attr_name, "") fallback logic would handle most of the cases where things don't exist because an error occurred. I'd propose maybe approaching the implementation from that perspective and see if that works out better?

hmstepanek (changes requested):
This is a good first whack at this. I noted a couple areas in the expected values for the exceptions that need to be changed-there are a lot more that I didn't specifically call out that are along the same lines. I'd prefer to see less drastic changes to the hook code (as I noted I'm not sure if this will work out in practice but maybe give it a try and see if it ends up being simpler).
There are quite a few linter errors that we might want to take a look at as well.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -268,10 +346,21 @@ def create_chat_completion_message_event(
     response_id,
     request_id,
     conversation_id,
+    choices_message,
 ):
     message_ids = []
     for index, message in enumerate(message_list):
```
hmstepanek, line 352: So we'd have two for loops here instead of just the one; one for choices and one for messages and the choices would have the is_response attr set.

hmstepanek (changes requested):
I had couple small fixups to simplify the logic. I also think we shouldn't be trying to handle feedback in the case that an error occurs but I've put a message in the slack channel asking everyone what they think as well.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -209,28 +286,29 @@ def wrap_chat_completion_create(wrapped, instance, args, kwargs):
...
     }
 
-    transaction.record_custom_event("LlmChatCompletionSummary", chat_completion_summary_dict)
+    transaction.record_custom_event("LlmChatCompletionSummary", full_chat_completion_summary_dict)
 
     message_list = list(messages)
+    choices_message = None
     if choices:
-        message_list.extend([choices[0].message])
+        choices_message = choices[0].message
```
hmstepanek, line 297: I think it's easier at this point to keep the choices list and the messages list separtate as that way we can tell which is sent vs received.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -27,73 +27,86 @@
...
+                "appName": settings.app_name,
+                "api_key_last_four_digits": api_key_last_four_digits,
+                "span_id": span_id,
+                "trace_id": trace_id,
+                "transaction_id": transaction.guid,
+                "input": request_args.get("input", ""),
+                "request.model": request_args.get("model") or request_args.get("engine") or "",
+                "vendor": "openAI",
+                "ingest_source": "Python",
+                "response.organization": "" if exc_organization is None else exc_organization,
```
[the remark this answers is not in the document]
umaannamalai: The reason I added this was because the `organization` attribute is found on the exception but it is `None` (at least for our test cases). The default of "" isn't returned since there was no attribute error so this is attempting to coerce that behavior.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -133,61 +145,119 @@ def wrap_chat_completion_create(wrapped, instance, args, kwargs):
...
+
+            create_chat_completion_message_event(
+                transaction,
+                app_name,
+                request_message_list,
+                completion_id,
+                span_id,
+                trace_id,
+                "",
+                error_response_id,
```
hmstepanek, line 219: We already have code inside create_message to default this correctly if it doesn't exist so just let that code run, otherwise we end up with uuid-0 instead of just uuid.
```suggestion
                None,
```

newrelic/hooks/mlmodel_openai.py
```diff
@@ -133,61 +145,119 @@ def wrap_chat_completion_create(wrapped, instance, args, kwargs):
...
+                "request.max_tokens": kwargs.get("max_tokens", ""),
+                "vendor": "openAI",
+                "ingest_source": "Python",
+                "response.organization": "" if exc_organization is None else exc_organization,
+                "duration": ft.duration,
+                "error": True,
+            }
+            transaction.record_custom_event("LlmChatCompletionSummary", error_chat_completion_dict)
+
+            error_response_id = str(uuid.uuid4())
```
hmstepanek, line 209:
```suggestion
```

newrelic/hooks/mlmodel_openai.py
```diff
@@ -383,69 +530,120 @@ async def wrap_chat_completion_acreate(wrapped, instance, args, kwargs):
...
+
+            create_chat_completion_message_event(
+                transaction,
+                app_name,
+                request_message_list,
+                completion_id,
+                span_id,
+                trace_id,
+                "",
+                error_response_id,
```
hmstepanek, line 604:
```suggestion
                None,
```

newrelic/hooks/mlmodel_openai.py
```diff
@@ -383,69 +530,120 @@ async def wrap_chat_completion_acreate(wrapped, instance, args, kwargs):
...
+                "request.max_tokens": kwargs.get("max_tokens", ""),
+                "vendor": "openAI",
+                "ingest_source": "Python",
+                "response.organization": "" if exc_organization is None else exc_organization,
+                "duration": ft.duration,
+                "error": True,
+            }
+            transaction.record_custom_event("LlmChatCompletionSummary", error_chat_completion_dict)
+
+            error_response_id = str(uuid.uuid4())
```
hmstepanek, line 594:
```suggestion
```

hmstepanek (approved):
Thank you for sticking it out through this review process. It's been a lot of back and forth! There's only one minor issue here so I'm just gonna pre-approve this and assume you'll make that change before merging.

PR #991, opened 2023-12-04
Refactor Bedrock Error Tracing

umaannamalai:
This PR refactors error tracing for Bedrock with the following changes:

- Only pass error attrs to `notice_error()`

- Attach `completion_id`/ `embedding_id` to error trace to distinguish AI errors

- Send chat completion and embedding events on error as well

- Add `error: True` to chat completion summary/ embedding events

Add `is_response` attr to chat completion messages that are LLM responses
exercise your changes.


tests/external_botocore/_test_bedrock_embeddings.py
```diff
@@ -21,7 +21,7 @@
                 "request_id": "11233989-07e8-4ecb-9ba6-79601ba6d8cc",
                 "response.usage.total_tokens": 6,
                 "response.usage.prompt_tokens": 6,
-                "vendor": "bedrock",
+                "vendor": "Bedrock",
```
hmstepanek, line 24: Sounds like we are leaving these as lower case bedrock.

tests/external_botocore/_test_bedrock_embeddings.py
```diff
@@ -43,30 +43,65 @@
...
-        "request_id": "aece6ad7-e2ff-443b-a953-ba7d385fd0cc",
-        "api_key_last_four_digits": "-KEY",
-        "request.model": "amazon.titan-embed-text-v1",
-        "vendor": "Bedrock",
-        "ingest_source": "Python",
         "http.statusCode": 403,
         "error.message": "The security token included in the request is invalid.",
         "error.code": "UnrecognizedClientException",
     },
     "amazon.titan-embed-g1-text-02": {
```
hmstepanek, line 104: We need to add the embedding_id's into these.

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -180,21 +186,24 @@ def test_bedrock_chat_completion_disabled_settings(set_trace_info, exercise_mode
...
+            "transaction_id": "transaction-id",
             "conversation_id": "my-awesome-id",
-            "request_id": "f4908827-3db9-4742-9103-2bbc34578b03",
+            "span_id": None,
+            "trace_id": "trace-id",
             "api_key_last_four_digits": "CRET",
+            "duration": None,  # Response time varies each test run
             "request.model": "does-not-exist",
+            "response.model": "does-not-exist",
+            "request_id": "",
```
hmstepanek, line 203: Why did the request_id get removed here?

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -209,22 +218,39 @@ def test_bedrock_chat_completion_disabled_settings(set_trace_info, exercise_mode
...
-            accept="application/json",
-            contentType="application/json",
-        )
+    @validate_custom_events(chat_completion_invalid_model_error_events)
+    @validate_error_trace_attributes(
+        "botocore.errorfactory:ValidationException",
+        exact_attrs={
+            "agent": {},
+            "intrinsic": {},
+            "user": {
```
hmstepanek, line 227: We should be checking for completion_id here too.

hmstepanek (changes requested):
A couple minor tweaks but otherwise looks good.

tests/external_botocore/test_bedrock_chat_completion.py
umaannamalai: Totally agree with you. I had added that in previously but given the `completion_id` is UUID and will differ on each test run, I set this to `None` but this validator does accept that behavior. It just causes the whole test to fail expecting the ID to be the specific UUID for that test run.

PR #992, opened 2023-12-04
Use named function traces

hmstepanek:
# Overview
Use named function traces in OpenAI:
* `Llm/completion/OpenAI/<function name>`
* `Llm/embedding/OpenAI/<function name>`

Use named function traces in Bedrock:
* `Llm/completion/Bedrock/<function name>`
* `Llm/embedding/Bedrock/<function name>`

### DIFF
--- a/newrelic/hooks/mlmodel_openai.py
+++ b/newrelic/hooks/mlmodel_openai.py
@@ -48,7 +48,6 @@ def wrap_embedding_create(wrapped, instance, args, kwargs):
 
     settings = transaction.settings if transaction.settings is not None else global_settings()
 
-
     function_name = wrapped.__name__
 
     with FunctionTrace(name=function_name, group="Llm/embedding/OpenAI") as ft:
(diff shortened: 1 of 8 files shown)


newrelic/hooks/mlmodel_openai.py
```diff
@@ -432,8 +433,9 @@ async def wrap_embedding_acreate(wrapped, instance, args, kwargs):
 
     settings = transaction.settings if transaction.settings is not None else global_settings()
 
-    ft_name = callable_name(wrapped)
-    with FunctionTrace(ft_name) as ft:
+    function_name = wrapped.__name__
```
umaannamalai, line 436: With OpenAI v1, all functions are named `create` and there's no concept of `acreate` because the clients themselves become sync or async so we won't be able to distinguish between the two. It may be something we consider for the OpenAI v1 PRs.

TimPansino (approved):

PR #993, opened 2023-12-04
Update wrapt

TimPansino:
# Overview

* Update `wrapt` to v1.16.0
* Add tests for wrapt function wrapper attributes
* Update older duplicate object wrapper functions to use functions imported from wrapt.
* Update inheritance tree of `newrelic.common.object_wrapper.ObjectWrapper` to fix issues with newer versions of wrapt.

### DIFF
--- a/newrelic/common/object_wrapper.py
+++ b/newrelic/common/object_wrapper.py
@@ -24,6 +24,12 @@ import inspect
 
 from newrelic.packages import six
 
+from newrelic.packages.wrapt import (
+    resolve_path,
+    apply_patch,
+    wrap_object,
+    wrap_object_attribute,
+)
 from newrelic.packages.wrapt import (ObjectProxy as _ObjectProxy,
         FunctionWrapper as _FunctionWrapper,
         BoundFunctionWrapper as _BoundFunctionWrapper)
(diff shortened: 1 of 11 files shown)


tests/agent_unittests/test_wrappers.py
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from newrelic.common.object_wrapper import function_wrapper
+
+
+@function_wrapper
```
hmstepanek, line 18: This should really be a fixture:
```suggestion
@pytest.fixture
def wrapper():
    @function_wrapper
    def wrapper(wrapped, instance, args, kwargs):
        return wrapped(*args, **kwargs)
        
    return wrapper
```

TimPansino: After breaking the big test function apart I made these into fixtures anyway

newrelic/common/object_wrapper.py
```diff
@@ -96,166 +106,45 @@ def _nr_parent(self):
...
-            self._self_last_object = getattr(self.__wrapped__,
-                    '_nr_last_object', self.__wrapped__)
-            return self._self_last_object
-
-
-class CallableObjectProxy(ObjectProxy):
+class CallableObjectProxy(ObjectProxy, _CallableObjectProxy):
+    pass
 
-    def __call__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
TimPansino: Yep, this code was originally just duplicated from wrapt but instead it inherits these methods now. I added `_CallableObjectProxy` to the inheritance tree.

PR #996, opened 2023-12-06
Deprecate ObjectWrapper API

TimPansino:
# Overview
## Context 

`ObjectWrapper` was previously used in instrumentation but was simply a custom and more manual version of `FunctionWrapper` from `wrapt`. There is no functional difference in how it is used today in the agent and should be deprecated and removed.

## Changes

* Deprecate `ObjectWrapper` API.
* Add `CallableObjectProxy` to public APIs to replace some of the functionality of `ObjectWrapper`
* Add missing `NewRelicContextFormatter` supportability wrapping in `newrelic.agent`.
* Remove instrumentation references to `ObjectWrapper`, replacing with `FunctionWrapper` or `ObjectProxy`.
* Remove custom `ObjectWrapper` implementation and instead create a pass through to `FunctionWrapper`.


hmstepanek (changes requested):
In general this looks good. I had one question about the import path change. It looks correct to me but I think we should change all of them or none of them as there are a lot more occurrences than just this one.

newrelic/hooks/framework_webpy.py
```diff
@@ -21,7 +21,7 @@
 import newrelic.api.in_function
 import newrelic.api.out_function
 import newrelic.api.pre_function
-from newrelic.api.object_wrapper import callable_name
+from newrelic.common.object_wrapper import callable_name
```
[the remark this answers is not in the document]
TimPansino: I've changed all these references now to point to the new object wrapper location

newrelic/hooks/framework_pylons.py
```diff
@@ -16,14 +16,14 @@
 import newrelic.api.transaction_name
 import newrelic.api.function_trace
 import newrelic.api.error_trace
-import newrelic.api.object_wrapper
+import newrelic.common.object_wrapper
```
hmstepanek, line 19: All of these callable name usages should import from `newrelic.common.object_names`? At this point, callable_name isn't even imported in object_wrapper. The pylint linter caught this issue if you look at the megalinter logs. The tests didn't fail though cause we don't actually have tests for this file.

hmstepanek (changes requested):
It looks like we are importing callable_name from object_wrappers when it should be importing from object_names. Note the linter caught this issue but the tests did not. One of those rare cases where the linter caught something super important.

newrelic/hooks/framework_pylons.py
hmstepanek: I just pushed up a couple fixes for this.

hmstepanek (approved):
I think this looks good now. Maybe just have a quick review of the commits I made before merging as a double check?

PR #1002, opened 2023-12-11
Add support for OpenAI v1 embeddings 

umaannamalai:
This PR adds instrumentation for OpenAI v1 embeddings. Embedding errors support for v1 will be in a subsequent PR.


hmstepanek (commented):
A couple small requests otherwise looks good!

newrelic/hooks/mlmodel_openai.py
```diff
@@ -93,25 +97,38 @@ def wrap_embedding_create(wrapped, instance, args, kwargs):
...
+    available_metadata = get_trace_linking_metadata()
+    span_id = available_metadata.get("span.id", "")
+    trace_id = available_metadata.get("trace.id", "")
+    embedding_id = str(uuid.uuid4())
+
     request_id = response_headers.get("x-request-id", "") if response_headers else ""
+    response_model = response.get("model", "")
+    response_usage = response.get("usage", {})
+    api_type = getattr(response, "api_type", "")
+    api_key = getattr(instance._client, "api_key", "") if OPENAI_V1 else getattr(response, "api_key", "")
```
hmstepanek, line 115: Should this api key logic go up above with the other api key logic on line 45?

hmstepanek, line 115: Let's move this up above to the rest of the openai logic. It looks like in v0 we were getting the key from openai not from the response so we should probably keep that the same.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -484,25 +496,33 @@ async def wrap_embedding_acreate(wrapped, instance, args, kwargs):
     if not response:
         return response
 
-    response_model = response.get("model", "")
-    response_usage = response.get("usage", {})
     response_headers = getattr(response, "_nr_response_headers", None)
+
+    # In v1, response objects are pydantic models so this function call converts the object back to a dictionary for backwards compatibility
+    if OPENAI_V1:
+        response = response.model_dump()
```
hmstepanek, line 503: Just noticed this bug while running tests locally. We'll need to put this into a different variable as this currently changes the response of the wrapped method to `response.model_dump()`.

hmstepanek (approved):

PR #1005, opened 2023-12-13
OpenAI v1 embeddings errors

umaannamalai:
# Overview
Add support for openai v1 embedding error traces and events.

### DIFF
--- a/tests/mlmodel_openai/conftest.py
+++ b/tests/mlmodel_openai/conftest.py
@@ -65,7 +65,6 @@ else:
         "test_chat_completion.py",
         "test_get_llm_message_ids.py",
         "test_chat_completion_error.py",
-        "test_embeddings_error_v1.py",
     ]
 
 
--- a/tests/mlmodel_openai/conftest.py
+++ b/tests/mlmodel_openai/conftest.py
@@ -68,7 +68,6 @@ else:
         "test_embeddings_error_v1.py",
     ]
 
-
 OPENAI_AUDIT_LOG_FILE = os.path.join(os.path.realpath(os.path.dirname(__file__)), "openai_audit.log")
 OPENAI_AUDIT_LOG_CONTENTS = {}
 # Intercept outgoing requests and log to file for mocking
--- a/tests/mlmodel_openai/conftest.py
+++ b/tests/mlmodel_openai/conftest.py
@@ -68,6 +68,7 @@ else:
         "test_embeddings_error_v1.py",
     ]
 
+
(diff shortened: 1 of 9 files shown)


tests/mlmodel_openai/test_embeddings_error_v1.py
```diff
@@ -12,17 +12,305 @@
...
+    exact_attrs={
+        "agent": {},
+        "intrinsic": {},
+        "user": {},
+    },
+)
+@validate_span_events(
+    exact_agents={
+        "error.message": "create() missing 1 required keyword-only argument: 'model'"
+        if sys.version_info < (3, 10)
```
hmstepanek, line 72: Ugh that's annoying.

tests/mlmodel_openai/test_embeddings_error_v1.py
```diff
@@ -12,17 +12,305 @@
...
+
+
+@dt_enabled
+@reset_core_stats_engine()
+@validate_error_trace_attributes(
+    callable_name(openai.InternalServerError),
+    exact_attrs={
+        "agent": {},
+        "intrinsic": {},
+        "user": {},
```
hmstepanek, line 160: I'd expect a minimum to have a status code in here but also usually the error.code is populated as well. I'm guessing these were removed cause the api was miss-behaving yesterday but since we are mocking these should be consistent.
```suggestion
        "user": {
            'http.statusCode': 404,
            'error.code': 'model_not_found',
        },
```

tox.ini
```diff
@@ -495,4 +495,4 @@ source =
 directory = ${TOX_ENV_DIR-.}/htmlcov
 
 [coverage:xml]
-output = ${TOX_ENV_DIR-.}/coverage.xml
+output = ${TOX_ENV_DIR-.}/coverage.xml
```
hmstepanek, line 498: This looks like a no-op change-just a line ending so just remove this tox.ini from the changes.

tests/mlmodel_openai/conftest.py
```diff
@@ -65,7 +65,6 @@
         "test_chat_completion.py",
         "test_get_llm_message_ids.py",
         "test_chat_completion_error.py",
```
hmstepanek, line 67: I made a commit to remove this line. I had added it for my PR since these tests were failing in my branch. I think because the tests weren't actually running before and now they are there's some errors that need to be fixed in them.

hmstepanek (changes requested):
Mainly just need to fix up the tests now that they are actually running and there's a couple minor fixups in the hook file to handle None cases.

hmstepanek (approved):
Looks good!

PR #1006, opened 2023-12-13
Add support for openai v1 completions

hmstepanek:
# Overview
Adds support for openai v1 completions
* Move wrapper logic for request capturing of mock server into lower level _request method where we can capture both error and successful responses.
* Ignore embedding error tests for v1 for now since they haven't been implemented yet.
* Adds v1 tests for successful and error responses for v1 chat completions for both async and sync.
* Adds v1 tests for message id capturing and feedback.
* Change v1 tests in tox to test latest.


tests/mlmodel_openai/test_chat_completion_error_v1.py
```diff
@@ -0,0 +1,414 @@
...
+        sync_openai_client.chat.completions.create(
+            messages=_test_openai_chat_completion_messages, temperature=0.7, max_tokens=100
+        )
+
+
+@dt_enabled
+@reset_core_stats_engine()
+@validate_error_trace_attributes(
+    callable_name(TypeError),
+    exact_attrs={
```
umaannamalai, line 140: We may want to add in the user attributes here.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -192,18 +191,38 @@ def wrap_chat_completion_create(wrapped, instance, args, kwargs):
...
-            exc_organization = getattr(exc, "organization", "")
+            if OPENAI_V1:
+                response = getattr(exc, "response", "")
+                response_headers = getattr(response, "headers", "")
+                exc_organization = response_headers.get("openai-organization", "") if response_headers else ""
+                # There appears to be a bug here in openai v1 where despite having code,
+                # param, etc in the error response, they are not populated on the exception
+                # object so grab them from the response object instead.
+                content = getattr(response, "content", b"{}")
+                response = json_decode(content.decode("utf-8")).get("error", {})
```
umaannamalai, line 204: It looks like the status code and other notice error attributes may be available on `exc.body` `exc.message`.

hmstepanek: I've changed it to use the body. The exc.message is not the message we are looking for. 😂 It includes the json inside of it which is the same as the default message that we would grab without overriding it so I'm using the message from the body instead still.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -192,18 +190,37 @@ def wrap_chat_completion_create(wrapped, instance, args, kwargs):
...
+                response = getattr(exc, "response", "")
+                response_headers = getattr(response, "headers", "")
+                exc_organization = response_headers.get("openai-organization", "") if response_headers else ""
+                # There appears to be a bug here in openai v1 where despite having code,
+                # param, etc in the error response, they are not populated on the exception
+                # object so grab them from the response body object instead.
+                body = getattr(exc, "body", {})
+                notice_error_attributes = {
+                    "http.statusCode": getattr(exc, "status_code", "") or "",
+                    "error.message": body.get("message", "") or "",
```
umaannamalai, line 205: Should we check if body has anything in it to avoid NoneType errors?

tests/mlmodel_openai/test_chat_completion_error_v1.py
```diff
@@ -0,0 +1,416 @@
...
+
+from newrelic.api.background_task import background_task
+from newrelic.api.transaction import add_custom_attribute
+from newrelic.common.object_names import callable_name
+
+_test_openai_chat_completion_messages = (
+    {"role": "system", "content": "You are a scientist."},
+    {"role": "user", "content": "What is 212 degrees Fahrenheit converted to Celsius?"},
+)
+
```
umaannamalai, line 39: In v0, we also had the test case of no API key being supplied at all - it may be good to add that here so we're consistent with tests cases across versions.

hmstepanek: It just throws the same error as the api_key = "DEADBEEF" case except it's None so I figured we didn't really need to test for that as it's not different.

umaannamalai (approved):

PR #1008, opened 2023-12-15
Log Attribute Filtering

TimPansino:
# Overview

* Add filtering to context data from log events leveraging existing attribute system.

### DIFF
--- a/newrelic/core/attribute.py
+++ b/newrelic/core/attribute.py
@@ -184,9 +184,6 @@ def resolve_logging_context_attributes(attr_dict, attribute_filter, attr_prefix,
     c_attrs = attr_class()
 
     for attr_name, attr_value in attr_dict.items():
-        if attr_value is None:
-            continue
-
         dest = attribute_filter.apply(attr_name, DST_LOG_EVENT_CONTEXT_DATA)
 
         if dest & DST_LOG_EVENT_CONTEXT_DATA:
--- a/newrelic/core/attribute.py
+++ b/newrelic/core/attribute.py
@@ -18,6 +18,7 @@ from collections import namedtuple
 from newrelic.core.attribute_filter import (
     DST_ALL,
     DST_ERROR_COLLECTOR,
+    DST_LOG_EVENT_CONTEXT_DATA,
     DST_SPAN_EVENTS,
     DST_TRANSACTION_EVENTS,
     DST_TRANSACTION_SEGMENTS,
@@ -172,6 +173,35 @@ def resolve_agent_attributes(attr_dict, attribute_filter, target_destination, at
             a_attrs[attr_name] = attr_value
 
     return a_attrs
+
+
+def resolve_logging_context_attributes(attr_dict, attribute_filter, attr_prefix, attr_class=dict):
+    """
+    Helper function for processing logging context attributes that require a prefix. Correctly filters attribute names
+    before applying the required prefix, and then applies the process_user_attribute after the prefix is applied to
+    correctly check length requirements.
+    """
+    c_attrs = attr_class()
+
+    for attr_name, attr_value in attr_dict.items():
+        if attr_value is None:
+            continue
+
+        dest = attribute_filter.apply(attr_name, DST_LOG_EVENT_CONTEXT_DATA)
+
+        if dest & DST_LOG_EVENT_CONTEXT_DATA:
+            try:
+                attr_name, attr_value = process_user_attribute(attr_prefix + attr_name, attr_value)
+                if attr_name:
+                    c_attrs[attr_name] = attr_value
+            except Exception:
+                _logger.debug(
+                    "Log event context attribute failed to validate for unknown reason. Dropping context attribute: %s. Check traceback for clues.",
+                    attr_name,
+                    exc_info=True,
+                )
+
+    return c_attrs
 
 
 def create_user_attributes(attr_dict, attribute_filter):
--- a/newrelic/core/config.py
+++ b/newrelic/core/config.py
@@ -902,10 +902,10 @@ _settings.application_logging.forwarding.enabled = _environ_as_bool(
 _settings.application_logging.forwarding.context_data.enabled = _environ_as_bool(
     "NEW_RELIC_APPLICATION_LOGGING_FORWARDING_CONTEXT_DATA_ENABLED", default=False
 )
-_settings.application_logging.forwarding.context_data.include = _environ_as_bool(
+_settings.application_logging.forwarding.context_data.include = _environ_as_set(
     "NEW_RELIC_APPLICATION_LOGGING_FORWARDING_CONTEXT_DATA_INCLUDE", default=""
 )
-_settings.application_logging.forwarding.context_data.exclude = _environ_as_bool(
+_settings.application_logging.forwarding.context_data.exclude = _environ_as_set(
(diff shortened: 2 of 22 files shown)


newrelic/core/attribute.py
```diff
@@ -174,6 +175,35 @@ def resolve_agent_attributes(attr_dict, attribute_filter, target_destination, at
...
+def resolve_logging_context_attributes(attr_dict, attribute_filter, attr_prefix, attr_class=dict):
+    """
+    Helper function for processing logging context attributes that require a prefix. Correctly filters attribute names
+    before applying the required prefix, and then applies the process_user_attribute after the prefix is applied to
+    correctly check length requirements.
+    """
+    c_attrs = attr_class()
+
+    for attr_name, attr_value in attr_dict.items():
+        if attr_value is None:
```
hmstepanek, line 187: If we are adding user attributes to a transaction, we would capture an attribute {"foo": None} on the transaction as {"foo": "None"}. Here however, we drop the transaction instead. It seems like these should behave the same in both scenarios?

newrelic/core/config.py
```diff
@@ -902,6 +902,12 @@ def default_otlp_host(host):
 _settings.application_logging.forwarding.context_data.enabled = _environ_as_bool(
     "NEW_RELIC_APPLICATION_LOGGING_FORWARDING_CONTEXT_DATA_ENABLED", default=False
 )
+_settings.application_logging.forwarding.context_data.include = _environ_as_bool(
```
hmstepanek, line 905: This `_environ_as_bool` doesn't seem right to me. These aren't booleans these are lists. Perhaps `_environ_as_set`?

newrelic/core/config.py
```diff
@@ -902,6 +902,12 @@ def default_otlp_host(host):
...
+    "NEW_RELIC_APPLICATION_LOGGING_FORWARDING_CONTEXT_DATA_INCLUDE", default=""
```
hmstepanek, line 906: We don't have env vars for any of the other attribute include/excludes. I wonder if we should include env vars at all for these?

tests/agent_features/test_log_events.py
```diff
@@ -102,10 +118,10 @@ def exercise_record_log_event():
         "message": "serialized_attributes",
         "context.str_attr": "Value",
         "context.bytes_attr": b"value",
-        "context.int_attr": "1",
-        "context.dict_attr": '{"key":"value"}',
+        "context.int_attr": 1,
+        "context.dict_attr": "{'key': 'value'}",
         "context.non_serializable_attr": "<NonSerializableObject object>",
-        "context.non_printable_attr": "<unprintable NonPrintableObject object>",
+        "context.attr_value_too_long": "*" * 255,
```
hmstepanek, line 124: Looks like this line replaced the previous instead of adding a new line.

tests/agent_features/test_log_events.py
```diff
@@ -219,3 +250,86 @@ def test():
...
+    @background_task()
+    def test():
+        if attr.startswith("context."):
+            record_log_event("A", attributes={attr.lstrip("context."): 1})
+        else:
+            record_log_event("A")
+
+    test()
+
+
```
hmstepanek, line 310: Do you think we need to test the non-datatype-specific `attributes.include` and `attributes.exclude` here at all?

tests/testing_support/validators/validate_log_events.py
```diff
@@ -79,6 +79,11 @@ def _check_log_attributes(expected, forgone_attrs, captured, mismatches):
                     mismatches.append("key: %s, value:<%s><%s>" % (key, value, captured_value))
                     return False
 
+        for key in required_attrs:
+            if not hasattr(captured, key) and key not in captured.attributes:
```
hmstepanek, line 83: This looks like it's checking both the attributes and directly on captured for the key. Why not do the same with the forgone_attrs check below for consistency?

hmstepanek (changes requested):
A couple small things but otherwise looks good.

newrelic/core/attribute.py
TimPansino: This matches the implementation in resolve_user_attributes which is used by spans and trace segments, unsure why the behavior is different between these. I'm not super tied to one or the other, but this matches most closely to these other resolve functions from this file.

newrelic/core/config.py
TimPansino: Fixed

newrelic/core/config.py
TimPansino: We could, but I'm not inclined to make that change with this PR.

tests/agent_features/test_log_events.py
TimPansino: Yeah, the `context.non_printable_attr` moved to a lower section for forgone attrs. These shouldn't be included anymore to match the implementation for existing attribute systems where a thrown exception drops the attribute (or all attributes).

tests/agent_features/test_log_events.py
TimPansino: Probably not, they're pretty well tested in their own section on the attribute filter rules in the cross agent tests. This implementation uses that system so we shouldn't need to reproduce those tests as well.

We could be a lot less thorough here if we wanted and rely on the precedence definitions in the cross agent tests. I could simplify this to just a couple runs, but it's not really hurting anything by staying as these are quite fast.

tests/testing_support/validators/validate_log_events.py
TimPansino: That's what the elif block did, but I moved some things around to make it maybe more readable.

newrelic/core/attribute.py
hmstepanek: span attributes actually behave the same way. The attributes actually appear to be processed twice which is different than logging user attributes and seems to be why the None is slipping past that check on spans and transactions. They are processed originally [here](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/core/node_mixin.py#L23) which converts the None to the string "None" and then again the resolve_user_attributes function. I'm not actually sure why there's a None check in there at all, since it doesn't seem like that would be reachable in the code. Either way, if you just take out the None check in this new function I think the behavior will be consistent across all the data types. I don't really see the point in processing them twice-I would think this would actually be a bad thing for performance. It seems like a mistake to me, but this code is complex so I could be missing something.

hmstepanek: Sorry got a little confused there when I was explaining that ^. It looks like the original function doesn't process the attributes again so they are only processed once but the point is that they are processed before the resolve_ function is called so the None gets converted to a "None" string.

TimPansino: That feels like it might be a bug but there's nothing in the spec either way that I can see so let's stick with sending it up then. I changed the PR.

hmstepanek: I'm inclined to agree with you on that but for now, I think it should be consistently wrong rather than a mix of the two depending on the datatype.

newrelic/core/config.py
hmstepanek: I think you may have miss-understood me here. I was suggesting that we could delete these env vars and not provide them since we don't provide them for any of the other attribute settings but if you want to keep these that's fine too.

hmstepanek (approved):
Looks good!

PR #1014, opened 2023-12-22
Add Dictionary Log Message Support

TimPansino:
# Overview

* To enable structured logging (JSON logging), accept dict values for `message` argument of `record_log_event`.
  * Unpack and add all message attributes to the event with the prefix `message.`
  * Retrieve any `message` field in the dictionary and treat as the log event's message for truncation purposes.
* Instrumentation changes for logging frameworks to follow in separate PRs.


newrelic/api/transaction.py
```diff
@@ -1540,17 +1540,49 @@ def record_log_event(self, message, level=None, timestamp=None, attributes=None,
...
+                if context_attributes:
+                    collected_attributes.update(context_attributes)
+
+            if message_attributes:
+                message_attributes = resolve_logging_context_attributes(message_attributes, settings.attribute_filter, "message.")
+                if message_attributes:
+                    collected_attributes.update(message_attributes)
+
+            # Exit early if no message or attributes found after filtering
+            if not collected_attributes and no_message:
```
hmstepanek, line 1580:
```suggestion
            if no_message and not collected_attributes:
```

newrelic/api/transaction.py
```diff
@@ -1540,17 +1540,49 @@ def record_log_event(self, message, level=None, timestamp=None, attributes=None,
...
+                _logger.debug("record_log_event called where message was missing, and no attributes found. No log event will be sent.")
+                return
 
-        collected_attributes = get_linking_metadata()
-        if attributes and (settings and settings.application_logging.forwarding.context_data.enabled):
-            context_attributes = resolve_logging_context_attributes(attributes, settings.attribute_filter, "context.")
-            if context_attributes:
-                collected_attributes.update(context_attributes)
+        # Finally, add in linking attributes after checking that there is a valid message or at least 1 attribute
```
hmstepanek, line 1584: What do you think about moving the above conditional:
```
            if no_message and not collected_attributes:
                _logger.debug("record_log_event called where message was missing, and no attributes found. No log event will be sent.")
                return
```
down here outside the `context_data.enabled` conditional and removing the conditional above that checks the no_message and attributes before they are filtered?

I could see a good reason for keeping these conditionals separate if the first one improved performance by exiting early before the filtering but I don't see that it results in a worthwhile performance improvement. If the attributes are not present, it doesn't try to filter them at all so I don't see the point in having two of these early exit checks. It seems like it'd be simpler to have just the one at the end.

newrelic/core/stats_engine.py
```diff
@@ -1233,17 +1233,49 @@ def record_log_event(self, message, level=None, timestamp=None, attributes=None,
         timestamp = timestamp if timestamp is not None else time.time()
         level = str(level) if level is not None else "UNKNOWN"
 
-        if not message or message.isspace():
-            _logger.debug("record_log_event called where message was missing. No log event will be sent.")
+        # Unpack message and attributes from dict inputs
```
hmstepanek, line 1236: I'm assuming this implementation is the same as the one above and the same feedback from the previous applies here as well.

hmstepanek (changes requested):

newrelic/api/transaction.py
```diff
@@ -1539,18 +1539,48 @@ def record_log_event(self, message, level=None, timestamp=None, attributes=None,
...
+                    collected_attributes.update(context_attributes)
+
+            if message_attributes:
+                message_attributes = resolve_logging_context_attributes(message_attributes, settings.attribute_filter, "message.")
+                if message_attributes:
+                    collected_attributes.update(message_attributes)
+
+            # Exit early if no message or attributes found after filtering
+            if (not message or message.isspace()) and not context_attributes and not message_attributes:
+                _logger.debug("record_log_event called where message was not found, and no attributes were found. No log event will be sent.")
```
hmstepanek:
```suggestion
                _logger.debug("record_log_event called where no message and no attributes were found. No log event will be sent.")
```

newrelic/api/transaction.py
```diff
@@ -1539,18 +1538,56 @@ def record_log_event(self, message, level=None, timestamp=None, attributes=None,
...
+                message_attributes = resolve_logging_context_attributes(
+                    message_attributes, settings.attribute_filter, "message."
+                )
+                if message_attributes:
+                    collected_attributes.update(message_attributes)
+
+            # Exit early if no message or attributes found after filtering
+            if (not message or message.isspace()) and not context_attributes and not message_attributes:
+                _logger.debug(
+                    "record_log_event called where message was not found, and no attributes were found. No log event will be sent."
```
hmstepanek, line 1585:
```suggestion
                    "record_log_event called where no message and no attributes were found. No log event will be sent."
```

newrelic/core/stats_engine.py
```diff
@@ -1232,18 +1232,56 @@ def record_log_event(self, message, level=None, timestamp=None, attributes=None,
...
+                message_attributes = resolve_logging_context_attributes(
+                    message_attributes, settings.attribute_filter, "message."
+                )
+                if message_attributes:
+                    collected_attributes.update(message_attributes)
+
+            # Exit early if no message or attributes found after filtering
+            if (not message or message.isspace()) and not context_attributes and not message_attributes:
+                _logger.debug(
+                    "record_log_event called where message was not found, and no attributes were found. No log event will be sent."
```
hmstepanek, line 1279:
```suggestion
                    "record_log_event called where no message and no attributes were found. No log event will be sent."
```

hmstepanek (approved):
This looks good now. I had one minor suggestion/fixup to the warning message but I'm just gonna pre-approve this assuming you'll make that change.

PR #1015, opened 2023-12-22
Logging Attribute Instrumentation

TimPansino:
# Overview

* Update logging instrumentation to support dict messages.
* Update context attributes to actually send default context attributes, in addition to extras.

# Linked PRs

Depends on #1014


tests/logger_logging/test_local_decorating.py
```diff
@@ -35,10 +35,10 @@ def set_trace_ids():
         trace.guid = "abcdefgh"
 
 
-def exercise_logging(logger):
+def exercise_logging(logger, message="C"):
```
hmstepanek, line 38: It would really be great if this was a fixture but since it's pre-existing it's not a must have.

tests/logger_logging/test_attributes.py
```diff
@@ -0,0 +1,90 @@
...
+@validate_log_event_count(1)
+@background_task()
+def test_logging_stack_info_context_attributes(logger):
+    logger.error("stack_info", stack_info=True)
+
+
+@validate_log_events([{"message": "A", "message.attr": 1}])
+@validate_log_event_count(1)
+@background_task()
+def test_logging_message_attributes(logger):
```
hmstepanek, line 82:
```suggestion
def test_logging_dict_message_and_attributes(logger):
```

tests/logger_logging/test_attributes.py
```diff
@@ -0,0 +1,90 @@
...
+    logger.error({"message": "A", "attr": 1})
+
+
+@validate_log_events([{"message.attr": 1}])
+@validate_log_event_count(1)
+@background_task()
+def test_logging_attributes_only(logger):
```
hmstepanek, line 89:
```suggestion
def test_logging_dict_attributes_only(logger):
```

tests/logger_logging/test_attribute_forwarding.py
```diff
@@ -1,65 +0,0 @@
...
-    test()
-
-
-@reset_core_stats_engine()
-@override_application_settings(
-    {
-        "application_logging.forwarding.context_data.enabled": True,
-    }
-)
-def test_attributes_outside_transaction(logger):
```
hmstepanek, line 59: It looks like this was testing logging outside of a transaction. If we have a registered application it should still record a log event even outside of a transaction. Do we have any tests for calling logging outside a transaction?

hmstepanek (changes requested):
In general this looks really good. I had a couple small requests and a question about the tests for outside a transaction.

tests/logger_logging/test_local_decorating.py
TimPansino: Yeah not inclined to change it with this PR, maybe in the future

tests/logger_logging/test_attribute_forwarding.py
TimPansino: I don't think it's necessary, as we're not trying to exercise `record_log_event` really. We're trying to exercise the instrumentation path that harvests all the attributes and supplies them to the API.

`record_log_event` is tested under `agent_features` both inside and outside of transactions, so the API doesn't need to be exercised again.

The older tests didn't have record_log_event tests separately, so they're kind of a mix. I didn't remove all the existing ones since it seemed a bit pointless.

hmstepanek (approved):

PR #1017, opened 2023-12-30
Remove case sensitive check in ASGIBrowserMiddleware check.

### LINKED ISSUE #1011: ASGIBrowserMiddleware does not apply when the headers are not lower cased.
**Description**
When attempting to run a server (Django/Daphne in my case), when the `content-type` header is not in lower case the `should_insert_html` function returns false.

Reviewing the Commit History this has been in place since ASGI support was introduced.

Refer to: https://asgi.readthedocs.io/en/latest/specs/www.html#http

> Header names should be lowercased, but it is not required; servers should preserve header case on a best-effort basis.

**Expected Behavior**
Headers should be case insensitve

umaannamalai:
Fixes #1011


TimPansino (approved):
🚀

PR #1019, opened 2024-01-02
Error trace linking

lrafeei:
# Overview
This PR adds `guid` as an intrinsic field instead of just having a guid in the error trace when Distributed Tracing is enabled.  This links the Transaction Error and the Error Trace even if distributed tracing is off, allowing for the user to see more attributes that are attached to the Error Trace in the Errors Inbox UI.


newrelic/core/stats_engine.py
```diff
@@ -882,9 +882,15 @@ def notice_error(self, error=None, attributes=None, expected=None, ignore=None,
...
 
+        # Instead of creating a brand new transaction, we
+        # simply attach the entity_guid to the intrinsics
+        # as well as the TracedError
+        no_transaction_guid = getattr(settings, "entity_guid", None)
+
         # pass expected attribute in to ensure we capture overrides
         attributes["intrinsics"] = {
             "error.expected": is_expected,
+            "guid": no_transaction_guid,
```
TimPansino, line 893: I think we should separate this part of the change until we can confirm 100% what the behavior will be like, taking this to the spec reviewers meeting.

newrelic/core/stats_engine.py
```diff
@@ -894,7 +900,12 @@ def notice_error(self, error=None, attributes=None, expected=None, ignore=None,
                 attributes["agentAttributes"][attr.name] = attr.value
 
         error_details = TracedError(
-            start_time=time.time(), path="Exception", message=message, type=fullname, parameters=attributes
+            start_time=time.time(),
+            path="Exception",
+            message=message,
+            type=fullname,
+            parameters=attributes,
+            guid=no_transaction_guid,
```
TimPansino, line 908: This line should be `None` until we decide on this behavior

tests/agent_features/test_distributed_tracing.py
```diff
@@ -43,7 +43,7 @@
 from newrelic.api.web_transaction import WSGIWebTransaction
 from newrelic.api.wsgi_application import wsgi_application
 
-distributed_trace_intrinsics = ["guid", "traceId", "priority", "sampled"]
+distributed_trace_intrinsics = ["traceId", "priority", "sampled"]
```
TimPansino, line 46: I think this file should probably be reverted

tests/agent_features/test_asgi_distributed_tracing.py
```diff
@@ -12,80 +12,91 @@
...
-inbound_payload_intrinsics = ['parent.type', 'parent.app', 'parent.account',
-        'parent.transportType', 'parent.transportDuration']
+distributed_trace_intrinsics = ["traceId", "priority", "sampled"]
+inbound_payload_intrinsics = [
+    "parent.type",
+    "parent.app",
+    "parent.account",
+    "parent.transportType",
+    "parent.transportDuration",
+]
```
TimPansino, line 37: I think this file should probably be reverted

tests/agent_features/test_attributes_in_action.py
```diff
@@ -83,7 +83,6 @@
     "parent.transportType",
     "parent.transportDuration",
     "parentId",
-    "guid",
```
TimPansino, line 86: Let's put this line back as well

TimPansino (changes requested):
Lots of things to move around, let me know when this is separated and in the meantime I'll have to bring this up with the spec reviewers.

tests/agent_features/test_asgi_distributed_tracing.py
lrafeei: In this case, the distributed tracing intrinsics list goes into an expected absenses attributes list.  Since guid is now DT agnostic, it is expected that this make it to the intrinsics list.

tests/agent_features/test_attributes_in_action.py
lrafeei: In this case, the distributed tracing intrinsics list goes into an expected absenses attributes list. Since guid is now DT agnostic, it is expected that this make it to the intrinsics list.

TimPansino (approved):
Waiting for tests on this one to pass but otherwise it looks good to go

PR #1021, opened 2024-01-03
Temporarily pin hypercorn version in tests

lrafeei:
# Overview
Temporarily pin hypercorn testing suite from latest to <0.16

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -174,6 +174,7 @@ deps =
     adapter_gunicorn-aiohttp03-py312: aiohttp==3.9.0rc0
     adapter_gunicorn-gunicorn19: gunicorn<20
     adapter_gunicorn-gunicornlatest: gunicorn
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -174,7 +174,7 @@ deps =
     adapter_gunicorn-aiohttp03-py312: aiohttp==3.9.0rc0
     adapter_gunicorn-gunicorn19: gunicorn<20
     adapter_gunicorn-gunicornlatest: gunicorn
-    adapter_hypercorn-hypercornlatest: hypercorn
+    adapter_hypercorn-hypercornlatest: hypercorn<0.16
```
umaannamalai, line 177: Maybe we could add a comment here so we remember to come back and address the failures on the latest version?

lrafeei: Good call.  Done!

umaannamalai (approved):

PR #1022, opened 2024-01-03
Add 3.12 wheels to build

hmstepanek:
# Overview
This PR fixes up some issues with supporting 3.12.
* Add 3.12 into the list of wheel build envs.

### DIFF
--- a/.github/workflows/deploy-python.yml
+++ b/.github/workflows/deploy-python.yml
@@ -57,7 +57,7 @@ jobs:
         uses: pypa/cibuildwheel@v2.11.1
         env:
           CIBW_PLATFORM: linux
-          CIBW_BUILD: cp37-manylinux* cp38-manylinux* cp39-manylinux* cp310-manylinux* cp311-manylinux*
+          CIBW_BUILD: cp37-manylinux* cp38-manylinux* cp39-manylinux* cp310-manylinux* cp311-manylinux* cp312-manylinux*
           CIBW_ARCHS: x86_64 aarch64
           CIBW_ENVIRONMENT: "LD_LIBRARY_PATH=/opt/rh/devtoolset-8/root/usr/lib64:/opt/rh/devtoolset-8/root/usr/lib:/opt/rh/devtoolset-8/root/usr/lib64/dyninst:/opt/rh/devtoolset-8/root/usr/lib/dyninst:/usr/local/lib64:/usr/local/lib"
 


.github/containers/Dockerfile
```diff
@@ -96,12 +96,12 @@ RUN echo 'eval "$(pyenv init -)"' >>${HOME}/.bashrc && \
     pyenv update
 
 # Install Python
-ARG PYTHON_VERSIONS="3.11 3.10 3.9 3.8 3.7 3.12 2.7 pypy2.7-7.3.12 pypy3.8-7.3.11"
+ARG PYTHON_VERSIONS="3.12 3.11 3.10 3.9 3.8 3.7 3.12 2.7 pypy2.7-7.3.12 pypy3.8-7.3.11"
```
TimPansino, line 99:
```suggestion
ARG PYTHON_VERSIONS="3.11 3.10 3.9 3.8 3.7 3.12 2.7 pypy2.7-7.3.12 pypy3.8-7.3.11"
```

3.12 is already in the versions list, it's at the end because we usually use a more stable version as our default for running tools such as tox.

TimPansino (changes requested):
One change and this looks good

newrelic/packages/isort/stdlibs/py312.py
```diff
@@ -0,0 +1,220 @@
+"""
+File contains the standard library of Python 3.12.
+
+DO NOT EDIT. If the standard library changes, a new list should be created
+using the mkstdlibs.py script.
+"""
+
```
TimPansino, line 7: Is this file necessary at all? Based on the below code I don't think we ever reference it.

```python
    elif python_version >= (3, 10):
        stdlibs = sys.stdlib_module_names
```

TimPansino: It really doesn't require python 3.10 or 3.11 either, but those were probably just copied in since they were there. I'm unsure if there's any reason to update these.

.github/containers/Dockerfile
hmstepanek: oh no! I see it now oops.

newrelic/packages/isort/stdlibs/py312.py
hmstepanek: Ah true. It looks like there were only changes to 3.11 and 3.12 here. Given that both those aren't used, I guess we can just skip upgrading isort for now.

TimPansino (approved):

.github/workflows/deploy-python.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         uses: pypa/cibuildwheel@v2.11.1
```
TimPansino, line 57:
```suggestion
        uses: pypa/cibuildwheel@v2.16.2
```

TimPansino (changes requested):
This needs to be updated or it won't have 3.12 support and the CI will crash.

TimPansino:
Reposting this comment

> Closing this since this PR does it and more: https://github.com/newrelic/newrelic-python-agent/pull/1024.

PR #1024, opened 2024-01-04
Parallel Wheel Builds

TimPansino:
# Overview

* Add parallelized build strategy for wheels to reduce build times.
* Add Python 3.12 wheels.
* Add mullinux wheels in addition to manylinux wheels.
* Add validation tests (agent_unittests) to end of wheel build process.
* Update actions used in build and deploy to newer versions.

### DIFF
--- a/tests/agent_unittests/test_package_version_utils.py
+++ b/tests/agent_unittests/test_package_version_utils.py
@@ -16,7 +16,6 @@ import sys
 import warnings
 
 import pytest
-import six
 from testing_support.validators.validate_function_called import validate_function_called
 
 from newrelic.common.package_version_utils import (
@@ -26,6 +25,7 @@ from newrelic.common.package_version_utils import (
     get_package_version,
     get_package_version_tuple,
 )
+from newrelic.packages import six
 
 # Notes:
 # importlib.metadata was a provisional addition to the std library in PY38 and PY39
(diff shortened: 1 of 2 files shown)


tests/agent_unittests/test_package_version_utils.py
```diff
@@ -26,6 +25,7 @@
     get_package_version,
     get_package_version_tuple,
 )
+from newrelic.packages import six
```
hmstepanek, line 28: This seems random. There are other places where we use six directly in the tests too. Not that it's a big deal, but maybe don't do this in this PR as it's not really related to PR's purpose.

hmstepanek (approved):
This looks good. I had one minor request but it's not a big deal. I'm just gonna pre-approve this regardless of that request.

hmstepanek:
Closing this since this PR does it and more: https://github.com/newrelic/newrelic-python-agent/pull/1024.

TimPansino:
@hmstepanek You closed the wrong one 😆

tests/agent_unittests/test_package_version_utils.py
TimPansino: It actually is related, the `agent_unittests` are running on `cibuildwheel` in their wheel validation step now. Without `tox` setting up the virtual environment, `six` doesn't end up accessible, unsure if that's because it's normally installed or if it's available due to our `PYTHONPATH` shenanigans.

PR #1025, opened 2024-01-08
Loguru Attribute Instrumentation

TimPansino:
# Overview

* Add context attributes instrumentation for `loguru`.


tests/logger_loguru/test_stack_inspection.py
```diff
@@ -1,56 +0,0 @@
-# Copyright 2010 New Relic, Inc.
```
hmstepanek, line 1: Why did this test get removed?

hmstepanek (commented):
I had one question about the regression test getting removed but otherwise this looks good.

newrelic/hooks/logger_loguru.py
```diff
@@ -18,19 +18,24 @@
...
+ALLOWED_LOGURU_OPTIONS_LENGTHS = frozenset((8, 9))
 
-def loguru_version():
-    from loguru import __version__
 
-    return tuple(int(x) for x in __version__.split("."))
+def _filter_record_attributes(record):
+    attrs = {k: v for k, v in record.items() if k not in LOGURU_FILTERED_RECORD_ATTRS}
+    extra_attrs = dict(record.get("extra", {}))
+    attrs.update({"extra.%s" % k: v for k, v in extra_attrs.items()})
```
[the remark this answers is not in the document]
TimPansino: There weren't attributes in any published releases yet. This is changed from an older unreleased PR that was included on the develop branch.

tests/logger_loguru/test_stack_inspection.py
TimPansino: Those can be validated as attributes now instead of the clunky message string check we had before. They're included in `test_loguru_default_context_attributes`. Here's the 2 of them that are checked for value that are equivalent to what was tested before.

```python
            "context.file": "(name='%s', path='%s')" % ("test_attributes.py", str(__file__)),
            "context.function": "test_loguru_default_context_attributes",
```

hmstepanek (approved):

PR #1026, opened 2024-01-08
StructLog Attribute Instrumentation

TimPansino:
# Overview

* Add context attributes instrumentation for `structlog`.


newrelic/hooks/logger_structlog.py
```diff
@@ -50,16 +47,28 @@ def wrap__process_event(wrapped, instance, args, kwargs):
...
+        else:
+            # Unclear how to proceed, ignore log. Avoid logging an error message or we may incur an infinite loop.
+            return event
+
+        if settings.application_logging.local_decorating.enabled:
+            message = add_nr_linking_metadata(message)
+            if isinstance(event, (str, bytes, bytearray)):
+                event = message
+            elif isinstance(event, dict) and "event" in event:
+                # TODO CHECK ON THIS
```
umaannamalai, line 66: Was this TODO checked?

newrelic/hooks/logger_structlog.py
```diff
@@ -50,16 +47,28 @@ def wrap__process_event(wrapped, instance, args, kwargs):
         settings = global_settings()
 
     # Return early if application logging not enabled
-    if settings and settings.application_logging and settings.application_logging.enabled:
-        if settings.application_logging.local_decorating and settings.application_logging.local_decorating.enabled:
-            event = add_nr_linking_metadata(event)
+    if settings and settings.application_logging.enabled:
```
umaannamalai, line 50: We probably can remove this `if settings` check since settings should be supplied somehow in the if check above. For an extra layer of safety in case `global_settings` is empty for some reason, we could also just do a `if settings` check before all this and return out of the wrapper if we don't get anything back.

newrelic/hooks/logger_structlog.py
```diff
@@ -69,15 +78,34 @@ def wrap__process_event(wrapped, instance, args, kwargs):
...
+def wrap__process_event(wrapped, instance, args, kwargs):
+    transaction = current_transaction()
+
+    if transaction:
+        settings = transaction.settings
+    else:
+        settings = global_settings()
+
+    # Return early if application logging not enabled
+    if settings and settings.application_logging and settings.application_logging.enabled:
```
umaannamalai, line 100: We can probably remove the if `settings.application_logging` check.

newrelic/hooks/logger_structlog.py
TimPansino: Nope, that definitely needs checked

newrelic/hooks/logger_structlog.py
TimPansino: We want to guard against transaction settings being set to None, which has happened in the past.

newrelic/hooks/logger_structlog.py
TimPansino: I checked what this was about, it should be being tested.

umaannamalai (approved):

PR #1028, opened 2024-01-08
Add instrumentation for sync LangChain tools.

umaannamalai:
This PR adds instrumentation for synchronous LangChain tools.

### DIFF
--- a/newrelic/hooks/mlmodel_langchain.py
+++ b/newrelic/hooks/mlmodel_langchain.py
@@ -20,6 +20,8 @@ from newrelic.api.transaction import current_transaction
 from newrelic.common.object_names import callable_name
 from newrelic.common.object_wrapper import wrap_function_wrapper
 from newrelic.common.package_version_utils import get_package_version
+from newrelic.common.signature import bind_args
+from newrelic.core.config import global_settings
 
 LANGCHAIN_VERSION = get_package_version("langchain")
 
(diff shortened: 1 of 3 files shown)


newrelic/hooks/mlmodel_langchain.py
```diff
@@ -255,10 +257,138 @@ def wrap_similarity_search(wrapped, instance, args, kwargs):
...
+    if not transaction:
+        return wrapped(*args, **kwargs)
+
+    # Framework metric also used for entity tagging in the UI
+    transaction.add_ml_model_info("Langchain", LANGCHAIN_VERSION)
+
+    run_args = bind_args(wrapped, args, kwargs)
+    tool_input = run_args.get("tool_input", "")
+    tags = run_args.get("tags", "")
+    metadata = run_args.get("metadata", "")
```
hmstepanek, line 271:
```suggestion
    metadata = run_args.get("metadata") or {}
```

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -255,10 +257,138 @@ def wrap_similarity_search(wrapped, instance, args, kwargs):
...
+                }
+            )
+
+            run_manager_info = getattr(transaction, "_nr_run_manager_info", {})
+            if hasattr(transaction, "_nr_run_manager_info"):
+                del transaction._nr_run_manager_info
+            run_id = run_manager_info.get("run_id", "")
+
+            # Make sure the builtin attributes take precedence over metadata attributes.
+            error_tool_event_dict = {"metadata.%s" % key: value for key, value in metadata.items()} if metadata else {}
```
hmstepanek, line 305:
```suggestion
            error_tool_event_dict = {"metadata.%s" % key: value for key, value in metadata.items()}
```

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -255,10 +257,138 @@ def wrap_similarity_search(wrapped, instance, args, kwargs):
...
+    transaction = current_transaction()
+    if not transaction:
+        return wrapped(*args, **kwargs)
+
+    # Framework metric also used for entity tagging in the UI
+    transaction.add_ml_model_info("Langchain", LANGCHAIN_VERSION)
+
+    run_args = bind_args(wrapped, args, kwargs)
+    tool_input = run_args.get("tool_input", "")
+    tags = run_args.get("tags", "")
```
hmstepanek, line 270:
```suggestion
    tags = run_args.get("tags")
```

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -255,10 +257,138 @@ def wrap_similarity_search(wrapped, instance, args, kwargs):
...
+    return return_val
+
+
+def wrap_on_tool_start(wrapped, instance, args, kwargs):
+    run_manager = wrapped(*args, **kwargs)
+    transaction = current_transaction()
+    if not transaction:
+        return run_manager
+    # Only capture the first run_id.
+    if not hasattr(transaction, "_nr_run_manager_info"):
```
hmstepanek, line 370: I think this is the exact same name as I used in my PR for chains so they will conflict with each other which is not good. Whichever tools or chains finishes first, it will eat this value and it won't be available after and same goes for whichever creates it first. Maybe name this `_nr_run_manager_tools_info`?

newrelic/hooks/mlmodel_langchain.py
hmstepanek, line 270: If we just grab the tags and metadata from the input args we could be missing tags. See https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/chains/base.py#L139. Note it passes self.tags and tags and self.metadata and metadata into the callback manager and those ultimately get combined together and passed to each callback.

hmstepanek (changes requested):
I haven't taken an in-depth look at the tests yet but I wanted to get this review back to you anyway. I like the way you are pulling the metadata and tags from the input args but we also need to pull them from the instance as well. In the chain PR I pulled them from the call manager callback but I could instead re-create them myself which might be a better option. That being said we also need tests for metadata and tags being added on the instance as well as provided as input args to the function call.

newrelic/hooks/mlmodel_langchain.py
hmstepanek: Just and FYI here: I ended up changing the name in the chains PR to just `_nr_chain_run_id` so we shouldn't have a conflict anymore.

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -256,6 +258,132 @@ def wrap_similarity_search(wrapped, instance, args, kwargs):
...
+        try:
+            return_val = wrapped(**run_args)
+        except Exception as exc:
+            ft.notice_error(
+                attributes={
+                    "tool_id": tool_id,
+                }
+            )
+
+            run_manager_info = getattr(transaction, "_nr_run_manager_tools_info", {})
```
hmstepanek, line 307: Might as well just grab this whole thing in one line.
```suggestion
            run_id = getattr(transaction, "_nr_run_manager_tools_info", {}).get("run_id", "")
```

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -256,6 +258,132 @@ def wrap_similarity_search(wrapped, instance, args, kwargs):
...
+            if hasattr(transaction, "_nr_run_manager_tools_info"):
+                del transaction._nr_run_manager_tools_info
+            run_id = run_manager_info.get("run_id", "")
```
hmstepanek, line 310:
```suggestion
```

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -256,6 +258,132 @@ def wrap_similarity_search(wrapped, instance, args, kwargs):
...
+            transaction.record_custom_event("LlmTool", error_tool_event_dict)
+
+            raise
+
+    if not return_val:
+        return return_val
+
+    response = return_val
+
+    run_manager_info = getattr(transaction, "_nr_run_manager_tools_info", {})
```
hmstepanek, line 343: Same here
```suggestion
    run_id = getattr(transaction, "_nr_run_manager_tools_info", {}).get("run_id", "")
```

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -256,6 +258,132 @@ def wrap_similarity_search(wrapped, instance, args, kwargs):
...
+    if hasattr(transaction, "_nr_run_manager_tools_info"):
+        del transaction._nr_run_manager_tools_info
+    run_id = run_manager_info.get("run_id", "")
```
hmstepanek, line 346:
```suggestion
```

tests/mlmodel_langchain/test_tool.py
```diff
@@ -0,0 +1,242 @@
...
+    validate_transaction_metrics,
+)
+
+from newrelic.api.background_task import background_task
+from newrelic.common.object_names import callable_name
+
+disabled_custom_insights_settings = {"custom_insights_events.enabled": False}
+
+
+@tool
```
hmstepanek, line 39: Let's make these tools into fixtures:
```
@pytest.fixture
def single_arg_tool():
    @tool
    def single_arg_tool_(query: str):
        """A test tool that returns query string"""
        return query
    return single_arg_tool_

@pytest.fixture
def multi_arg_tool():
    @tool
    def multi_arg_tool_(first_num: int, second_num: int):
        """A test tool that adds two integers together"""
        return first_num + second_num
    return multi_arg_tool_
```

tests/mlmodel_langchain/test_tool.py
```diff
@@ -0,0 +1,242 @@
...
+    rollup_metrics=[("Llm/tool/Langchain/run", 1)],
+    custom_metrics=[
+        ("Python/ML/Langchain/%s" % langchain.__version__, 1),
+    ],
+    background_task=True,
+)
+@background_task()
+def test_langchain_multi_arg_tool(set_trace_info):
+    set_trace_info()
+    multi_arg_tool.run(
```
hmstepanek, line 132: I think we could probably just combine this all into one test for the metadata and tags and then delete `test_langchain_tags_and_metadata_on_instance`:
```suggestion
    multi_arg_tool.metadata = {"test_run": True}
    multi_arg_tool.tags=["test_tags"]
    multi_arg_tool.run(
```

tests/mlmodel_langchain/test_tool.py
```diff
@@ -0,0 +1,242 @@
...
+        {"first_num": 53, "second_num": 28},
+        tags=["test_tags", "python"],
```
hmstepanek, line 134:
```suggestion
        tags=["python"],
```

tests/mlmodel_langchain/test_tool.py
```diff
@@ -0,0 +1,242 @@
...
+        metadata={"test_run": True, "test": "langchain"},
```
hmstepanek, line 135:
```suggestion
        metadata={"test": "langchain"},
```

hmstepanek (changes requested):
Looks really good just a couple minor suggestions.

hmstepanek (approved):
Looks good! There's quite a few tests failing but those should be fixed in the chains PR.

PR #1030, opened 2024-01-09
Add bedrock feedback into preview

lrafeei:
# Overview
Merge changes from #958 into the preview branch


tests/external_botocore/_mock_external_bedrock_server.py
```diff
@@ -3473,7 +3473,6 @@ def __init__(self, handler=simple_get, port=None, *args, **kwargs):
 if __name__ == "__main__":
     # Use this to sort dict for easier future incremental updates
     print("RESPONSES = %s" % dict(sorted(RESPONSES.items(), key=lambda i: (i[1][1], i[0]))))
```
hmstepanek, line 3475: This looks like a no-op. Can you just remove this file from the list of changes?
```
git co develop-ai-limited-preview tests/external_botocore/_mock_external_bedrock_server.py
git add tests/external_botocore/_mock_external_bedrock_server.py
git commit -m "Remove untouched file from change list"
```

tests/external_botocore/_test_bedrock_embeddings.py
```diff
@@ -68,7 +68,7 @@
                 "request_id": "",
                 "vendor": "bedrock",
                 "ingest_source": "Python",
-                "error": True
```
hmstepanek, line 71: Again, this looks like a no-op. Just remove this file from the list of changes.

tests/external_botocore/test_bedrock_embeddings.py
```diff
@@ -19,8 +19,8 @@
 import pytest
 from _test_bedrock_embeddings import (
     embedding_expected_client_errors,
-    embedding_expected_events,
     embedding_expected_error_events,
```
hmstepanek, line 22: Same thing here-basically a no-op. Just drop it from the list of changes.

hmstepanek (changes requested):
Thank you for doing this-can't believe we missed cherry-picking this over! It looks like there are several files in here that are basically no-ops. I'm just requesting that they be dropped as the changes are purely minor re-formats and aren't required. (Likely just merge conflict leftovers.)

tests/external_botocore/test_bedrock_embeddings.py
lrafeei: I figure I might keep this one only because this one is a linter formatting thing where it takes the imports and alphabetizes them.

hmstepanek (approved):

PR #1031, opened 2024-01-09
Obfuscate License Keys in Logs

TimPansino:
# Overview

* Add obfuscation to all instances where license keys are printed or logged.

# Testing

* Manual testing must be done to ensure logs don't contain the new relic license keys.
  * Run a test app with a valid license key, with debug and audit logs sent to files.
  * Use `grep -r . MYLICENSEKEY` to find any stray license keys.
  * Repeat the above with `newrelic-admin validate-config newrelic.ini agent.log`
  * Repeat the above with `newrelic-admin license-key newrelic.ini agent.log`


newrelic/common/agent_http.py
```diff
@@ -119,6 +123,11 @@ def log_request(cls, fp, method, url, params, payload, headers, body=None, compr
         if not fp:
             return
 
+        # Obfuscate license key from params
```
hmstepanek, line 126: The API key is part of the headers in the OTLP class and the NewRelicLogHandler (See line 557 in newrelic/core/agent_protocol.py). I believe those get passed into this function call as headers and get printed out to the file below on line 149 so we need to obfuscate that too.

hmstepanek (changes requested):
Thank you for making this change! It looks like we might still be missing a couple places in the code. Primarily the other thing to look out for is if the license key is added as a `api-key` header. I've noted one in a comment but here's another:
* newrelic/bootstrap/sitecustomize.py: 137

newrelic/common/agent_http.py
TimPansino: You're right, the headers elsewhere get filtered but not in the audit log. I added exclusions for those 2 cases and the `record_deploy` command which uses `X-Api-Key`.

TimPansino:
> Thank you for making this change! It looks like we might still be missing a couple places in the code. Primarily the other thing to look out for is if the license key is added as a `api-key` header. I've noted one in a comment but here's another:
> 
> * newrelic/bootstrap/sitecustomize.py: 137

```python
log_message("initialize_agent = %r", bool(license_key or config_file))
bool(license_key or config_file)
```

This is cast to a bool, it's not printing the license key out. It's just logging whether the agent will be initialized or not as True/False.

hmstepanek (approved):
Looks good assuming the tests pass now.

PR #1041, opened 2024-01-18
Support capturing run_id during multiple async calls

hmstepanek:
# Overview
Support capturing run_id during multiple async calls.


tests/mlmodel_langchain/test_chain.py
```diff
@@ -1841,6 +1845,94 @@ def test_async_langchain_chain_outside_transaction(
...
+    @validate_transaction_metrics(
+        name="test_chain:test_multiple_async_langchain_chain.<locals>._test",
+        custom_metrics=[
+            ("Python/ML/Langchain/%s" % langchain.__version__, 1),
+        ],
+        background_task=True,
+    )
+    @background_task()
+    def _test():
+        with patch("langchain_core.callbacks.manager.uuid", autospec=True) as mock_uuid:
```
umaannamalai, line 1910: I was trying to see if there was a way to do this without needing to introduce another test dependency through `mock` but this may be fine as is.

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -664,24 +672,34 @@ def create_chat_completion_message_event(
...
 
 def wrap_on_chain_start(wrapped, instance, args, kwargs):
-    run_manager = wrapped(*args, **kwargs)
     transaction = current_transaction()
     if not transaction:
-        return run_manager
-    # Only capture the first run_id.
-    if not hasattr(transaction, "_nr_chain_run_id"):
-        transaction._nr_chain_run_id = getattr(run_manager, "run_id", "")
+        return wrapped(*args, **kwargs)
```
umaannamalai, line 677: Is there a need to implement this logic for sync runs? If the `completion_id` storing/ checks are done in the async run wrapper, can we avoid changing the sync logic here?

hmstepanek: Yeah as far as I know monkeypatch could be used to override the function but I wouldn't be able to get the side_effect behavior that I have here-it would be a lot more awkward to implement.

tests/mlmodel_langchain/test_chain.py
hmstepanek: Yeah as far as I know monkeypatch could be used to override the function but I wouldn't be able to get the side_effect behavior that I have here-it would be a lot more awkward to implement.

tests/mlmodel_langchain/test_chain.py
```diff
@@ -1841,6 +1845,94 @@ def test_async_langchain_chain_outside_transaction(
...
+    prompt,
+    chat_openai_client,
+    create_function,
+    call_function,
+    call_function_args,
+    call_function_kwargs,
+    expected_events,
+    loop,
+):
+    call1 = copy.deepcopy(expected_events)
```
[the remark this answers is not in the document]
hmstepanek: I changed this to just a obj.copy.

newrelic/hooks/mlmodel_langchain.py
hmstepanek: I suppose we don't have to change the sync logic. It seemed better design wise to just do the same thing for both async and sync though so they maintain symmetry.

hmstepanek: The current implementation shares the same _get_run_manager_info function which we wouldn't be able to do if they had different behaviors.

umaannamalai (approved):

PR #1042, opened 2024-01-19
Add required files for Dependabot

lrafeei:
# Overview
This PR adds a yml file for Dependabot as well as adding a `requirements.txt` file within the `newrelic/packages/` directory to include the New Relic vendored frameworks and versions used.

Note: Vendored packages are those that exist within the `newrelic/packages/` directory.  They do not need to be installed separately.  To avoid a scenario where an application such as Poetry is used to install New Relic and all its vendored packages are re-installed, `requirements.txt` was added to the `newrelic/packages/` directory.  `requirements.txt` cannot exist in the root directory and `pyproject.toml` cannot be used to add these dependencies because applications such as Poetry will install these packages as well.


TimPansino:
> To keep CI files out of the root of the repo, can we move the dependabot folder to `.github/dependabot`?

Actually it might be better to add the `requirements.txt` file directly to `newrelic/packages/` that way it's easily seen when looking at the package code to update it.

TimPansino (approved):

TimPansino:
🚢

PR #1046, opened 2024-01-24
Add langchain async tools instrumentation

umaannamalai:
This PR adds instrumentation for async langchain tools


newrelic/hooks/mlmodel_langchain.py
```diff
@@ -371,7 +371,115 @@ def wrap_tool_sync_run(wrapped, instance, args, kwargs):
...
+        return await wrapped(*args, **kwargs)
+    # Framework metric also used for entity tagging in the UI
+    transaction.add_ml_model_info("Langchain", LANGCHAIN_VERSION)
+    tool_id = str(uuid.uuid4())
+
+    run_args = bind_args(wrapped, args, kwargs)
+
+    metadata = {}
+    metadata.update(run_args.get("metadata") or {})
+    metadata.update(getattr(instance, "metadata", None) or {})
```
hmstepanek, line 386: Probably shouldn't add the metadata from the instance into the run args metadata and pass that to wrapped and same with tags. Maybe move this update logic down into where we actually record the metadata so we aren't duplicating the instance settings into the argument settings?

hmstepanek (changes requested):
Just had one request about not passing the instance metadata and tags into the run argument metadata and tags but everything else looks good!

hmstepanek (approved):

PR #1047, opened 2024-01-24
Add dependabot support

hmstepanek:
Add dependabot support via requirements.txt.


hmstepanek:
Fixed in https://github.com/newrelic/newrelic-python-agent/pull/1042.

PR #1051, opened 2024-01-29
Update botocore tests.

umaannamalai:
This PR updates the `moto` decorators used in `botocore` tests and removes py27 dependencies listed in `tox.ini`. It also drops testing for Python 3.7 since `botocore` no longer supports this version.

### DIFF
--- a/tests/external_botocore/test_boto3_iam.py
+++ b/tests/external_botocore/test_boto3_iam.py
@@ -16,7 +16,7 @@ import sys
 import uuid
 
 import boto3
-import moto
+from moto import mock_aws
 from testing_support.fixtures import dt_enabled
 from testing_support.validators.validate_span_events import validate_span_events
 from testing_support.validators.validate_transaction_metrics import (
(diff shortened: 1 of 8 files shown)


TimPansino (approved):
LGTM

lrafeei:
Could we take this out as well since v1.3.1 was released in 2018 and 3.7 support was dropped?

```
# patch earlier versions of moto to support py37
if sys.version_info >= (3, 7) and MOTO_VERSION <= (1, 3, 1):
    import re
    moto.packages.responses.responses.re._pattern_type = re.Pattern
```

umaannamalai:
> Could we take this out as well since v1.3.1 was released in 2018 and 3.7 support was dropped?
> 
> ```
> # patch earlier versions of moto to support py37
> if sys.version_info >= (3, 7) and MOTO_VERSION <= (1, 3, 1):
>     import re
>     moto.packages.responses.responses.re._pattern_type = re.Pattern
> ```

Sure thing! I'll add that in a followup PR.

PR #1055, opened 2024-02-02
Update CherryPy pypy3 tests to use pypy310

lrafeei:
# Overview
This PR uses pypy310 instead of pypy38 in CherryPy


TimPansino:
Opened a similar PR, I'm personally inclined to keep everything on the same version of PyPy3 and just pin dependancies.

PR #1061, opened 2024-02-15
Openai stream v1 support

hmstepanek:
Add support for streaming in openai v1

### DIFF
--- a/newrelic/hooks/mlmodel_openai.py
+++ b/newrelic/hooks/mlmodel_openai.py
@@ -1275,23 +1275,17 @@ async def wrap_engine_api_resource_create_async(wrapped, instance, args, kwargs)
         return return_val
 
 
-async def wrap_stream_iter_events_async(wrapped, instance, args, kwargs):
+def wrap_stream_iter_events_async(wrapped, instance, args, kwargs):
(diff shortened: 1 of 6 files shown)


newrelic/hooks/mlmodel_openai.py
```diff
@@ -1245,6 +1275,25 @@ async def wrap_engine_api_resource_create_async(wrapped, instance, args, kwargs)
...
+
+    async_iterator = wrapped(*args, **kwargs)
+    try:
+        async for return_val in async_iterator:
+            record_stream_chunk(instance, return_val)
+            yield return_val
+    except Exception as exc:
+        record_error(instance, transaction, exc)
+        raise
+    record_events_on_stop_iteration(instance, transaction)
```
TimPansino, line 1294: I'm not positive but this might be catching the StopAsyncIteration exception as well. I think you'd need another except clause above this one to handle it.

TimPansino: Disregard that, `except Exception` does not include StopAsyncIteration so this is fine.

TimPansino (approved):

PR #1064, opened 2024-02-16
Drop attributes with a value of None

hmstepanek:
# Overview
This PR attempts to align the Python agent with the agent spec by dropping attributes with a value of None.

According to the spec:
> Agents **SHOULD NOT** report `null` attribute values. The behavior of `IS NULL` queries in insights makes it so that omitted keys behave the same as `null` keys. Since there is no difference between omitting the key and sending a `null`, we **SHOULD** reduce the payload size by omitting `null` values from the payload entirely.

> Since there should not be a difference in behavior for customers recording `null` attributes versus customers omitting an attribute, agents **SHOULD** add debug logging when a null value is recorded. This will give agents observability on recording of `null` attributes outside of audit logs.

> "empty" values such as `""` and `0` **MUST** be sent as an attribute in the payload.

It also performs some minor cleanup and adds descriptions to some function inside our attribute filtering system.
It also contains a fix for previously failing sklearn tests due to a non integer version: `0.0.post12`.

Recommended review strategy: review each commit one at a time.

### DIFF
--- a/newrelic/core/attribute.py
+++ b/newrelic/core/attribute.py
@@ -147,6 +147,12 @@ def create_agent_attributes(attr_dict, attribute_filter):
 
 
 def resolve_user_attributes(attr_dict, attribute_filter, target_destination, attr_class=dict):
+    """
+    Returns an attr_class of key value attributes filtered to the target_destination.
+
+    process_user_attribute MUST be called before this function to filter out invalid
+    attributes.
+    """
     u_attrs = attr_class()
 
     for attr_name, attr_value in attr_dict.items():
@@ -200,11 +206,6 @@ def resolve_logging_context_attributes(attr_dict, attribute_filter, attr_prefix,
                 )
 
     return c_attrs
-
-
-def create_user_attributes(attr_dict, attribute_filter):
-    destinations = DST_ALL
-    return create_attributes(attr_dict, destinations, attribute_filter)
 
 
 def truncate(text, maxsize=MAX_ATTRIBUTE_LENGTH, encoding="utf-8", ending=None):
--- a/newrelic/core/attribute.py
+++ b/newrelic/core/attribute.py
@@ -130,6 +130,13 @@ def create_attributes(attr_dict, destinations, attribute_filter):
 
 
 def create_agent_attributes(attr_dict, attribute_filter):
+    """
+    Returns a dictionary of Attribute objects with appropriate destinations.
+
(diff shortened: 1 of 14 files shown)


newrelic/core/attribute.py
```diff
@@ -311,9 +324,29 @@ def sanitize(value):
...
+    #   debug logging when a null value is recorded. This will give agents observability
+    #   on recording of `null` attributes outside of audit logs.
+    #
+    #   "empty" values such as `""` and `0` **MUST** be sent as an attribute in the
+    #   payload.
+    if value is None:
+        raise NullValueException(
+            "Attribute value is of type: None. Omitting value since there is "
+            "no difference between omitting the key and sending None."
+        )
```
TimPansino, line 354: I think you could probably shorten this comment to a summary instead of copying in the entire agent spec. We typically don't stick the agent spec inline as it can change.

tests/logger_loguru/test_attributes.py
```diff
@@ -53,7 +53,7 @@ def _patcher(d):
         bound_logger.error("context_attrs: {}", "arg1", kwarg_attr=4)
 
 
-@validate_log_events([{"message": "exc_info"}], required_attrs=["context.exception"])
+@validate_log_events([{"message": "exc_info"}], required_attrs=["context.file"])
 @validate_log_event_count(1)
 @background_task()
 def test_loguru_exception_context_attributes(logger):
```
TimPansino, line 59: If this test is for exception context attributes, why are we not checking the exception attribute? Shouldn't that be set?

TimPansino (commented):
Left comments, I've reviewed all the commits and it looks mostly good overall.

tests/logger_logging/test_attributes.py
```diff
@@ -58,7 +58,7 @@ def test_logging_extra_attributes(logger):
     logger.error("extras", extra={"extra_attr": 1})
 
 
-@validate_log_events([{"message": "exc_info"}], required_attrs=["context.exc_info", "context.exc_text"])
+@validate_log_events([{"message": "exc_info"}], required_attrs=["context.exc_info"])
```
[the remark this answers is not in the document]
hmstepanek: Leaving these as is because at the time of instrumentation these are None. We determined that these are set in a formatter which is applied after we instrument.

TimPansino (approved):

PR #1065, opened 2024-02-17
Update PyPy in CI to 3.10

TimPansino:
# Overview

* Add PyPy3.10 to CI image
* Fix a broken settings test that seems to throw errors on PyPy3.10

### DIFF
--- a/.github/containers/Dockerfile
+++ b/.github/containers/Dockerfile
@@ -96,7 +96,7 @@ RUN echo 'eval "$(pyenv init -)"' >>${HOME}/.bashrc && \
     pyenv update
 
 # Install Python
-ARG PYTHON_VERSIONS="3.11 3.10 3.9 3.8 3.7 3.12 2.7 pypy2.7-7.3.12 pypy3.8-7.3.11"
+ARG PYTHON_VERSIONS="3.11 3.10 3.9 3.8 3.7 3.12 2.7 pypy2.7-7.3.12 pypy3.10-7.3.15 pypy3.8-7.3.11"
 COPY --chown=1000:1000 --chmod=+x ./install-python.sh /tmp/install-python.sh
 RUN /tmp/install-python.sh && \
     rm /tmp/install-python.sh
--- a/.github/containers/Dockerfile
+++ b/.github/containers/Dockerfile
@@ -96,7 +96,7 @@ RUN echo 'eval "$(pyenv init -)"' >>${HOME}/.bashrc && \
     pyenv update
 
 # Install Python
-ARG PYTHON_VERSIONS="3.11 3.10 3.9 3.8 3.7 3.12 2.7 pypy2.7-7.3.12 pypy3.10-7.3.15 pypy3.8-7.3.11"
+ARG PYTHON_VERSIONS="3.11 3.10 3.9 3.8 3.7 3.12 2.7 pypy2.7-7.3.12 pypy3.10-7.3.15"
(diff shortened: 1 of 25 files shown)


.github/containers/Dockerfile
```diff
@@ -96,7 +96,7 @@ RUN echo 'eval "$(pyenv init -)"' >>${HOME}/.bashrc && \
     pyenv update
 
 # Install Python
-ARG PYTHON_VERSIONS="3.11 3.10 3.9 3.8 3.7 3.12 2.7 pypy2.7-7.3.12 pypy3.8-7.3.11"
+ARG PYTHON_VERSIONS="3.11 3.10 3.9 3.8 3.7 3.12 2.7 pypy2.7-7.3.12 pypy3.10-7.3.15 pypy3.8-7.3.11"
```
hmstepanek, line 99: I think we can just drop pypy3.8 here now can't we?

tests/mlmodel_sklearn/test_cluster_models.py
```diff
@@ -19,10 +19,10 @@
 )
 
 from newrelic.api.background_task import background_task
```
hmstepanek, line 21: These changes are in another PR so let's just remove the sklearn changes from this PR.

tests/agent_unittests/test_utilization_settings.py
```diff
@@ -118,7 +118,7 @@ def reset(wrapped, instance, args, kwargs):
     return reset
 
 
-@reset_agent_config(INI_FILE_WITHOUT_UTIL_CONF, ENV_WITHOUT_UTIL_CONF)
+@reset_agent_config(INI_FILE_WITHOUT_UTIL_CONF, {"NEW_RELIC_HOST": "collector.newrelic.com"})
```
hmstepanek, line 121: It's concerning to me that the HOST is needed here. It shouldn't need the host to default the OTLP_HOST. Can you explain more about this change/why this test was failing? I just tried running it locally and it didn't fail for me.

hmstepanek (commented):
A couple minor points of feedback. Overall the switch from pypy3.8 to pypy3.10 looks good.

.github/containers/Dockerfile
TimPansino: I think there might be at least 1 framework that doesn't support pypy3.8, I have 3 left that haven't passed yet with 3.10. Will take another look tomorrow to see if I can get them all to pass and I'll drop 3.8 if that works.

tests/agent_unittests/test_utilization_settings.py
TimPansino: This test is really annoying and fails if you've got NEW_RELIC_HOST set to point to staging. It causes my local instance to fail all the time.

If you've got a better idea to make this test not fail I'd be happy to change it.

.github/containers/Dockerfile
TimPansino: I think I got them all running now, I dropped 3.8 completely

tests/agent_unittests/test_utilization_settings.py
hmstepanek: I was thinking we had other tests for when this was specifically set that would make this change to it a duplicate test but I don't actually see any so I guess this change is ok.

hmstepanek (approved):
Looks good!

TimPansino:
Last test run passing, switching the CI image tag back to latest

https://github.com/newrelic/newrelic-python-agent/actions/runs/7982302622?pr=1065

PR #1066, opened 2024-02-17
Fix scikit-learn version checks

TimPansino:
# Overview

* Fix version checking in scikit-learn tests to handle non-integer patch versions


hmstepanek (changes requested):
I don't think we have to make all these other changes. I think you can just replace the version line with: `SKLEARN_VERSION = get_package_version_tuple("sklearn")`.

hmstepanek:
Closing this and using #1067 instead.

PR #1067, opened 2024-02-20
Fix sklearn version tuple parsing in tests

hmstepanek:
Utilize the common parsing method that already handels non integer versions rather than building our own.


TimPansino (approved):

TimPansino:
As long as this doesn't fail on non-integer patch versions it looks fine

PR #1069, opened 2024-02-20
Add `ai.monitoring.enabled` config setting

umaannamalai:
This PR adds the `ai.monitoring.enabled` config setting


tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -173,14 +173,26 @@ def test_bedrock_chat_completion_outside_txn(set_trace_info, exercise_model):
...
 )
-@background_task(name="test_bedrock_chat_completion_disabled_settings")
-def test_bedrock_chat_completion_disabled_settings(set_trace_info, exercise_model):
+@background_task(name="test_bedrock_chat_completion_disabled_custom_events_settings")
+def test_bedrock_chat_completion_disabled_custom_events_settings(set_trace_info, exercise_model):
+    set_trace_info()
+    exercise_model(prompt=_test_bedrock_chat_completion_prompt, temperature=0.7, max_tokens=100)
+
+
+disabled_ai_monitoring_settings = {"ai_monitoring.enabled": False}
```
hmstepanek, line 188: For the custom events disabled - I think we can delete these tests entirely. You don't have to do that here in this PR if you don't want to we can do it in a follow up. Originally they were here for ml events but for custom events there's a common test to verify events aren't created when this setting is disabled so I don't think we need to have a test in every single framework we instrument for it. 

As for the ai_monitoring let's do the following for each framework instead:
Inside conftest.py add the following:
```
disabled_ai_monitoring_settings = override_application_settings({"ai_monitoring.enabled": False})
```
Then inside each of the test files you can import and use this in place of the existing fixture:
```
from conftest import disabled_ai_monitoring_settings

@disabled_ai_monitoring_settings
def test...
```

hmstepanek (changes requested):
One suggestion that I think will help cleanup the code a bit and another that just reduces the overall code churn. Overall this looks really good though.

tests/external_botocore/test_bedrock_chat_completion.py
umaannamalai: I have a test cleanup PR planned so I will add this to that 👍🏽

hmstepanek (approved):

PR #1068, opened 2024-02-20
Add ai_monitoring.streaming.enabled setting

hmstepanek:
Add ai_monitoring.streaming.enabled setting.
* `ai_monitoring.streaming.enabled` -- When disabled does not instrument the stream and does not record events.
* Records a supportability metric when streaming is disabled.

PR #1070, opened 2024-02-21
Run langchain agent tests

hmstepanek:
Previously, langchain agent tests were skipped because they were dependent on streaming instrumentation support which hadn't been implemented yet. Now we support streaming in openai so remove the skip condition and actually run these tests.

Note: this is dependent on #1068 which should be merged first.


umaannamalai (approved):
Looks good! Will wait on merging until the streaming settings PR is merged.

PR #1077, opened 2024-03-01
Add ML supportability metrics

umaannamalai:
This PR updates Bedrock, LangChain, and OpenAI framework metrics to supportability metrics for cross-agent consistency.


hmstepanek (changes requested):
I was gonna say this looks great and merge it but since we added more streaming tests it's failing now cause those tests need to look for the new metric name. After we fix those this should be good to go though!

hmstepanek (approved):
Looks good.

PR #1083, opened 2024-03-05
Add trace based feedback.

umaannamalai:
This PR adds support for trace based feedback

### DIFF
--- a/newrelic/api/ml_model.py
+++ b/newrelic/api/ml_model.py
@@ -58,9 +58,7 @@ def get_llm_message_ids(response_id=None):
     return []
 
 
-def record_llm_feedback_event(
-    trace_id, rating, category=None, message=None, metadata=None
-):
+def record_llm_feedback_event(trace_id, rating, category=None, message=None, metadata=None):
     transaction = current_transaction()
     if not transaction:
         warnings.warn(
--- a/newrelic/api/ml_model.py
+++ b/newrelic/api/ml_model.py
@@ -59,7 +59,7 @@ def get_llm_message_ids(response_id=None):
 
 
 def record_llm_feedback_event(
-    message_id, rating, conversation_id=None, request_id=None, category=None, message=None, metadata=None
+    trace_id, rating, category=None, message=None, metadata=None
 ):
     transaction = current_transaction()
     if not transaction:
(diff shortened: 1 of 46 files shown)


newrelic/api/ml_model.py
```diff
@@ -74,10 +72,8 @@ def record_llm_feedback_event(
     feedback_message_event.update(
         {
             "id": feedback_message_id,
```
hmstepanek, line 74: Could you change the name of this variable to be `feedback_event_id` instead of `feedback_message_id`. Since it's not tied to a specific message anymore, I think this new name is more appropriate.

tests/agent_features/test_record_llm_feedback_event.py
```diff
@@ -46,12 +45,10 @@ def test_record_llm_feedback_event_all_args_supplied():
...
         record_llm_feedback_event(
             rating=1,
-            message_id="message_id",
+            trace_id="123456789abcdefgh",
             category="informative",
-            request_id="request_id",
-            conversation_id="conversation_id",
             message="message",
-            metadata={"foo": "bar", "message": "custom-message"},
+            metadata={"foo": "bar", "my-message": "custom-message"},
```
hmstepanek, line 51: Leave this as message - it's intentionally testing that the user can't override the builtins by passing a metadata key of the same name. Perhaps we should add a comment here explaining this so it's more clear?
```suggestion
            metadata={"foo": "bar", "message": "custom-message"},
```

tests/agent_features/test_record_llm_feedback_event.py
```diff
@@ -29,14 +29,13 @@ def test_record_llm_feedback_event_all_args_supplied():
...
+                "trace_id": "123456789abcdefgh",
                 "category": "informative",
                 "rating": 1,
-                "message_id": "message_id",
-                "request_id": "request_id",
-                "conversation_id": "conversation_id",
                 "ingest_source": "Python",
                 "message": "message",
                 "foo": "bar",
+                "my-message": "custom-message",
```
hmstepanek, line 38:
```suggestion
```

hmstepanek (changes requested):
Just a couple minor things otherwise looks good!

hmstepanek (approved):
Looks good.

PR #1087, opened 2024-03-08
Add attribute value limit exceptions for LLM input/content

lrafeei:
# Overview
This PR removes the attribute character limitation for specific LLMs' inputs/content

Event name | Attribute name
-- | --
LlmChatCompletionMessage | content
LlmEmbedding | input


hmstepanek (changes requested):
One small request-otherwise looks good.

tests/agent_features/test_custom_events.py
```diff
@@ -229,3 +231,103 @@ def test_transaction_create_custom_event_not_called():
...
+
+
+@pytest.mark.parametrize(
+    "max_val,expected_events",
+    (
+        (255, recorded_events_256),
+        (4095, recorded_events_4096),
+    ),
+)
+def test_create_custom_event_no_limit(max_val, expected_events):
```
hmstepanek, line 318: Let's take full advantage or parametrize here and do something like the code below instead. I find this more explicit and concise as it has both the input and expected output next to each other. This way we aren't testing multiple events at once as well-which is generally considered better practice. There's that quote about, "Tests should test one and one thing only." Admittedly this is something we do very poorly in our code base today but we can still try to move in the right direction when writing new tests. Given that the 255 or 4095 is just the max attribute setting, I don't think we need to test the full min and max range here. We have other tests to make sure the setting is set correctly based on it's allowed range. Maybe just pick a random value for max attribute like 300 and roll with it. Also note I dropped the if transaction from this. There should always be a transaction because you are using the background_task decorator.
```suggestion
@pytest.mark.parametrize(
    "event_type,event_data,expected_event_data",
    (
        [
            "LlmChatCompletionMessage",
            {
                "content": "A" * 9000,
                "input": "B" * 9000,
                "foo": "b"* 9000,
            },
            {
                "content": "A" * 9000,
                "input": "B" * 300,
                "foo": "b" * 300,
            },
        ],
        [
            "LlmEmbedding",
            {
                "content": "A" * 5001,
                "input": "B" * 5001,
                "foo": "b"* 5001,
            },
            {
                "content": "A" * 300,
                "input": "B" * 5001,
                "foo": "b" * 300,
            },
        ],
        [
            "MyCustomEvent",
            {
                "content": "A" * 5001,
                "input": "B" * 5001,
                "foo": "b"* 5001,
            },
            {
                "content": "A" * 300,
                "input": "B" * 300,
                "foo": "b" * 300,
            },
        ],
    ),
)
def test_create_custom_event_no_limit(event_type, event_data, expected_event_data):
    @reset_core_stats_engine()
    @override_application_settings({"custom_insights_events.max_attribute_value": 300})
    @validate_custom_event_count(3)
    @validate_custom_events(({"type": event_type}, expected_event_data))
    @background_task()
    def _test():
        transaction.record_custom_event(event_type, event_body)

    _test()

```

hmstepanek (changes requested):
Apologies for not posting this on the original review. I was looking at the tests a bit closer and I think we can simplify these tests.

hmstepanek (approved):
You should be good to merge this now that we've merged the bedrock streaming PR.

PR #1089, opened 2024-03-12
Move validate_custom_event validators into validators directory

lrafeei:
# Overview
This PR moves `validate_custom_event_in_application_stats_engine` and `validate_custom_event_count` into its own separate file within the testing_support/validators directory


TimPansino (approved):
LGTM

PR #1091, opened 2024-03-12
Bedrock Streaming Error Testing

TimPansino:
# Overview

* Expand error testing for botocore's bedrock instrumentation.
  * Includes a variety of corner cases where exceptions are thrown in different parts of the streaming flow.
* Update instrumentation for new corner cases identified in error handling.


tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -416,5 +440,366 @@ def _test():
...
+        chunks = list(response["body"])
+        assert chunks, "No response chunks returned"
+        for chunk in chunks:
+            with pytest.raises(json.decoder.JSONDecodeError):
+                json.loads(chunk["chunk"]["bytes"])
+
+    _test()
+
+
+_headers_error = botocore.eventstream.InvalidHeadersLength
```
hmstepanek, line 603: These are only used in one place. Let's move these to where they are used instead of abstracting them into a constant somewhere else in the test file. This is a really terrible pattern we seem to use all over our tests. It makes the code a lot harder to read when we do this. If it's used in multiple places-it makes sense to use this pattern but here, it's only used once so let's just put it where it's used-there's no reason to add a layer of abstraction.

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -416,5 +440,366 @@ def _test():
...
+_headers_error_name = "botocore.eventstream:InvalidHeadersLength"
+
+
+@reset_core_stats_engine()
+def test_bedrock_chat_completion_error_malformed_response_streaming_chunk(
+    bedrock_server,
+    set_trace_info,
+):
+    """
+    After a streaming call, the chunk returned by the server is invalid and cannot be parsed.
```
hmstepanek, line 613: I'm finding these descriptions really hard to understand. It's not your fault-this is really difficult to explain. I think what might help here is being more explicit about the malformation and forming it more like a story rather than individual bullet points/facts. So for example here, we can say something like:
```
The chunk inside the stream returned by the server is missing the prelude which causes an InvalidHeadersLength exception to be raised during parsing of the chunk. Since the streamed chunk is not able to be parsed, the response attribute on the raised exception is not present. This means all streamed response data will be missing from the recorded events.
```

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -416,5 +440,366 @@ def _test():
...
+
+
+@reset_core_stats_engine()
+@disabled_ai_monitoring_record_content_settings
+def test_bedrock_chat_completion_error_streaming_exception_no_content(
+    bedrock_server,
+    set_trace_info,
+):
+    """
+    Duplicate of the above test, but with content recording is disabled.
```
hmstepanek, line 741:
```suggestion
    Duplicate of `test_bedrock_chat_completion_error_streaming_exception`, but content recording is disabled.
```

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -416,5 +440,366 @@ def _test():
...
+_event_stream_error_name = "botocore.exceptions:EventStreamError"
+
+
+@reset_core_stats_engine()
+def test_bedrock_chat_completion_error_streaming_exception(
+    bedrock_server,
+    set_trace_info,
+):
+    """
+    During a streaming call, the chunk's headers indicate an exception.
```
hmstepanek, line 678: Re-writing another one as a second example:
```
During a streaming call, the streamed chunk's headers indicate an error. The streamed chunk's response body contains the error information from the server. The library uses the streamed chunk's error information to raise an actual exception. The response attribute on the raised exception will contain the error information. This means all response message data will be missing from the recorded events since the server returned an error instead of message data inside the streamed response.
```

hmstepanek (changes requested):
Thank you for writing all these error handling tests and conditions-these were not easy! I would like to improve these descriptions before merging so that when we look at this in the future, it's very clear what's going on here. I had some suggestions for that and some other minor points of feedback. Let me know what you think of the story-based format. I felt like it was clearer to me when written this way but I'm open to other ideas.

tests/external_botocore/conftest.py
```diff
@@ -140,16 +140,18 @@ def wrap_botocore_endpoint_Endpoint__do_get_response(wrapped, instance, args, kw
...
     try:
-        if isinstance(request.body, io.BytesIO):
-            request.body.seek(0)
-            body = json.loads(request.body.read())
-        else:
-            body = json.loads(request.body)
+        content = json.loads(body)
     except Exception:
-        body = {}
+        content = body.decode("utf-8")
```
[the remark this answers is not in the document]
TimPansino: No actually this is to handle malformed JSON payloads, while still being able to respond with the mock server

newrelic/hooks/external_botocore.py
```diff
@@ -658,8 +661,11 @@ def close(self):
 
 def record_stream_chunk(self, return_val):
     if return_val:
-        chunk = json.loads(return_val["chunk"]["bytes"].decode("utf-8"))
-        self._nr_model_extractor(chunk, self._nr_bedrock_attrs)
+        try:
```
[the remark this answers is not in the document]
TimPansino: I added a log message to this one, see if that sufficiently answers the question. If not we could add more comments.

hmstepanek (changes requested):
The descriptions on the error tests are much clearer now-thank you for re-wording those! I had some additional thoughts/suggestions about the log messages/try-catch blocks. Otherwise this looks really good!

newrelic/hooks/external_botocore.py
```diff
@@ -38,6 +39,12 @@
 
 
 _logger = logging.getLogger(__name__)
+
+EXCEPTION_HANDLING_FAILURE_LOG_MESSAGE = "Exception occured in botocore instrumentation for AWS Bedrock: While reporting an exception in botocore, another exception occured. Report this issue to New Relic Support.\n%s"
+REQUEST_EXTACTOR_FAILURE_LOG_MESSAGE = "Exception occured in botocore instrumentation for AWS Bedrock: Failed to extract request information. Report this issue to New Relic Support.\n%s"
+RESPONSE_EXTRACTOR_FAILURE_LOG_MESSAGE = "Exception occured in botocore instrumentation for AWS Bedrock: Failed to extract response information. If the issue persists, report this issue to New Relic support.\n%s"
+RESPONSE_PROCESSING_FAILURE_LOG_MESSAGE = "Exception occured in botocore instrumentation for AWS Bedrock: Failed to report response data. Report this issue to New Relic Support.\n%s"
```
hmstepanek, line 46:
```suggestion
RESPONSE_PROCESSING_FAILURE_LOG_MESSAGE = "Exception occurred in botocore instrumentation for AWS Bedrock: Failed to report response data. Report this issue to New Relic Support.\n%s"
```

newrelic/hooks/external_botocore.py
```diff
@@ -38,6 +39,12 @@
 
 
 _logger = logging.getLogger(__name__)
+
+EXCEPTION_HANDLING_FAILURE_LOG_MESSAGE = "Exception occured in botocore instrumentation for AWS Bedrock: While reporting an exception in botocore, another exception occured. Report this issue to New Relic Support.\n%s"
+REQUEST_EXTACTOR_FAILURE_LOG_MESSAGE = "Exception occured in botocore instrumentation for AWS Bedrock: Failed to extract request information. Report this issue to New Relic Support.\n%s"
+RESPONSE_EXTRACTOR_FAILURE_LOG_MESSAGE = "Exception occured in botocore instrumentation for AWS Bedrock: Failed to extract response information. If the issue persists, report this issue to New Relic support.\n%s"
```
hmstepanek, line 45:
```suggestion
RESPONSE_EXTRACTOR_FAILURE_LOG_MESSAGE = "Exception occurred in botocore instrumentation for AWS Bedrock: Failed to extract response information. If the issue persists, report this issue to New Relic support.\n%s"
```

newrelic/hooks/external_botocore.py
```diff
@@ -38,6 +39,12 @@
 
 
 _logger = logging.getLogger(__name__)
+
+EXCEPTION_HANDLING_FAILURE_LOG_MESSAGE = "Exception occured in botocore instrumentation for AWS Bedrock: While reporting an exception in botocore, another exception occured. Report this issue to New Relic Support.\n%s"
+REQUEST_EXTACTOR_FAILURE_LOG_MESSAGE = "Exception occured in botocore instrumentation for AWS Bedrock: Failed to extract request information. Report this issue to New Relic Support.\n%s"
```
hmstepanek, line 44:
```suggestion
REQUEST_EXTACTOR_FAILURE_LOG_MESSAGE = "Exception occurred in botocore instrumentation for AWS Bedrock: Failed to extract request information. Report this issue to New Relic Support.\n%s"
```

newrelic/hooks/external_botocore.py
```diff
@@ -38,6 +39,12 @@
 
 
 _logger = logging.getLogger(__name__)
+
+EXCEPTION_HANDLING_FAILURE_LOG_MESSAGE = "Exception occured in botocore instrumentation for AWS Bedrock: While reporting an exception in botocore, another exception occured. Report this issue to New Relic Support.\n%s"
```
hmstepanek, line 43:
```suggestion
EXCEPTION_HANDLING_FAILURE_LOG_MESSAGE = "Exception occurred in botocore instrumentation for AWS Bedrock: While reporting an exception in botocore, another exception occurred. Report this issue to New Relic Support.\n%s"
```

hmstepanek (approved):

PR #1094, opened 2024-03-13
Add LLM token count callback.

umaannamalai:
This PR adds a new, public callback `set_llm_token_count_callback`. This callback calls a user function to calculate total LLM token counts for a chat completion or embedding and stores in on the corresponding LLM events.


tests/agent_features/test_llm_token_count_callback.py
```diff
@@ -0,0 +1,70 @@
...
+    set_llm_token_count_callback(basic_callback)
+    assert settings.ai_monitoring.llm_token_count_callback is not None, "Failed to set callback."
+    set_llm_token_count_callback(None)
+    assert settings.ai_monitoring.llm_token_count_callback is None, "Failed to clear callback."
+
+
+@pytest.mark.parametrize(
+    "callback,accepted", [(basic_callback, True), (lambda x, y: None, True), (None, False), ("string", False)]
+)
+def test_set_llm_token_count_callback(callback, accepted):
```
hmstepanek, line 61: Let's cleanup these tests to look like the following:
```
@pytest.mark.parametrize(
    "set_args,call_args,expected_value",
    [
        ((lambda model, content: 45,), ("model", "content"), 45),
        ((lambda model, content: 45, application().settings), ("model", "content"), 45),
        ((lambda model, content: 1.1,), ("model", "content"), None),
        ((lambda model, content: -1,), ("model", "content"), None),
        ((lambda model, content: 45,), (None, "content"), None),
        ((lambda model, content: 45,), ("model", None), None),
    ]
)
def test_set_llm_token_count_callback(set_args, call_args, expected_value):
    set_llm_token_count_callback(*set_args)
    assert settings.ai_monitoring.llm_token_count_callback(*call_args) == expected_value


def test_unset_llm_token_count_callback():
    set_llm_token_count_callback(lambda model, content: 45)
    assert callable(settings.ai_monitoring.llm_token_count_callback)
    set_llm_token_count_callback(None)
    
    assert settings.ai_monitoring.llm_token_count_callback is None

# Another test for if the application isn't active.

# Another test for if the callback raises an exception - it should pass through and not be squashed.
```

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -143,6 +147,15 @@ def expected_events_no_content(expected_events):
     return events
 
 
+@pytest.fixture(scope="module")
+def expected_events_with_token_count(expected_events):
+    events = copy.deepcopy(expected_events)
+    for event in events:
+        if event[0]["type"] != "LlmChatCompletionSummary":
```
hmstepanek, line 154:
```suggestion
        if event[0]["type"] == "LlmChatCompletionMessage":
```

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -255,6 +268,53 @@ def _test():
     _test()
 
 
+@pytest.mark.parametrize(
+    "llm_token_callback",
+    [
+        llm_token_count_callback_success,
```
hmstepanek, line 274: Given that we are now extensively testing the common implementation in agent features we can reduce the tests here to with and without the token counting callback. Note we already have tests without the callback so we just need to add tests with the callback returning a valid count.

hmstepanek: Note we _do_ need to test running with the callback during error conditions to make sure we are still capturing token counts where we have the content during error conditions.

hmstepanek (changes requested):
Thank you for implementing this so quickly! I have a lot of feedback here. Some cleanup, some moving around of code, and some missing tests. Let me know if you have any questions about any of it. Happy to do some pairing on it if need be!

tests/external_botocore/test_bedrock_chat_completion.py
umaannamalai: This was kept with the negative check because the same function is used for embeddings.

newrelic/api/ml_model.py
```diff
@@ -61,3 +65,62 @@ def record_llm_feedback_event(trace_id, rating, category=None, message=None, met
...
+            _logger.debug(
+                "The content argument passed to the user-defined token calculation callback is None. The callback will not be run."
+            )
+            return None
+
+        token_count_val = callback(model, content)
+
+        if not isinstance(token_count_val, int) or token_count_val < 0:
+            _logger.warning(
+                "llm_token_count_callback returned an invalid value of %s. This value must be a positive integer."
```
hmstepanek, line 119:
```suggestion
                "llm_token_count_callback returned an invalid value of %s. This value must be a positive integer. This value will not be recorded for the token_count."
```

newrelic/hooks/mlmodel_openai.py
```diff
@@ -559,13 +582,15 @@ def _record_completion_error(transaction, linking_metadata, completion_id, kwarg
     output_message_list = []
     if "content" in kwargs:
         output_message_list = [{"content": kwargs.get("content"), "role": kwargs.get("role")}]
+
```
hmstepanek, line 585:
```suggestion
```

tests/agent_features/test_llm_token_count_callback.py
```diff
@@ -0,0 +1,88 @@
...
+def test_exception_in_user_callback():
+    settings = application().settings
+
+    def user_exc():
+        raise TypeError()
+
+    set_llm_token_count_callback(user_exc)
+
+    with pytest.raises(TypeError):
+        assert settings.ai_monitoring.llm_token_count_callback("model", "content")
```
hmstepanek, line 81:
```suggestion
        settings.ai_monitoring.llm_token_count_callback("model", "content")
```

tests/agent_features/test_llm_token_count_callback.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def test_with_application_not_active():
+    settings = application(activate=False).settings
+
+    set_llm_token_count_callback(lambda model, content: 45)
```
hmstepanek, line 87: I like to kinda group the tests into the following where each section is separated by and empty line.
```
# test setup

# run the test

# test assertions
```
```suggestion
    set_llm_token_count_callback(lambda model, content: 45)

```

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -144,6 +147,15 @@ def expected_events_no_content(expected_events):
     return events
 
 
+@pytest.fixture(scope="module")
+def expected_events_with_token_count(expected_events):
```
hmstepanek, line 151: I'm going to suggest copying something Lalleh did in her PR where she made a generic function in conftest that goes through and modifies any expected event input generically. It looks like you did this in openai as well. I think this is overall simpler than trying to utilize a fixture as the expected events are not named the same in each test.
So in conftest put the following:
```
def add_token_count_to_message_events(expected_events):
    events = copy.deepcopy(expected_events)
    for event in events:
        if event[0]["type"] == "LlmChatCompletionMessage":
            event[1]["token_count"] = 105
    return events
```

And then you can use it everywhere like:
```
@validate_custom_events(add_token_count_to_message_events(expected_events_with_llm_metadata))
```

tests/external_botocore/conftest.py
```diff
@@ -66,6 +66,10 @@
 )
 
 
+def llm_token_count_callback_success(model, content):
```
hmstepanek, line 69: I think we can drop the `_success` since there's only one of these now:
```suggestion
def llm_token_count_callback(model, content):
```

tests/external_botocore/test_bedrock_chat_completion.py
```diff
@@ -452,5 +502,49 @@ def _test():
...
+            add_custom_attribute("llm.conversation_id", "my-awesome-id")
+            add_custom_attribute("llm.foo", "bar")
+            add_custom_attribute("non_llm_attr", "python-agent")
+
+            exercise_model(prompt="Invalid Token", temperature=0.7, max_tokens=100)
+
+    _test()
+
+
 def test_bedrock_chat_completion_functions_marked_as_wrapped_for_sdk_compatibility(bedrock_server):
```
hmstepanek, line 549: We need to be testing streaming here too as that takes a different code path. The streaming tests may not have existed at the time that you started working on this.

hmstepanek (changes requested):
Some small fixups and a couple requests for the tests but overall looks really good!

newrelic/hooks/mlmodel_openai.py
umaannamalai: I had intentionally added that newline in for readability to separate out the call to `create_chat_completion_message_event` a bit more 😄

tests/external_botocore/test_bedrock_chat_completion.py
umaannamalai: It looks like none of the expected event test cases in Bedrock follow this approach where it's passed to the event modifying function. It's definitely cleaner as you said so I have updated the token count events to follow this format and will fix the rest of them (no metadata, no content, etc.) in a subsequent cleanup PR.

hmstepanek (approved):
This looks great! I'm just waiting for the tests to pass before merging but I assume they will.

PR #1100, opened 2024-03-15
Bedrock Embedding Updates

TimPansino:
# Overview

* Update testing for bedrock embedding to match chat completion testing.
* Add new log message for unsupported combination of bedrock embedding and response streaming.


hmstepanek (approved):
This looks good to me. The tests that are failing look like that asgi django issue and they don't seem related but there are so many failing it's time consuming to look at every suite that failed.

PR #1102, opened 2024-03-18
Merge main into develop-ai-limited-preview-3

TimPansino:
# Overview

* Merge main contents into `develop-ai-limited-preview-3`

### DIFF
--- a/.github/workflows/build-ci-image.yml
+++ b/.github/workflows/build-ci-image.yml
@@ -11,13 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+---
 name: Build CI Image
 
 on:
-  workflow_dispatch:  # Allow manual trigger
+  workflow_dispatch: # Allow manual trigger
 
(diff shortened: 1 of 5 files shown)


TimPansino:
Make sure not to squash and merge this, use a merge commit instead

.github/workflows/build-ci-image.yml
```diff
@@ -11,13 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+---
```
hmstepanek, line 14: It's really bothering me that there are spaces being added here but given that it's on main I don't think we should be addressing that here.

hmstepanek (approved):
Looks good.

.github/workflows/build-ci-image.yml
TimPansino: That's not a space, it's a `---` marker to signify the end of the yaml headers. It was added by one of our yaml linters.

hmstepanek: ooooh ok all good then!

PR #1105, opened 2024-03-20
Test embedding streaming openai v1

hmstepanek:
# Overview
Add early exit condition to not wrap embedding streaming in openai v1. It appears that streaming for embeddings isn't support in openai v0 though I'm keeping the early exit condition for that in place just in case.

Note no mocked response was added specifically for streamed embeddings. This is because the response in streamed vs unstreamed is the same format, it just sends the data in chunks instead of all at once. In other words: the non-streamed mock response will also work when mocking streaming mode.

### DIFF
--- a/newrelic/hooks/mlmodel_openai.py
+++ b/newrelic/hooks/mlmodel_openai.py
@@ -38,10 +38,14 @@ _logger = logging.getLogger(__name__)
 
 
 def wrap_embedding_sync(wrapped, instance, args, kwargs):
+    breakpoint()
     transaction = current_transaction()
-    if not transaction or kwargs.get("stream", False):
+    if (
+        not transaction
+        or kwargs.get("stream", False)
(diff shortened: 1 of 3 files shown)


tests/mlmodel_openai/conftest.py
```diff
@@ -70,6 +71,7 @@
         "test_chat_completion_error.py",
         "test_chat_completion_stream.py",
         "test_chat_completion_stream_error.py",
+        "test_embeddings_stream.py",
```
hmstepanek, line 74:
```suggestion
```

newrelic/hooks/mlmodel_openai.py
```diff
@@ -38,10 +38,14 @@
 
 
 def wrap_embedding_sync(wrapped, instance, args, kwargs):
+    breakpoint()
```
hmstepanek, line 41:
```suggestion
```

TimPansino (approved):

PR #1109, opened 2024-03-22
Add support for langchain bedrock payloads

hmstepanek:
# Overview
Add tests for calling bedrock via langchain and add support for these payloads into the bedrock instrumentation.


umaannamalai (approved):
Validated this PR in the UI - the response is now loading and printing as we would expect 🚀

tests/external_botocore/conftest.py
```diff
@@ -22,6 +22,7 @@
     MockExternalBedrockServer,
```
umaannamalai, line 22: We may want to add a `collect_ignore` here to not run the langchain tests for older versions of botocore since we only install langchain on botocorelatest

hmstepanek: Really good point! I believe this is already handled by this line in the bedrock_server fixture but I did need to move the langchain imports into the test function as langchain isn't available in earlier botocore envs:
```
    if get_package_version_tuple("botocore") < (1, 31, 57):
        pytest.skip(reason="Bedrock Runtime not available.")
```

PR #1115, opened 2024-03-26
Add support for cohere embeddings.

umaannamalai:
This PR adds support for cohere.embed-english-v3 and cohere.embed-multilingual-v3 models


tests/external_botocore/test_bedrock_embeddings.py
```diff
@@ -77,6 +79,7 @@ def exercise_model(bedrock_server, model_id, request_streaming):
 
     def _exercise_model(prompt):
         body = (payload_template % prompt).encode("utf-8")
+
```
hmstepanek, line 82:
```suggestion
```

newrelic/hooks/external_botocore.py
```diff
@@ -218,7 +218,16 @@ def extract_bedrock_titan_embedding_model_request(request_body, bedrock_attrs):
     return bedrock_attrs
 
 
-def extract_bedrock_titan_embedding_model_response(response_body, bedrock_attrs):
+def extract_bedrock_cohere_embedding_model_request(request_body, bedrock_attrs):
+    request_body = json.loads(request_body)
+
+    try:
+        input_text = str(request_body.get("texts"))
```
hmstepanek, line 225: I'm not following this try catch. It tries to cast it to a string, otherwise it doesn't and it gets cast to a string later by our record custom event attribute validation. Seems like we could simplify this to just:
```
input_text = request_body.get("texts")
```

tests/external_botocore/_mock_external_bedrock_server.py
```diff
@@ -3818,6 +3818,2080 @@
             "inputTextTokenCount": 6,
         },
     ],
+    "cohere.embed-english-v3::This is an embedding test.": [
```
hmstepanek, line 3821: Are these two cohere.embed models different in terms of parsing? If not, then I would just follow the existing precedent here of only testing one of them.

hmstepanek (changes requested):
Looks really good, just a couple minor points of feedback.

hmstepanek (approved):

PR #1117, opened 2024-03-27
Use six from packages

hmstepanek:
# Overview
Replace six imports with package six.
Fix some lint issues at the same time.


TimPansino (approved):
Fix the accidental formatting on urllib3's code and this will be fine

PR #1122, opened 2024-04-10
Add config setting to toggle memory usage runtime metrics.

umaannamalai:
The Python Agent reports a large volume of memory usage related metrics because it sends data up for each PID. This PR adds a setting to allow users to toggle memory usage metrics reporting to reduce overall metrics being sent.


hmstepanek (approved):

hmstepanek (commented):
There's a couple lint errors but otherwise looks good:

PR #1135, opened 2024-05-07
AIOBotocore instrumentation

lrafeei:
This PR provides instrumentation support for AIOBotocore


tests/external_aiobotocore/test_aiobotocore_sqs.py
```diff
@@ -0,0 +1,113 @@
...
+                resp = await client.send_message(QueueUrl=queue_url, MessageBody="hello_world")
+                assert resp["ResponseMetadata"]["HTTPStatusCode"] == 200
+
+                # Receive message
+                # This works in an actual application, but not with the proxy set up
+                # for the test.  The aiobotocore's proxy requests redirect to aiohttp's
+                # request call which results in double instrumentation and invalid XML
+                # generation in response_dict["headers"]["body"]
+                # resp = await client.receive_message(QueueUrl=queue_url)
+                # assert resp["ResponseMetadata"]["HTTPStatusCode"] == 200
```
TimPansino, line 94: Are we sure this actually works with a real AWS connection running through a proxy? Or does the existence of a proxy cause issues here?

lrafeei: In this case, the aiobotocore's proxy functionality uses aiohttp and urllib3 under the hood.  Because it used those under the hood, New Relic's browser monitoring was injecting `_js_agent_header_fragment` into the response.  This in turn caused the response message to be returned as an invalid XML.

newrelic/hooks/external_aiobotocore.py
```diff
@@ -0,0 +1,54 @@
...
+    method = request_dict.get("method")
+
+    with ExternalTrace(library="aiobotocore", url=url, method=method, source=wrapped) as trace:
+        try:
+            # Because AIOBotocore's proxy functionality uses aiohttp
+            # and urllib3 under the hood, New Relic has portions that
+            # are classified as Web Transactions.  This means that
+            # browser monitoring will now be true.  However, this will
+            # inject unwanted JS Agent Header Fragments into SQS responses.
+            trace.settings.browser_monitoring.enabled = False
```
TimPansino, line 39: I don't think this something you can do, the settings object is not a copy but just a reference to the application settings. You're permanently disabling browser monitoring everywhere in this application. 

Browser monitoring shouldn't be at all relevant to an ExternalTrace anyway. The issue you're likely having is the mock server having issues with it being an aiohttp server. Why don't you try just disabling browser monitoring in `conftest`?

TimPansino: Also for the record, the correct way to disable browser monitoring for an individual transaction is `newrelic.agent.disable_browser_autorum()`

lrafeei: I kind of went back and forth with this.  I originally had it disabled in conftest but my rationale here was that, if a customer wanted to use the proxy functionality of aiobotocore SQS, the customer would have to disable the setting themselves just because of how the proxy server in aiobotocore works.  If we are OK with this, I can put something in the docs about this as well.

TimPansino (approved):

TimPansino (approved):

PR #1138, opened 2024-05-10
Support for OpenAI's .with_raw_response.

lrafeei:
This PR adds support for [`.with_raw_response.`](https://github.com/openai/openai-python?tab=readme-ov-file#accessing-raw-response-data-eg-headers) prepending to HTTP method calls in OpenAI.

Currently, the New Relic instrumentation will instrument this, but during the collection of custom AI event attributes, this modified response type (LegacyAPIResponse) will result in a crash.

This is part 1 of two PRs.  The second part will cover the streaming version of this functionality, `.with_streaming_response.`


hmstepanek (changes requested):
Just had one requested change to the implementation otherwise looks good.

hmstepanek (changes requested):
I tested it manually and things show up in the UI. Couple additional fixes but after that I think we're good to merge this one.

hmstepanek (approved):

PR #1149, opened 2024-05-22
Update loguru hook to include milliseconds in timestamp

julia-tadej-wttech:
This PR fixes timestamp for loguru logs. 
Currently, milliseconds for log records coming from loguru are always equal to 0. After this change, milliseconds will be recorded correctly (so the log order will be correct).


TimPansino (approved):

TimPansino:
Thanks @julia-tadej-wttech, we'll get this into the next release.

PR #1151, opened 2024-05-24
Fix LLM event durations

RyanKadri:
# Overview
In order to match behavior across agents, the python agent should report the `duration` field on AI Monitoring events in milliseconds rather than seconds.

# Testing
The test suite does not currently check the `duration` field because the integration tests have variability between runs


RyanKadri:
I ended up just manually converting from seconds to ms where we were referencing the trace duration. LMK if you want this to be in a shared function somewhere

umaannamalai:
We tend to do manual conversion throughout the agent as well so this approach is fine. We can always create a common utility function down the road after this PR gets merged.

umaannamalai (approved):

PR #1155, opened 2024-05-31
Psycopg3 Instrumentation

TimPansino:
# Overview

* Adds instrumentation for Psycopg3
  * Includes both Connection and AsyncConnection instrumentation.
  * Adds new `database_dbapi2_async.py` file which implements the DBAPI2 interface with async functions.
* Add test suite for Psycopg3
  * Includes tests for Python 3.8+, and PyPy
  * Tests pure python implementation, precompiled binary, and the locally built C extensions [recommended for production by psycopg](https://www.psycopg.org/psycopg3/docs/basic/install.html#local-installation)
  * Does NOT test:
    * Python 2 or 3.7 as they are unsupported.
    * psycopg v3.0, as pip for some reason refuses to believe those versions exist despite being available on PyPi.
* Update pytest to get rid of Python 3.12 warnings


TimPansino:
I definitely feel like a few of the lines of missed coverage are being run, maybe they're not picked up due to something in coverage. I'll check that out on Monday.

newrelic/hooks/database_dbapi2_async.py
```diff
@@ -0,0 +1,162 @@
...
+class AsyncConnectionFactory(ObjectProxy):
+
+    __connection_wrapper__ = AsyncConnectionWrapper
+
+    def __init__(self, connect, dbapi2_module):
+        super(AsyncConnectionFactory, self).__init__(connect)
+        self._nr_dbapi2_module = dbapi2_module
+
+    async def __call__(self, *args, **kwargs):
+        rollup = []
```
hmstepanek, line 150: Instead of appending:
```suggestion
        rollup = ["Datastore/all", "Datastore/%s/all" % self._nr_dbapi2_module._nr_database_product]
```

newrelic/hooks/database_psycopg.py
```diff
@@ -0,0 +1,500 @@
...
+                cursor = self.__wrapped__.execute(sql, **kwargs)
+
+        return self.__cursor_wrapper__(cursor, self._nr_dbapi2_module, self._nr_connect_params, (args, kwargs))
+
+    def __enter__(self):
+        name = callable_name(self.__wrapped__.__enter__)
+        with FunctionTrace(name, source=self.__wrapped__.__enter__):
+            self.__wrapped__.__enter__()
+
+        # Must return a reference to self as otherwise will be
```
hmstepanek, line 134: It would be helpful to explain this for the async version as well.

newrelic/hooks/database_psycopg.py
```diff
@@ -0,0 +1,500 @@
...
+
+def instance_info(args, kwargs):
+
+    p_host, p_hostaddr, p_port, p_dbname = _parse_connect_params(args, kwargs)
+    host, port, db_name = _add_defaults(p_host, p_hostaddr, p_port, p_dbname)
+
+    return (host, port, db_name)
+
+
+def _parse_connect_params(args, kwargs):
```
hmstepanek, line 337: Part of me wants to move these helper methods out of here and put them in a shared location. I realize that one is ever so slightly different. I wonder if we could just abstract that difference to an additional input argument or something similar. It would be easier for comparison purposes between the two versions and less code duplication.

hmstepanek (commented):
Still not done yet but wanted to post what I have to give you something to look at.

newrelic/hooks/database_psycopg.py
TimPansino: One of the problems is importing other hooks files will report that instrumentation package as used by the environment reporting feature, which just checks sys.modules for things. If we don't put it in another hooks file we'll be putting code specific to a few instrumentation packages in common places, which messes with project structure. 

I personally feel inclined to just let duplication happen, so that if we change something in this code for psycopg3 it doesn't possibly break psycopg2 in the future. We've gotten bit pretty hard in the past with trying too hard to reuse code an avoid duplication. (See everything async redis related.)

newrelic/hooks/database_psycopg.py
```diff
@@ -0,0 +1,514 @@
...
```
hmstepanek, line 351: I'm having a hard time finding this in the code. Do we handle the case for the connection being created with conninfo as opposed to the separate host, port, etc?
```
    conn = psycopg.connect(
        conninfo="postgres://user:password@localhost:5432/postgres"
    )
```

hmstepanek (commented):
This looks good. Just finished testing it manually. I just had one question regarding the alternative conninfo param option.

newrelic/hooks/database_psycopg.py
TimPansino: Good catch, it was originally an argument named `dsn` but that got renamed to `conninfo`. Our tests for parsing dsn are unit tests that don't actually run a connect, so that wasn't failing when it should. Fixed with the latest commit.

hmstepanek (approved):
Looks good. I notice the environment tests are failing (unrelated to this but I wonder if that sklearn test fix broke something).

PR #1157, opened 2024-05-31
Early exit for OpenAI's .with_streaming_response.

lrafeei:
This is part 2 of https://github.com/newrelic/newrelic-python-agent/pull/1138

For now, an exit condition will be added to instances where this wrapper is used and support for this will be added later.


newrelic/hooks/mlmodel_openai.py
```diff
@@ -403,6 +412,11 @@ async def wrap_chat_completion_async(wrapped, instance, args, kwargs):
 
     completion_id = str(uuid.uuid4())
 
+    # If `.with_streaming_response.` wrapper used, switch to streaming
```
hmstepanek, line 415: For symmetry purposes of the implementation can we move this up into the if not transaction block the same as the async implementation? Right now we are adding the llm true attribute which means it will show up on the AI dashboard as an llm transaction even though we aren't recording any additional info about it which may be confusing to customers.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -90,6 +90,11 @@ def wrap_chat_completion_sync(wrapped, instance, args, kwargs):
     completion_id = str(uuid.uuid4())
     request_message_list = kwargs.get("messages", [])
 
+    # If `.with_streaming_response.` wrapper used, switch to streaming
```
hmstepanek, line 93: Same thing here, let's move this up to the if not transaction early exit so we don't record those attributes on the transaction before exiting.

hmstepanek (changes requested):
Let's move the early exit condition to be a bit earlier, otherwise looks good.

hmstepanek (approved):

PR #1165, opened 2024-06-12
Disable package capture when testing

hmstepanek:
# Overview
Package capturing is slow and we don't need to do it for every test. Disable package capturing unless we are explicitly testing it.

Previously, when it took too long to capture the package info, it caused application connect timeouts that raised an AssertionError of application.active.


tests/mlmodel_sklearn/restats
lrafeei, line 1: Out of curiosity, what is this file?

lrafeei (commented):
I see the flag being set in the tests that do not need the package capture, but I am not seeing where the flag's status does anything with the function itself (in the config.py to set the setting/flag or the package capture function itself)

tests/mlmodel_sklearn/restats
hmstepanek: whoops...it's a cprofile output..how did that get in there!? Removing...

hmstepanek:
> I see the flag being set in the tests that do not need the package capture, but I am not seeing where the flag's status does anything with the function itself (in the config.py to set the setting/flag or the package capture function itself)

The package capture setting is in the config file here: https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/config.py#L567.
It is used here inside the environment_settings function to skip capturing the plugins if it is disabled: https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/core/environment.py#L207.

lrafeei:
> > I see the flag being set in the tests that do not need the package capture, but I am not seeing where the flag's status does anything with the function itself (in the config.py to set the setting/flag or the package capture function itself)
> 
> The package capture setting is in the config file here: https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/config.py#L567. It is used here inside the environment_settings function to skip capturing the plugins if it is disabled: https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/core/environment.py#L207.

Oh, OK, that's what I was missing--I did not realize this was a flag we already had.  Looks good to me!

PR #1168, opened 2024-06-25
NewRelicContextFormatter update to support adding a stack trace

fritzdj:
# Overview
Currently the `NewRelicContextFormatter` implementation doesn't support including the stack trace. It would be convenient if there was an option because it would allow stack frames to be included as an attribute for an log in NR (instead of them being broken up by frame, which is the default if structured logs are not used). This PR is adding opt-in support for including stack traces.

# Related Issue
https://github.com/newrelic/newrelic-python-agent/issues/1169

### DIFF
--- a/newrelic/api/log.py
+++ b/newrelic/api/log.py
@@ -66,7 +66,7 @@ class NewRelicContextFormatter(logging.Formatter):
         if stack_trace_limit is not None:
             if not isinstance(stack_trace_limit, int):
                 raise TypeError("stack_trace_limit must be None or a non-negative integer")
-            elif stack_trace_limit < 0:
+            if stack_trace_limit < 0:
                 raise ValueError("stack_trace_limit must be None or a non-negative integer")
         self._stack_trace_limit = stack_trace_limit
 
--- a/newrelic/api/log.py
+++ b/newrelic/api/log.py
@@ -54,14 +54,14 @@ def safe_json_encode(obj, ignore_string_types=False, **kwargs):
 class NewRelicContextFormatter(logging.Formatter):
     DEFAULT_LOG_RECORD_KEYS = frozenset(set(vars(logging.LogRecord("", 0, "", 0, "", (), None))) | {"message"})
 
-    def __init__(self, *args, stack_trace_limit=0, **kwargs):
+    def __init__(self, *args, **kwargs):
         """
         :param Optional[int] stack_trace_limit:
             Specifies the maximum number of frames to include for stack traces.
             Defaults to `0` to suppress stack traces.
             Setting this to `None` will make it so all available frames are included.
         """
-        super(NewRelicContextFormatter, self).__init__(*args, **kwargs)
+        stack_trace_limit = kwargs.pop("stack_trace_limit", 0)
 
         if stack_trace_limit is not None:
             if not isinstance(stack_trace_limit, int):
@@ -69,6 +69,8 @@ class NewRelicContextFormatter(logging.Formatter):
             elif stack_trace_limit < 0:
                 raise ValueError("stack_trace_limit must be None or a non-negative integer")
         self._stack_trace_limit = stack_trace_limit
+
+        super(NewRelicContextFormatter, self).__init__(*args, **kwargs)
 
     @classmethod
     def format_exc_info(cls, exc_info, stack_trace_limit=0):
--- a/newrelic/api/log.py
+++ b/newrelic/api/log.py
@@ -16,6 +16,8 @@ import json
 import logging
 import re
 import warnings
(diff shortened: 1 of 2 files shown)


newrelic/api/log.py
```diff
@@ -69,11 +55,46 @@ def safe_json_encode(obj, ignore_string_types=False, **kwargs):
...
-    def __init__(self, *args, **kwargs):
-        super(NewRelicContextFormatter, self).__init__()
+    def __init__(self, *args, stack_trace_limit: Optional[int] = 0, **kwargs):
+        """
+        :param Optional[int] stack_trace_limit:
+            Specifies the maximum number of frames to include for stack traces.
+            Defaults to `0` to suppress stack traces.
+            Setting this to `None` will make it so all available frames are included.
+        """
+        super(NewRelicContextFormatter, self).__init__(*args, **kwargs)
```
fritzdj, line 65: Reviewer note: `*args` and `**kwargs` were not being passed through to `logging.Formatter` init. This might have been a bug.

newrelic/api/log.py
```diff
@@ -69,11 +55,46 @@ def safe_json_encode(obj, ignore_string_types=False, **kwargs):
...
+        fullname = fullnames[0]
+
+        formatted = {
+            "error.class": fullname,
+            "error.message": message,
+        }
 
-    @classmethod
-    def log_record_to_dict(cls, record):
+        expected = is_expected_error(exc_info)
```
TimPansino: An issue here, we're taking `log_record_to_dict()` which is a classmethod and changing it to an instance method. This would be a breaking change to the public API.

The same goes for `format_exc_info()` changing from a top level function to an instance method.

We definitely have customers who use these functions, or at least `log_record_to_dict()` without instantiating a class for convenience.

I'll push a commit with my proposal, but I'd like to convert these both into class functions, and export the names to the top level of the module so they can be used like functions. I think passing in the stack trace limit as an argument to the functions should preserve the most backwards compatibility for other users.

newrelic/api/log.py
```diff
@@ -69,11 +55,46 @@ def safe_json_encode(obj, ignore_string_types=False, **kwargs):
 class NewRelicContextFormatter(logging.Formatter):
     DEFAULT_LOG_RECORD_KEYS = frozenset(set(vars(logging.LogRecord("", 0, "", 0, "", (), None))) | {"message"})
 
-    def __init__(self, *args, **kwargs):
-        super(NewRelicContextFormatter, self).__init__()
+    def __init__(self, *args, stack_trace_limit: Optional[int] = 0, **kwargs):
```
TimPansino:
```suggestion
    def __init__(self, *args, stack_trace_limit=0, **kwargs):
```

We still have Python 2 wheels at the moment, so this line is a syntax issue. Should be removed.

TimPansino: Actually having a default keyword argument after *args isn't supported in Python 2 at all, I'll remove it for now like we do for our trace APIs and just attempt to pop it off the kwargs list if it exists.

TimPansino:
I've several pushed commits for both @fritzdj and our team to review. Here's the summary of my changes.

* Python 2 issues in the code and tests fixed with transparent changes.
* The public APIs for `log_record_to_dict()` and `format_exc_info` were originally changed in a backwards incompatible manner.
  * These have been changed to be class methods of the `NewRelicContextFormatter` to maintain compatibility, and exported as top level names in the `newrelic.api.log` module for the same reason.
  * As part of this change, `stack_trace_limit` was added as a function argument to both of these APIs and will be supplied automatically by calls to `format()` if using the Formatter as an object, or can be supplied otherwise.

I think this will achieve the most flexible compatibility possible with existing customer uses, while making this feature possible. It's worth noting that in all cases the default behavior will remain unchanged, where no stack traces are attached.

fritzdj:
Thanks @TimPansino, these changes look good to me! Good catches on the breaking change to the public API and Python 2 compatibility issues.

lrafeei (approved):

rafaelclp:
We extended the `NewRelicContextFormatter` to add more data to the log record dict:

```python
class OurNewRelicContextFormatter(NewRelicContextFormatter):
    @classmethod
    def log_record_to_dict(cls, record) -> dict[str, Any]:
        log_dict = super().log_record_to_dict(record)
        # ... add more stuff to log_dict here ...
        return log_dict
```

When we updated to `newrelic==9.12`, we started getting errors because `log_record_to_dict` "takes 2 positional arguments but 3 were given".

We already fixed this from our side, by taking in `*args, **kwargs` and redirecting them to the super call, but wondering if we are the only ones affected...

PR #1172, opened 2024-06-27
Fix Environment Scraping Logic

TimPansino:
# Overview

* Fix the following error where the name `version` could be unbound.

```
  File "/__w/newrelic-python-agent/newrelic-python-agent/.tox/python-agent_unittests-py27-with_extensions/lib/python2.7/site-packages/newrelic/core/environment.py", line 246, in environment_settings
    if version or nr_hook:
UnboundLocalError: local variable 'version' referenced before assignment
```

### DIFF
--- a/newrelic/core/environment.py
+++ b/newrelic/core/environment.py
@@ -235,15 +235,16 @@ def environment_settings():
                 continue
 
             # Don't attempt to look up version information for our hooks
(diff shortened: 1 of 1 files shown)


hmstepanek (approved):
Ouch. Looks good.

PR #1183, opened 2024-07-24
Capture AWS SQS info on message spans

hmstepanek:
# Overview
Capture AWS SQS agent attributes on message spans.

### DIFF
--- a/newrelic/api/message_trace.py
+++ b/newrelic/api/message_trace.py
@@ -92,15 +92,7 @@ class MessageTrace(CatHeaderMixin, TimeTrace):
 
 
 def MessageTraceWrapper(
-    wrapped,
-    library,
-    operation,
-    destination_type,
-    destination_name,
-    params={},
-    terminal=True,
-    async_wrapper=None,
-    extract_agent_attrs=None,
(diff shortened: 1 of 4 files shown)


newrelic/api/message_trace.py
```diff
@@ -145,7 +175,16 @@ def _nr_message_trace_wrapper_(wrapped, instance, args, kwargs):
...
-def message_trace(library, operation, destination_type, destination_name, params={}, terminal=True, async_wrapper=None):
+def message_trace(
+    library,
+    operation,
+    destination_type,
+    destination_name,
+    params={},
+    terminal=True,
+    async_wrapper=None,
+    extract_agent_attrs=None,
```
TimPansino, line 186: I don't particularly care for muddying the public APIs with this. Can we instead define an SQS wrapper in the botocore instrumentation that starts a message trace, and adds the necessary attrs?

hmstepanek: I've recreated a simpler wrapper inside the botocore instrumentation that contains the call to create the agent attributes.

TimPansino (approved):

PR #1189, opened 2024-08-05
Change to using auto bind on openai base client

hmstepanek:
# Overview
Openai added a new argument which broke our openai instrumentation. In particular it broke our base client arg binding. Change to using auto detection of function signature using bind args.

### DIFF
--- a/newrelic/hooks/mlmodel_openai.py
+++ b/newrelic/hooks/mlmodel_openai.py
@@ -683,17 +683,6 @@ def wrap_convert_to_openai_object(wrapped, instance, args, kwargs):
     return returned_response
 
 
-def bind_base_client_process_response(
-    cast_to,
-    options,
-    response,
-    stream,
-    stream_cls,
-):
-    nr_response_headers = getattr(response, "headers", None) or {}
-    return nr_response_headers
-
-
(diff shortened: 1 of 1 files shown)


newrelic/hooks/mlmodel_openai.py
```diff
@@ -704,7 +693,9 @@ def wrap_base_client_process_response_sync(wrapped, instance, args, kwargs):
     if not settings.ai_monitoring.enabled:
         return wrapped(*args, **kwargs)
 
-    nr_response_headers = bind_base_client_process_response(*args, **kwargs)
+    bound_args = bind_args(wrapped, args, kwargs)
+    nr_response_headers = getattr(bound_args["response"], "headers", None) or {}
```
umaannamalai, line 697: Do we want to use a `.get()` here for the `response` key?

hmstepanek: The only time response wouldn't exist is if the function signature changed in which case I think it would probably be better to fail fast and alert us to that change rather than silently passing the issue along to later code.

umaannamalai (approved):

nimdraugsael:
@hmstepanek thank you so much for fixing that!
that also broke our production when we tried to bump openai-client and I was so happy to find that you fixed it already. Kudos!

PR #1173, opened 2024-06-28
Adds instrumentation for aiomcache

jairhenrique:
_Before contributing, please read our [contributing guidelines](https://github.com/newrelic/newrelic-python-agent/blob/main/CONTRIBUTING.rst) and [code of conduct](https://github.com/newrelic/.github/blob/master/CODE_OF_CONDUCT.md)._

# Overview
Adds support to [aiomcache](https://github.com/aio-libs/aiomcache/) instrumentation.

# Related Github Issue
Include a link to the related GitHub issue, if applicable

# Testing
The agent includes a suite of tests which should be used to
verify your changes don't break existing functionality. These tests will run with
Github Actions when a pull request is made. More details on running the tests locally can be found
[here](https://github.com/newrelic/newrelic-python-agent/blob/main/CONTRIBUTING.rst#testing-guidelines),
For most contributions it is strongly recommended to add additional tests which
exercise your changes.

PR #1194, opened 2024-08-15
Capture memcache host & port

hmstepanek:
# Overview
Capture host & port to create AWS entity metric on bmemcached, aiomcache, and pymemcache libraries.
Do not capture host and port info for no longer supported libraries umemcached (doesn't support Python3).

### DIFF
--- a/newrelic/hooks/datastore_aiomcache.py
+++ b/newrelic/hooks/datastore_aiomcache.py
@@ -18,7 +18,25 @@ _memcache_client_methods = (
 )
 
 
+def capture_host(self, *args, **kwargs):
+    if hasattr(self, "_pool") and hasattr(self._pool, "_host"):
+        return self._pool._host
+
+
+def capture_port(self, *args, **kwargs):
+    if hasattr(self, "_pool") and hasattr(self._pool, "_port"):
+        return self._pool._port
+
+
 def instrument_aiomcache_client(module):
     for name in _memcache_client_methods:
         if hasattr(module.Client, name):
-            wrap_datastore_trace(module, "Client.%s" % name, product="Memcached", target=None, operation=name)
+            wrap_datastore_trace(
+                module,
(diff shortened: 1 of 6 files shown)


jairhenrique:
It would be nice to add this information to the `aiomcache` database store as well.

hmstepanek:
> It would be nice to add this information to the `aiomcache` database store as well.

I have chosen not to do this since aiomcache has not been updated since 2014. It appears to be dead. It also does not support past Python 3.4 (we only officially support down to Python 3.7 right now). If it is updated at some point in the future then we can do this at that time. Right now I don't see that's it's worth the effort.

jairhenrique:
> I have chosen not to do this since aiomcache has not been updated since 2014. It appears to be dead. It also does not support past Python 3.4 (we only officially support down to Python 3.7 right now). If it is updated at some point in the future then we can do this at that time. Right now I don't see that it's worth the effort.

Maybe we are not talking about the same project. The latest version of aiomcache was released in May 2024. https://github.com/aio-libs/aiomcache/releases/tag/v0.8.2

I added the instrumentation in this [pr](https://github.com/newrelic/newrelic-python-agent/pull/1173).

hmstepanek:
> > I have chosen not to do this since aiomcache has not been updated since 2014. It appears to be dead. It also does not support past Python 3.4 (we only officially support down to Python 3.7 right now). If it is updated at some point in the future then we can do this at that time. Right now I don't see that it's worth the effort.
> 
> Maybe we are not talking about the same project. The latest version of aiomcache was released in May 2024. https://github.com/aio-libs/aiomcache/releases/tag/v0.8.2
> 
> I added the instrumentation in this [pr](https://github.com/newrelic/newrelic-python-agent/pull/1173).

You're right! We were talking about different libraries! Sorry-yes I will add that in!

TimPansino (approved):

PR #1200, opened 2024-08-22
Convert String Formatting to f-strings with Flynt

TimPansino:
# Overview

* Run [flynt](https://pypi.org/project/flynt/) linter to replace %, .format, and .join style string formatting with f-strings.
* Additional manual conversions done as follow up commits.

Note: There still exist quite a few % style string interpolations. They're used for the following:
* Lazy logging string interpolation for performance
* Byte string interpolation, only used in a few tests
* Very few select spots, particularly where constant templates are later interpolated with variable data.
  * Can be searched for loosely with this regex search in vscode or grep `['"]\s*(%|\.format)`

# Testing

Final flynt command used, which can be run for validation:

```bash
flynt -ll 999 -a -tc -tj ./ --exclude newrelic/packages/
```


tests/testing_support/validators/validate_transaction_errors.py
```diff
@@ -58,20 +58,16 @@ def _validate_transaction_errors(wrapped, instance, args, kwargs):
         else:
             compare_to = sorted([e.type for e in captured])
 
-        assert expected == compare_to, "expected=%r, captured=%r, errors=%r" % (expected, compare_to, captured)
+        assert expected == compare_to, f"expected={expected!r}, captured={compare_to!r}, errors={captured!r}"
```
lrafeei, line 61: I think this may have been unintentional (the `!r` after the variable name)

tests/testing_support/validators/validate_transaction_errors.py
```diff
@@ -58,20 +58,16 @@ def _validate_transaction_errors(wrapped, instance, args, kwargs):
...
             assert e.span_id
             for name, value in required_params:
-                assert name in e.custom_params, "name=%r, params=%r" % (name, e.custom_params)
-                assert e.custom_params[name] == value, "name=%r, value=%r, params=%r" % (
-                    name,
-                    value,
-                    e.custom_params,
-                )
+                assert name in e.custom_params, f"name={name!r}, params={e.custom_params!r}"
+                assert e.custom_params[name] == value, f"name={name!r}, value={value!r}, params={e.custom_params!r}"
```
lrafeei, line 67: Same `!r` issue as before

tests/testing_support/validators/validate_transaction_errors.py
```diff
@@ -58,20 +58,16 @@ def _validate_transaction_errors(wrapped, instance, args, kwargs):
...
 
             for name, value in forgone_params:
-                assert name not in e.custom_params, "name=%r, params=%r" % (name, e.custom_params)
+                assert name not in e.custom_params, f"name={name!r}, params={e.custom_params!r}"
```
lrafeei, line 70: Another `!r` instance

tests/testing_support/validators/validate_tt_parameters.py
```diff
@@ -37,15 +37,11 @@ def _validate_tt_parameters(wrapped, instance, args, kwargs):
...
 
         for name in required_params:
-            assert name in tt_intrinsics, "name=%r, intrinsics=%r" % (name, tt_intrinsics)
-            assert tt_intrinsics[name] == required_params[name], "name=%r, value=%r, intrinsics=%r" % (
-                name,
-                required_params[name],
-                tt_intrinsics,
-            )
+            assert name in tt_intrinsics, f"name={name!r}, intrinsics={tt_intrinsics!r}"
+            assert tt_intrinsics[name] == required_params[name], f"name={name!r}, value={required_params[name]!r}, intrinsics={tt_intrinsics!r}"
```
lrafeei, line 41: Another `!r` issue

tests/testing_support/validators/validate_tt_parameters.py
```diff
@@ -37,15 +37,11 @@ def _validate_tt_parameters(wrapped, instance, args, kwargs):
...
 
         for name in forgone_params:
-            assert name not in tt_intrinsics, "name=%r, intrinsics=%r" % (name, tt_intrinsics)
+            assert name not in tt_intrinsics, f"name={name!r}, intrinsics={tt_intrinsics!r}"
```
lrafeei, line 44: Same `!r` issue as before

tests/testing_support/validators/validate_transaction_metrics.py
```diff
@@ -100,17 +100,17 @@ def _validate(metrics, name, scope, count):
...
-                            out.append("{0}: {1}".format((metric_key, metric_tags), metric_value[0]))
+                            out.append(f"{metric_key, metric_tags}: {metric_value[0]}")
                     else:
-                        out.append("{0}: {1}".format(metric_key, metric_container[0]))
+                        out.append(f"{metric_key}: {metric_container[0]}")
                 return "\n".join(out)
 
             def _metric_details():
-                return "metric=%r, count=%r" % (key, metric.call_count)
+                return f"metric={key!r}, count={metric.call_count!r}"
```
lrafeei, line 113: `!r` issue

lrafeei (changes requested):
I noticed that there were some instances where `%r` did not convert the string into an f string without attaching `!r` to the end of the variable.  I did not actually include every instance of this occurring, but it seems that the instances where this has occurred can be found with a simple search for `!r`

tests/testing_support/validators/validate_transaction_errors.py
TimPansino: That's actually correct, the !r is the same raw string conversion that %r was doing. I considered changing it to drop the !r but decided to leave the behavior exactly the same for such a sweeping set of changes.

lrafeei (approved):

PR #1205, opened 2024-09-03
Remove frameworks that only work in Python 2.7

lrafeei:
There have been frameworks that have only supported Python 2.7.  While these frameworks are no longer supported by New Relic (by virtue of having a last release of 3+ years ago) the convention has been to keep the instrumentation in the code and just remove the test runner.  However, with a lack of Python 2.7 support, these frameworks will not work and will need to be removed.

Frameworks removed:
- umemcache
- oursql
- pywapi
- pylons
- web2py
- weberror


TimPansino (approved):
You might want to rebase and delete those commits from me

PR #1211, opened 2024-09-16
Fix AI test failures & do not run graphqlmaster tests

hmstepanek:
# Overview
Fix the following AI test failures:
* Bedrock changed the order of payload assertions in commit: https://github.com/boto/botocore/commit/db41663ea2fa84cb77d548934c92d3d5c26c2c05 which resulted in the expected error being different due to the validator order change.
* Langchain added metadata to Document interface so update expected events with this field.
* pydantic ValidationError moved locations so adjust the path to the error in the tests.
* Langchain doesn't list support for Python3.8 so do not test it on Python3.8.
* Graphql main tests are failing so comment these out for now.


umaannamalai (approved):
Approving - independent of LangChain test failures

hmstepanek:
@umaannamalai Would you mind giving this another review? I've added a couple additional fixes for langchain and graphql tests.

umaannamalai (approved):

PR #1214, opened 2024-09-20
LLM Custom Attributes Context Manager API

umaannamalai:
Add new WithLlmCustomAttributes context manager API


newrelic/api/llm_custom_attributes.py
```diff
@@ -0,0 +1,56 @@
...
+from newrelic.core.function_node import FunctionNode
+
+_logger = logging.getLogger(__name__)
+
+
+class WithLlmCustomAttributes(object):
+    def __init__(self, custom_attr_dict):
+        transaction = current_transaction()
+        if not isinstance(custom_attr_dict, dict) or custom_attr_dict is None:
+            raise TypeError("custom_attr_dict must be a dictionary. Received type: %s" % type(custom_attr_dict))
```
hmstepanek, line 32: The second half of the if statement (`or custom_attr_dict is None`) is not reachable since None is not of typed dict. I wonder do we want to instead be checking for it being empty?
```suggestion
            raise TypeError("custom_attr_dict must be a non empty dictionary. Received type: %s" % type(custom_attr_dict))
```

newrelic/api/llm_custom_attributes.py
```diff
@@ -0,0 +1,56 @@
...
+from newrelic.common.object_wrapper import FunctionWrapper, wrap_object
+from newrelic.core.function_node import FunctionNode
+
+_logger = logging.getLogger(__name__)
+
+
+class WithLlmCustomAttributes(object):
+    def __init__(self, custom_attr_dict):
+        transaction = current_transaction()
+        if not isinstance(custom_attr_dict, dict) or custom_attr_dict is None:
```
hmstepanek, line 31:
```suggestion
        if not custom_attr_dict or not isinstance(custom_attr_dict, dict):
```

newrelic/api/llm_custom_attributes.py
```diff
@@ -0,0 +1,56 @@
...
+            raise TypeError("custom_attr_dict must be a dictionary. Received type: %s" % type(custom_attr_dict))
+
+        # Add "llm." prefix to all keys in attribute dictionary
+        prefixed_attr_dict = {}
+        for k, v in custom_attr_dict.items():
+            if not k.startswith("llm."):
+                _logger.warning("Invalid attribute name %s. Renamed to llm.%s." % (k, k))
```
hmstepanek, line 38: I wonder if we should be logging a warning at all here. It might get a bit cumbersome logging-wise and I think it's totally fine if the customer wants to let us convert the names to be `llm.` prefixed for them. We can just document it as part of the api.

newrelic/api/llm_custom_attributes.py
```diff
@@ -0,0 +1,56 @@
...
+
+class WithLlmCustomAttributes(object):
+    def __init__(self, custom_attr_dict):
+        transaction = current_transaction()
+        if not isinstance(custom_attr_dict, dict) or custom_attr_dict is None:
+            raise TypeError("custom_attr_dict must be a dictionary. Received type: %s" % type(custom_attr_dict))
+
+        # Add "llm." prefix to all keys in attribute dictionary
+        prefixed_attr_dict = {}
+        for k, v in custom_attr_dict.items():
```
hmstepanek, line 36: It would be faster to use a dictionary comprehension on this one:
```suggestion
        {k if k.startswith("llm.") else f"llm.{k}": v for k, v in custom_attr_dict.items()}
```

newrelic/api/llm_custom_attributes.py
```diff
@@ -0,0 +1,56 @@
...
+            if not k.startswith("llm."):
+                _logger.warning("Invalid attribute name %s. Renamed to llm.%s." % (k, k))
+                prefixed_attr_dict["llm." + k] = v
+
+        context_attrs = prefixed_attr_dict if prefixed_attr_dict else custom_attr_dict
```
hmstepanek, line 41: Is there a bug here? What happens when there are mixed `llm.` prefixed and non-prefixed keys? I believe this would discard the already prefixed keys if that were the case.

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -698,7 +698,7 @@ def _get_run_manager_info(transaction, run_args, instance, completion_id):
     # metadata and tags are keys in the config parameter.
     metadata = {}
     metadata.update((run_args.get("config") or {}).get("metadata") or {})
-    # Do not report intenral nr_completion_id in metadata.
+    # Do not report internal nr_completion_id in metadata.
```
hmstepanek, line 701: lol oops.

newrelic/hooks/mlmodel_openai.py
```diff
@@ -780,15 +789,15 @@ def _record_events_on_stop_iteration(self, transaction):
         self._nr_ft.__exit__(None, None, None)
         try:
             openai_attrs = getattr(self, "_nr_openai_attrs", {})
-
             # If there are no openai attrs exit early as there's no data to record.
             if not openai_attrs:
                 return
 
             completion_id = str(uuid.uuid4())
+            llm_context_attrs = getattr(self, "nr_llm_context_attrs", None)
```
hmstepanek, line 797: Given that we missed this one-we are probably missing a test for this case too:
```suggestion
            llm_context_attrs = getattr(self, "_nr_llm_context_attrs", None)
```

newrelic/hooks/mlmodel_openai.py
```diff
@@ -475,12 +479,12 @@ def _handle_completion_success(transaction, linking_metadata, completion_id, kwa
                 # openai._legacy_response.LegacyAPIResponse
                 response = json.loads(response.http_response.text.strip())
 
-        _record_completion_success(transaction, linking_metadata, completion_id, kwargs, ft, response_headers, response)
+        _record_completion_success(transaction, linking_metadata, completion_id, kwargs, ft, response_headers, None, response)
```
hmstepanek, line 482: Why are a bunch of these calls passing None for the attrs?

hmstepanek (changes requested):
I'm going to let you fix the implementation issues first and then I'll take a look at the tests. It looks like there's just some small but large impact issues with the implementation and I think once we fix those the tests will probably be in better shape too.

newrelic/api/llm_custom_attributes.py
umaannamalai: I tend to agree with this. I originally added this logging statement to match the Node implementation but I would be fine to take it out.

hmstepanek (changes requested):
I think we should just use our regular approach of attaching the attrs to the transaction without using contextvars. I don't see that it's needed in this case and I think simpler is better here. Other than that, I just had one minor suggestion for one of the tests.

hmstepanek (dismissed):

newrelic/api/llm_custom_attributes.py
```diff
@@ -0,0 +1,47 @@
...
+            _logger.warning("WithLlmCustomAttributes must be called within the scope of a transaction.")
+            return self
+
+        self.transaction._llm_context_attrs = self.attr_dict
+        return self
+
+    def __exit__(self, exc, value, tb):
+        # Clear out context attributes once we leave the current context
+        if self.transaction:
+            self.transaction._llm_context_attrs = None
```
hmstepanek, line 47:
```suggestion
            del(self.transaction._llm_context_attrs)
```

tests/agent_features/test_llm_custom_attributes.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@pytest.mark.parametrize("context_attrs", (None, "not-a-dict"))
+@background_task()
+def test_llm_custom_attributes_no_attrs(context_attrs):
+    transaction = current_transaction()
+
+    with pytest.raises(TypeError):
+        with WithLlmCustomAttributes(context_attrs):
+            assert transaction._llm_context_attrs is None
```
hmstepanek, line 36: I believe it's failing before this assertion so this assertion is unreachable but either way this assertion isn't correct anymore. Maybe move this outside the with statements and just assert that the attribute isn't on the transaction?
```suggestion
    assert not hasattr(transaction, "_llm_context_attrs")
```

tests/agent_features/test_llm_custom_attributes.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@background_task()
+def test_llm_custom_attributes_prefixed_attrs():
+    transaction = current_transaction()
+    with WithLlmCustomAttributes({"llm.test": "attr", "test1": "attr1"}):
+        # Validate API does not prefix attributes that already begin with "llm."
+        assert transaction._llm_context_attrs == {"llm.test": "attr", "llm.test1": "attr1"}
```
hmstepanek, line 44: Same thing here (just make sure we are cleaning up afterwards):
```suggestion
        assert transaction._llm_context_attrs == {"llm.test": "attr", "llm.test1": "attr1"}
    assert not hasattr(transaction, "_llm_context_attrs")
```

hmstepanek (dismissed):
Just a couple small suggestions about the tests-otherwise looks good!

hmstepanek (approved):

PR #1221, opened 2024-09-25
Add Python Version Specific Installation Errors

TimPansino:
# Overview

* Add python version specific errors that recommend past agent versions when installing on an unsupported Python version.

# Examples

`RuntimeError: The New Relic Python agent only supports Python 3.7+. We recommend upgrading to a newer version of Python. The last agent version to support Python 2.7 was v9.13.0.`

`RuntimeError: The New Relic Python agent only supports Python 3.7+. We recommend upgrading to a newer version of Python. The last agent version to support Python 3.6 was v7.16.0.178.`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,30 @@ import sys
 
 python_version = sys.version_info[:2]
 
-assert python_version >= (3, 7), "The New Relic Python agent only supports Python 3.7+."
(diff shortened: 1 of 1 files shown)


umaannamalai (approved):
🚀

PR #1226, opened 2024-10-02
Python 3.13 Support

TimPansino:
# Overview

* Add wheels and classifiers for Python 3.13.
  * Update cibuildwheel to use a version with Python 3.13 support.
* Add support for setuptools_scm v7 & v8 for Python 3.13 compatibility.
* Add Python 3.13 to all relevant tox matrices.
  * Add annotated and commented out environments where Python 3.13 is not yet supported by the library. Can be uncommented again to test in the next few weeks as more wheels become available, and should eventually be slotted into the existing matrix line.
* Update C Extensions to work with Python 3.13, and remove Python 2 macros.
* Fix broken tests for agent_unittests in developer mode.
* Fix broken metric across all memcache tests expecting too strict a localhost hostname.

### DIFF
--- a/tests/agent_unittests/test_full_uri_payloads.py
+++ b/tests/agent_unittests/test_full_uri_payloads.py
@@ -22,8 +22,8 @@ from newrelic.core.agent_protocol import AgentProtocol
 from newrelic.common.agent_http import HttpClient
 from newrelic.core.config import global_settings, _environ_as_bool
 
-SKIP_IF_NO_LICENSE_KEY = pytest.mark.skipif("NEW_RELIC_LICENSE_KEY" not in os.environ, reason="License key is not expected to be valid")
-SKIP_IF_DEVELOPER_MODE = pytest.mark.skipif(_environ_as_bool("NEW_RELIC_DEVELOPER_MODE", False), reason="Cannot connect to collector in developer mode")
+DEVELOPER_MODE = _environ_as_bool("NEW_RELIC_DEVELOPER_MODE", False) or "NEW_RELIC_LICENSE_KEY" not in os.environ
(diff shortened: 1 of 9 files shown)


hmstepanek (approved):
Looks good. Just a couple feedback comments.

tox.ini
```diff
@@ -231,13 +252,12 @@ deps =
     component_tastypie-tastypielatest: django-tastypie
     component_tastypie-tastypielatest: django<4.1
     component_tastypie-tastypielatest: asgiref<3.7.1  # asgiref==3.7.1 only suppport Python 3.10+
-    coroutines_asyncio-{py37,py38,py39,py310,py311,py312}: uvloop
+    coroutines_asyncio-{py37,py38,py39,py310,py311,py312,py313}: uvloop
```
umaannamalai, line 255: Was `py313` meant to be added here? This isn't being tested on 3.13 above due to the `uvloop` dependency.

tests/agent_unittests/test_full_uri_payloads.py
```diff
@@ -20,7 +20,10 @@
 from testing_support.fixtures import collector_agent_registration_fixture
 from newrelic.core.agent_protocol import AgentProtocol
 from newrelic.common.agent_http import HttpClient
-from newrelic.core.config import global_settings
+from newrelic.core.config import global_settings, _environ_as_bool
+
+SKIP_IF_NO_LICENSE_KEY = pytest.mark.skipif("NEW_RELIC_LICENSE_KEY" not in os.environ, reason="License key is not expected to be valid")
```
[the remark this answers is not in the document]
TimPansino: That's an ancient comment, it doesn't particularly make a lot of sense. I'm not sure who originally wrote it.

tox.ini
TimPansino: It was added for when we enable the actual tox environment in the future.

PR #1227, opened 2024-10-07
Add attrs for aws lambda entity

hmstepanek:
# Overview
Add attrs for aws lambda entity linking.
>When an application invokes a Lambda, linking between the application and AWS Lambdas is done using the following span attributes:

cloud.platform: the string aws_lambda
cloud.resource_id: the ARN for the Lambda, or an alias or version, if it can be determined.
Lambda ARNs follow the format:

arn:${Partition}:lambda:${Region}:${Account}:function:${FunctionName}
arn:${Partition}:lambda:${Region}:${Account}:function:${FunctionName}:${Alias}
arn:${Partition}:lambda:${Region}:${Account}:function:${FunctionName}:${Version}


newrelic/hooks/external_botocore.py
```diff
@@ -889,7 +891,24 @@ def _nr_sqs_message_trace_wrapper_(wrapped, instance, args, kwargs):
...
 
+def wrap_lambda_invoke(wrapped):
+    def _nr_wrap_lambda_invoke(wrapped, instance, args, kwargs):
+        transaction = current_transaction()
+        if not transaction:
+            return wrapped(*args, **kwargs)
+
+        bound_args = bind_args(wrapped, args, kwargs)
+        arn = bound_args["kwargs"].get("FunctionName")
+        if arn and hasattr(instance, "meta") and hasattr(instance.meta, "events"):
```
hmstepanek, line 902: This is just an object that I found that is also available inside the make_request. It's possible there is something better to attach this to but it was hard to find something that was accessible in both spots. I didn't want to attach it to the transaction as I thought there might be a conflict with that in async cases.

newrelic/hooks/external_botocore.py
```diff
@@ -951,6 +970,10 @@ def _nr_endpoint_make_request_(wrapped, instance, args, kwargs):
     with ExternalTrace(library="botocore", url=url, method=method, source=wrapped) as trace:
         try:
             trace._add_agent_attribute("aws.operation", operation_model.name)
+            lambda_arn = getattr(instance._event_emitter, "_nr_lambda_arn", None)
+            if lambda_arn:
+                trace._add_agent_attribute("cloud.platform", "aws_lambda")
+                trace._add_agent_attribute("cloud.resource_id", lambda_arn)
```
TimPansino, line 976: Are we confident this object isn't reused for future requests?

tests/external_botocore/test_boto3_lambda.py
```diff
@@ -0,0 +1,123 @@
...
+            region_name=AWS_REGION_NAME,
+        )
+        # Create IAM role
+        return iam.create_role(
+            RoleName="my-role",
+            AssumeRolePolicyDocument="some policy",
+            Path="/my-path/",
+        )[
+            "Role"
+        ]["Arn"]
```
TimPansino, line 121:
```suggestion
       role = iam.create_role(
            RoleName="my-role",
            AssumeRolePolicyDocument="some policy",
            Path="/my-path/",
        )
        return role["Role"]["Arn"]
```

Maybe do this to make the formatter less insufferable

tox.ini
```diff
@@ -303,6 +303,7 @@ deps =
     external_botocore-botocore128: botocore<1.29
     external_botocore-botocore0125: botocore<1.26
     external_botocore: moto
+    external_botocore: docker
```
TimPansino, line 306: What's the docker dependency for? Lambda/moto?

hmstepanek: Yes it's import inside the lambda invoke in moto.

newrelic/hooks/external_botocore.py
hmstepanek: I've updated this to use a different object that is generated for each request.

newrelic/hooks/external_botocore.py
```diff
@@ -1001,6 +1002,38 @@ def _nr_sqs_message_trace_wrapper_(wrapped, instance, args, kwargs):
...
+    if not transaction:
+        return wrapped(*args, **kwargs)
+
+    bound_args = bind_args(wrapped, args, kwargs)
+
+    api_params = wrapped(*args, **kwargs)
+
+    arn = bound_args.get("api_params").get("FunctionName")
+    if arn and hasattr(arn, "startswith") and arn.startswith("arn:"):
+        api_params["_nr_arn"] = arn
```
TimPansino, line 1016: Is it possible to use a bytes string for the name? I see what you're trying to do with the `hasattr(arn, "startswith")` but that line would still crash if there's a bytes string since you can't call startswith on that and pass a string. Would it be better to just try/except this clause?

TimPansino (dismissed):
Pre-approving because this looks fine either way

TimPansino (approved):

PR #1228, opened 2024-10-08
Package capture harvest

lrafeei:
This PR aims to alleviate some of the time consumed by package data collection in the `connect` cycle.  Instead of collecting all modules used in the application at startup, the agent uses the fast harvest cycle to incrementally collect this data and sends it though the `update_loaded_modules` endpoint instead of the `connect` endpoint (through the "environment" key).

Note: despite the term "Jars" not being used by Python to refer to packages/modules, this is the exact string the endpoint needs in order to process the data.


newrelic/core/application.py
```diff
@@ -1671,6 +1700,18 @@ def internal_agent_shutdown(self, restart=False):
 
         self.stop_data_samplers()
 
+        # Finishes collecting environment plugin information
+        # if this has not been completed during harvest
+        # lifetime of the application
+
+        while self.configuration.package_reporting.enabled and self._env_send:
```
hmstepanek, line 1707: Just trying to match the logic above:
```suggestion
        while self.configuration and self.configuration.package_reporting.enabled and self._env_send:
```

newrelic/core/application.py
```diff
@@ -1238,6 +1245,28 @@ def harvest(self, shutdown=False, flexible=False):
...
+                        configuration
+                        and configuration.package_reporting.enabled
+                        and self._env_send
+                        and ((time.time() - stopwatch_start) < 2.0)
+                    ):
+                        try:
+                            module_info = next(self.plugins)
+                            self.modules.append(module_info)
+                        except StopIteration:
+                            self._env_send = False
```
hmstepanek, line 1261: Can we name this something more explicit to what it's actually used for? Instead of `_env_send` perhaps `_plugins_to_iterate` or `_remaining_plugins`.

newrelic/core/application.py
```diff
@@ -1671,6 +1700,18 @@ def internal_agent_shutdown(self, restart=False):
...
+        # if this has not been completed during harvest
+        # lifetime of the application
+
+        while self.configuration.package_reporting.enabled and self._env_send:
+            try:
+                module_info = next(self.plugins)
+                self.modules.append(module_info)
+            except StopIteration:
+                self._env_send = False
+                self._active_session.send_loaded_modules(self.modules)
```
hmstepanek, line 1713: It's possible there are no modules to send in this case as well. Just in case, clear the modules after sending.
```suggestion
                if self.modules:
                   
self._active_session.send_loaded_modules(self.modules)
self.modules = []
```

newrelic/core/environment.py
```diff
@@ -198,11 +198,13 @@ def environment_settings():
...
+
+
+def plugins():
     # Module information.
     stdlib_builtin_module_names = _get_stdlib_builtin_module_names()
 
-    plugins = []
-
     settings = global_settings()
     if settings and settings.package_reporting.enabled:
```
hmstepanek, line 209: Since we already check for this in the harvester we should probably remove it from here as it's not needed.

hmstepanek (changes requested):
A couple minor things. Overall looks good!

newrelic/core/application.py
```diff
@@ -1238,6 +1245,28 @@ def harvest(self, shutdown=False, flexible=False):
...
                             )
 
+                    # Send environment plugin list
+
+                    stopwatch_start = time.time()
+                    while (
+                        configuration
+                        and configuration.package_reporting.enabled
+                        and self._env_send
+                        and ((time.time() - stopwatch_start) < 2.0)
```
hmstepanek, line 1255: I'm not sure if I mentioned this already or not but some agents have a configurable rate limit for this instead of a hardcoded value. That being said could we place 2.0 at the top of the file as a constant so it's not a magic number here? Something like:
```
MAX_PACKAGE_CAPTURE_TIME_PER_SLOW_HARVEST = 2.0
```

newrelic/core/application.py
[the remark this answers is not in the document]
lrafeei: I don't disagree with that.  In this case, I thought I would have to put this before any agent shutdown commands to ensure that it gets through.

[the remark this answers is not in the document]
hmstepanek: >I don't disagree with that. In this case, I thought I would have to put this before any agent shutdown commands to ensure that it gets through.

Yes I think you are correct there! I was kinda thinking allowed here-not saying what you have here needs to be changed in any way. Just more raising a possible concern but I think this is ok as is for now. Worse case scenario we don't shutdown nicely and get killed in this package capture loop but that's probably ok/won't hurt anything.

hmstepanek (approved):

newrelic/core/application.py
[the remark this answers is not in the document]
lrafeei: Just making sure I didn't miss anything!

hmstepanek (dismissed):

newrelic/core/application.py
```diff
@@ -26,12 +26,14 @@
 from functools import partial
 
 from newrelic.common.object_names import callable_name
+
+# from newrelic.common.encoding_utils import json_encode
```
TimPansino, line 30: Why is this commented out here?

TimPansino (changes requested):
Can we either remove or uncomment that import?

newrelic/core/application.py
lrafeei: Whoops, I meant to remove that.

hmstepanek (approved):

TimPansino (approved):

Issue #1229, opened 2024-10-08
Support reading configuration from pyproject.toml


Tatsh:
### Is your feature request related to a problem? Please describe.
No.

### Feature Description
Please read configuration from the shared configuration file `pyproject.toml` as well as `newrelic.ini` (one or the other, preferring `pyproject.toml`). Example mapping based on current `newrelic.ini` format:

```toml
[tool.newrelic]
log_level = "info"
high_security = false
monitor_mode = true
app_name = "app_name"

# Extends above
[tool.newrelic.production]
app_name = "app_name (Production)"

# [import-hook:django]
# instrumentation.scripts.django_admin = process_tasks runjob runjobs
# becomes:

[tool.newrelic.import-hook.django.instrumentation.scripts] # replace : with -
django_admin = ["process_tasks", "runjob", "runjobs"]
```

This will help to reduce the amount of project metadata files needed.

### Describe Alternatives
n/a

### Additional context

[PEP 621 – Storing project metadata in pyproject.toml](https://peps.python.org/pep-0621/)

### Priority
Really Want.

PR #1238, opened 2024-10-15
Add TOML configuration file support

Tatsh:
# Overview
This adds support for reading configuration from a TOML file such as `pyproject.toml` to reduce clutter.

Layout is:

```toml
[tool.newrelic]
app_name = "app name"
```

Keys with . in them are expanded to be subkeys:

```toml
[tool.newrelic.browser_monitoring]
enabled = true
```

Anything that is a list in the layout is made into a string joined with a single space.

`import-hook:...` is translated as such and requires quotes in the keys:

```toml
[tool.newrelic.import-hook.module_name]
"sub.sub" = ["item1", "item2"]
```

`error_collector.ignore_errors`:

```toml
[tool.newrelic.error_collector]
ignore_errors = ["module.name:ClassName", "module.name:ClassName2"]
```

Environments use the `tool.newrelic.env` namespace:

```toml
[tool.newrelic.env.production]
app_name = "production name"
```

Expansion of the subkeys includes doing so in an environment section:

```toml
[tool.newrelic.env.production.transaction_tracer]
enabled = false
```

# Related Github Issue

#1229 

# Testing

Tests have been added.


newrelic/config.py
```diff
@@ -911,6 +911,29 @@ def apply_local_high_security_mode_setting(settings):
...
+            fixed_value = " ".join(value) if isinstance(value, list) else value
+            path_split = _path.split(".")
+            # Handle environments
+            if _path.startswith("env."):
+                env_key = f"newrelic:{path_split[1]}"
+                fixed_key = ".".join((*path_split[2:], key))
+                top[env_key] = {**top.get(env_key, {}), fixed_key: fixed_value}
+            # Handle import-hook:... configuration
+            elif _path.startswith("import-hook."):
+                import_hook_key = f"import-hook:{path_split[1]}"
```
TimPansino, line 930:
```suggestion
                import_hook_key = f"import-hook:{'.'.join(path_split[1:])}"
```

This is missing the remainder of the path. This correction should fix that.

The following config gets parsed incorrectly due to that.

```
[tool.newrelic.import-hook.flask.app]
enabled = false
```

This gets parsed as `import-hook.flask.enabled = false`, and is missing the `.app`.

Tatsh: Fixed with latest push.

TimPansino (dismissed):
Approving from my end, waiting on @hmstepanek to approve as well before merging.

tests/agent_features/test_configuration.py
```diff
@@ -949,6 +949,69 @@ def test_initialize_developer_mode(section, expect_error, logger):
         assert "CONFIGURATION ERROR" not in logger.caplog.records
 
 
+newrelic_toml_contents = b"""
+[tool.newrelic]
+app_name = "test11"
+monitor_mode = true
```
hmstepanek, line 955: We have various rule configuration options. You can find them [here](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/core/config.py#L726). When I add one of them like the example below. It looks like the parsing fails because it can't handle a list of dictionaries.
```suggestion
monitor_mode = true

url_rules = [{ignore = false, match_expression = ".*\\\.(ace|arj|ini|txt|udl|plist|css|gif|ico|jpe?g|js|png|swf|woff|caf|aiff|m4v|mpe?g|mp3|mp4|mov)$"}]

```

hmstepanek (changes requested):
There are a couple more complicated settings that we have that I don't think are currently compatible with this parser. I went through our core config and looked for some others that are more complex like this. `transaction_segment_terms` is even more complex than the example below that I've noted and looks like:
```
    "transaction_segment_terms": [
      {
        "prefix": "WebTransaction/Custom",
        "terms":  ["one", "two", "three"]
      },
      {
        "prefix": "WebTransaction/Uri",
        "terms": ["seven", "eight", "nine"]
      }
    ]
```

tests/agent_features/test_configuration.py
Tatsh: I added in a fix for this but I was not able to check it in the test. `fetch_config_setting(settings, "url_rules")` always returns `[]`. Grepping for `url_rules` shows:

```
newrelic/common/agent_http.py
587:            "url_rules": [],

newrelic/core/application.py
399:                            "The URL normalization rules for %r are %r.", self._app_name, configuration.url_rules
412:                    self._rules_engine["url"] = RulesEngine(configuration.url_rules)

newrelic/core/config.py
726:_settings.url_rules = []
```

There is no call to `_process_setting()` for url_rules in `core/config.py`.

The following does work to fix the issue but I do not know if this is correct. Plus it might break any existing INI functionality assuming that is working for url_rules.

```diff
diff --git a/newrelic/config.py b/newrelic/config.py
index 73a847ba..11e5268c 100644
--- a/newrelic/config.py
+++ b/newrelic/config.py
@@ -14,6 +14,7 @@
 
 import configparser
 import fnmatch
+import json
 import logging
 import os
 import sys
@@ -309,6 +310,7 @@ def _process_setting(section, option, getter, mapper):
 
 
 def _process_configuration(section):
+    _process_setting(section, "url_rules", "get", json.loads)
     _process_setting(section, "feature_flag", "get", _map_feature_flag)
     _process_setting(section, "app_name", "get", None)
     _process_setting(section, "labels", "get", _map_labels)
diff --git a/tests/agent_features/test_configuration.py b/tests/agent_features/test_configuration.py
index a359a0de..8e83883b 100644
--- a/tests/agent_features/test_configuration.py
+++ b/tests/agent_features/test_configuration.py
@@ -999,6 +999,9 @@ def test_toml_parse_development():
         assert value.enabled is True
         assert value.ignore_classes[0] == "module:name1"
         assert value.ignore_classes[1] == "module:name"
+        value = fetch_config_setting(settings, "url_rules")
+        assert isinstance(value, list)
+        assert value[0]['ignore'] is False
 
 
 def test_toml_parse_production():
 ```

Also, how do I get rid of this error?

```
tests/agent_features/test_configuration.py::test_toml_parse_production ERROR

================================================================================================================================== ERRORS ===================================================================================================================================
_______________________________________________________________________________________________________________ ERROR at setup of test_toml_parse_production ________________________________________________________________________________________________________________

request = <SubRequest 'collector_available_fixture' for <Function test_toml_parse_production>>, collector_agent_registration = <newrelic.api.application.Application object at 0x7f3b7bf023c0>

    @pytest.fixture(scope="function")
    def collector_available_fixture(request, collector_agent_registration):
        application = application_instance()
        active = application.active
>       assert active
E       AssertionError

tests/testing_support/fixtures.py:277: AssertionError
========================================================================================================================== short test summary info ==========================================================================================================================
ERROR tests/agent_features/test_configuration.py::test_toml_parse_production - AssertionError
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
======================================================================================================================= 140 passed, 1 error in 1.24s ========================================================================================================================
```

Comment (TimPansino) [reply tests/agent_features/test_configuration.py:955]: Either set a valid `NEW_RELIC_LICENSE_KEY` environment variable or set `NEW_RELIC_DEVELOPER_MODE=true` to avoid sending any test data to our real backend.

Comment (TimPansino) [reply tests/agent_features/test_configuration.py:955]: @hmstepanek URL rules aren't a setting available in config files, they're a serverside setting that gets sent to us on connect. You can't manually create them in the INI file.

Any serverside only settings shouldn't be a problem for this PR, as we won't attempt to read that information at all from a config file.

File: newrelic/config.py [hunk 3]
```diff
@@ -911,6 +911,36 @@ def apply_local_high_security_mode_setting(settings):
...
+    _path = _path or ""
+    for key, value in d.items():
+        if isinstance(value, dict):
+            _toml_config_to_configparser_dict(value, top, f"{_path}.{key}" if _path else key)
+        else:
+            # When configparser.get() is called in _process_setting() the result will be a string
+            # unless a special getter such as getint() is used. Keep non-lists of strings as JSON
+            # strings so they can be decoded later. Anything in a TOML document should be
+            # JSON-encodable.
+            fixed_value = " ".join(value) if is_list_of_str(value) else json.dumps(value)
```
TimPansino, line 928: I think what you had before was fine, I don't think we need a JSON dump of anything here.

tests/agent_features/test_configuration.py
hmstepanek: Oh I see! You're right-none of these are processed for the config file. Given that, I think this is good as is then!

newrelic/config.py
Tatsh: Okay I restored it.

Tatsh:
> transaction_segment_terms

That also appears to be one that is processed separately like url_rules and not taken from the configuration.

TimPansino:
> > transaction_segment_terms
> 
> That also appears to be one that is processed separately like url_rules and not taken from the configuration.

Yes all 4 of these settings are not possible to configure with a config file.

```
_settings.url_rules = []
_settings.metric_name_rules = []
_settings.transaction_name_rules = []
_settings.transaction_segment_terms = []
```

hmstepanek (dismissed):

TimPansino (dismissed):

TimPansino:
We'll get this cleaned up and merged after we fix the failing tests from main.

TimPansino (approved):

TimPansino (approved):

PR #1249, opened 2024-11-07
Add cloud.aws.account_id config setting

hmstepanek:
# Overview
Add cloud.aws.account_id configuration setting. This is not used in this PR but will be used in future AWS linking for the user to provide the account_id when it cannot be derived by the instrumentation. See the Entity-Relationships agent spec.

### DIFF
--- a/newrelic/core/config.py
+++ b/newrelic/core/config.py
@@ -73,6 +73,21 @@ _logger = logging.getLogger(__name__)
 _logger.addHandler(_NullHandler())
 
 
+def _map_aws_account_id(s, logger):
+    # The AWS account id must be a 12 digit number.
+    # See https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-identifiers.html#awsaccountid.
+    if s and len(s) == 12:
+        if s.isdecimal():
+            account_id = int(s)
+            return account_id
+    # Only log a warning if s is set.
+    if s:
+        logger.warning(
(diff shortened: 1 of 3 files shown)


newrelic/core/config.py
```diff
@@ -73,6 +73,23 @@ def emit(self, record):
...
+            account_id = int(s)
+            return account_id
+        except ValueError:
+            pass
+    # Only log a warning if s is set.
+    if s:
+        _logger.warning(
+            "Improper configuration. cloud.aws.account_id = %s will be ignored because it is not a 12 digit number.", s
+        )
+    return None
```
jairhenrique, line 90:
```suggestion
    if s and len(s) == 12 and s.isdigit():
        return int(s)

    # Only log a warning if s is set.
    if s:
        _logger.warning(
            f"Improper configuration. cloud.aws.account_id = {s} will be ignored because it is not a 12 digit number."
        )

    return None
```

hmstepanek: isdigit looks cool! It looks like it also returns true if there is a power digit like ² which would result in a ValueError still being raised by the int cast though.
```
>>> s="01234²34"
>>> s.isdigit()
True
>>> int(s)
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '01234²34'
```

jairhenrique: @hmstepanek you can use `s.isdecimal()` instead!

TimPansino (approved):

TimPansino (approved):

PR #1250, opened 2024-11-08
NR Control Health Checks

umaannamalai:

# Overview
Describe the changes present in the pull request

# Related Github Issue

# Testing
exercise your changes.


tests/agent_features/test_super_agent_health_check.py
```diff
@@ -0,0 +1,215 @@
...
+            "license_key": "123LICENSEKEY",
+        }
+    )
+    protocol = AgentProtocol(settings, client_cls=HttpClientRecorder)
+
+    with pytest.raises(DiscardDataForRequest):
+        protocol.send("analytic_event_data")
+
+    # Give time for the scheduler to kick in and write to the health file
+    time.sleep(5)
```
TimPansino, line 152: I don't love leaving an intentional race condition in the tests but I don't have a much better solution.

tests/agent_features/test_super_agent_health_check.py
```diff
@@ -0,0 +1,215 @@
...
+
+    application_1.activate_session()
+    application_2.activate_session()
+
+    running_threads = threading.enumerate()
+
+    assert len(running_threads) == 6
+    assert running_threads[1].name == "NR-Control-Health-Main-Thread"
+    assert running_threads[2].name == "NR-Control-Health-Session-Thread"
+    assert running_threads[4].name == "NR-Control-Health-Session-Thread"
```
TimPansino, line 215: Why 6? Can we leave a comment here explaining what's being tested exactly?

tests/agent_unittests/test_agent_protocol.py
```diff
@@ -80,6 +80,11 @@ def send_request(
         request = Request(method=method, path=path, params=params, headers=headers, payload=payload)
         self.SENT.append(request)
         if self.STATUS_CODE:
+            # Define behavior for a 200 status code for use in test_super_agent_health.py
+            if self.STATUS_CODE == 200:
+                payload = {"return_value": "Hello World!"}
+                response_data = json_encode(payload).encode("utf-8")
+                return self.STATUS_CODE, response_data
```
TimPansino, line 87: This thing should probably be moved to `testing_support/` if it's getting reused in 2 different test suites now.

TimPansino (changes requested):

newrelic/config.py
```diff
@@ -1047,6 +1052,7 @@ def _load_configuration(
                 raise newrelic.api.exceptions.ConfigurationError("New Relic configuration not found in TOML file.")
             _config_object.read_dict(_toml_config_to_configparser_dict(newrelic_section))
     elif not _config_object.read([config_file]):
+        super_agent_instance.set_health_status("invalid_config")
         raise newrelic.api.exceptions.ConfigurationError(f"Unable to open configuration file {config_file}.")
```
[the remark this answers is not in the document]
umaannamalai: Good call, this feature was started before we merged the TOML logic in.

tests/agent_features/test_super_agent_health_check.py
```diff
@@ -0,0 +1,215 @@
...
+    monkeypatch.setenv("NEW_RELIC_SUPERAGENT_HEALTH_DELIVERY_LOCATION", file_path)
+
+    # Re-initialize the agent to allow the health check thread to start and assert that it did
+    _reset_configuration_done()
+    initialize()
+
+    running_threads = threading.enumerate()
+
+    assert len(running_threads) == 2
+    assert running_threads[1].name == "NR-Control-Health-Main-Thread"
```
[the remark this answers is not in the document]
umaannamalai: The two expected threads here that we are asserting on are one health thread as well as one main thread that runs in the agent. For the sake of the test, we only care about ensuring the health thread is running in addition to the main thread. I can add a comment here to clarify what those expected threads are supposed to be.

newrelic/core/super_agent_health.py
```diff
@@ -115,88 +132,81 @@ def health_check_enabled(self):
...
 
-        return True
+        return is_valid_file_delivery_location(health_file_location)
+
+    @property
+    def is_healthy(self):
+        return True if self.status == "Healthy" else False
 
     def set_health_status(self, health_status, response_code=None, info=None):
+        license_key_error = True if self.status == "License key missing in configuration" or self.status == "Invalid license key (HTTP status code 401)" else False
```
TimPansino, line 143: I don't think we can be comparing the previous status to the current one based on the text of the error message, that's both a slow operation and also incredibly fragile. If anyone updates the messages up above it'll introduce bugs unless they happen to spot that this message is duplicated and used in logic.

Can we store either the `license_key_error = True/False` on `self`, or store the `health_status` enum value on the `self`? I would think the latter would be more useful and general.

TimPansino:
There were way too many string comparisons and hard coded strings going on in the super agent health file, so I've refactored it somewhat.

I'm looking for it to use integers in every conceivable spot to minimize string comparisons and copies, as well as reduce the risk of a future change introducing bugs because we're comparing status messages that are hard coded in 2 places, rather than using integer codes.

To that end, I've redone the file to use the original integer codes as the enum values, eliminated all cases where we have a string like `NR-APM-003`, and instead construct those when writing out to the file. I removed all hard coded references and replaced them with integers referenced via enum.

I also removed the large block of logic to construct different status messages and modeled it somewhat after the logic used in `newrelic/core/agent_protocol.py` to always attempt to format the string with the passed in values, and rely on string formatting to throw away unused values.

tests/agent_features/test_super_agent_health_check.py
```diff
@@ -0,0 +1,219 @@
...
+    super_agent_instance.write_to_health_file()
+
+    contents = get_health_file_contents(tmp_path)
+
+    # Assert on contents of health file
+    assert len(contents) == 5
+    assert contents[0] == "healthy: False\n"
+    assert contents[1] == "status: Invalid license key (HTTP status code 401)\n"
+    assert contents[2] == "start_time_unix_nano: 1234567890\n"
+    assert re.search(r"status_time_unix_nano:\s\d+", contents[3]) is not None
```
TimPansino, line 84:
```suggestion
    assert int(re.search(r"status_time_unix_nano: (\d+)", contents[3]).group(1)) > 0
```

tests/agent_features/test_super_agent_health_check.py
```diff
@@ -0,0 +1,219 @@
...
+    super_agent_instance.write_to_health_file()
+
+    contents = get_health_file_contents(tmp_path)
+
+    # Assert on contents of health file
+    assert len(contents) == 4
+    assert contents[0] == "healthy: True\n"
+    assert contents[1] == "status: Healthy\n"
+    assert re.search(r"status_time_unix_nano:\s\d+", contents[3]) is not None
+    assert contents[3].startswith("status_time_unix_nano:") is True
```
TimPansino, line 61:
```suggestion
    assert int(re.search(r"status_time_unix_nano: (\d+)", contents[3]).group(1)) > 0
```

TimPansino (approved):

hmstepanek (approved):
Looks good!

PR #1251, opened 2024-11-08
Daphne HTTP/2 Testing

TimPansino:
# Overview

* Add tests to Daphne that exercise the HTTP/2 server capabilities.
  * Identical to the tests for Hypercorn.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1202,12 +1202,12 @@ jobs:
 
   valkey:
     env:
-      TOTAL_GROUPS: 2
+      TOTAL_GROUPS: 1
 
     strategy:
       fail-fast: false
       matrix:
-        group-number: [1, 2]
+        group-number: [1]
 
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -1202,12 +1202,12 @@ jobs:
 
   valkey:
     env:
-      TOTAL_GROUPS: 2
+      TOTAL_GROUPS: 1
```
hmstepanek, line 1205: Along with this change I think we also will need to edit the valkey_settings in db_settings to only return the first group too.

hmstepanek (commented):
Just one thing about updating the db_settings for valkey, otherwise looks good.

.github/workflows/tests.yml
TimPansino: This doesn't change the number of valkey "instances" available, it's just reducing the number of GHA runners used to run these tests. db_settings doesn't need to change.

hmstepanek (approved):

PR #1254, opened 2024-11-19
Instrument PyMongo AsyncMongoClient

TimPansino:
# Overview

* Add instrumentation and testing for PyMongo's AsyncMongoClient
* Update instrumentation of existing synchronous MongoClient with newly added methods
* Add test to validate that there are no uninstrumented future client methods
* Add instance info to existing synchronous MongoClient for simple cases
  * Load balancing setups intentionally left without instance info to avoid reporting confusing and often incorrect information about query routing.


hmstepanek (commented):
Just one thing regarding using safer attribute access-otherwise looks good!

hmstepanek (approved):

PR #1255, opened 2024-11-19
Add Motor Instrumentation

TimPansino:
# Overview

* Add instrumentation and testing for Motor
  * Just like PyMongo, load balancing setups intentionally left without instance info to avoid reporting confusing and often incorrect information about query routing.


newrelic/hooks/datastore_motor.py
```diff
@@ -12,35 +12,131 @@
...
+    "update_one",
+    "update_search_index",
+)
 
-def _nr_wrapper_Motor_getattr_(wrapped, instance, args, kwargs):
 
-    def _bind_params(name, *args, **kwargs):
-        return name
+def instance_info(collection):
+    nodes = collection.database.client.nodes
```
hmstepanek, line 61: Should we use getattr/hasattr here in case these attr's don't exist? There's a couple place where we should probably do this but I'm just noting one occurrence here.

hmstepanek (commented):
Same thing as the other PR, I just want to make sure we aren't going to fail with an attr doesn't exist error.

hmstepanek (approved):

hmstepanek (approved):

PR #1261, opened 2024-11-22
Package json fix

lrafeei:
This PR fixes a bug in the package_version extraction where it passes a module type instead of a string.  This happens in the case of a version being declared within a class property.

After transferring the module versioning populating mechanism from startup to the harvest cycle, this bug disabled all data collection for affected apps.


newrelic/common/package_version_utils.py
```diff
@@ -86,6 +84,15 @@ def _get_package_version(name):
...
                     continue
 
+                # Some frameworks (such as `pypdfium2`) may use a class
+                # property to define the version.  Because class properties
+                # are not callable we need to check if the result is
+                # anything other than a string, tuple, or list.  If so,
+                # we need to skip this method of version retrieval and use
+                # `pkg_resources` or `importlib.metadata`.
+                if version and not isinstance(version, (str, tuple, list)):
+                    break
```
TimPansino, line 94:
```suggestion
                # Some frameworks (such as `pypdfium2`) may use a class
                # property to define the version.  Because class properties
                # are not callable we need to check if the result is
                # anything other than a string, tuple, or list.  If so,
                # we need to skip this method of version retrieval and use
                # `pkg_resources` or `importlib.metadata`.
                if version and not isinstance(version, (str, tuple, list)):
                    continue
```

Instead of `break` can we use `continue` to check if any of the other attributes are potentially the version we're looking for? This would let us combine these 2 logic blocks.

For example, if `module.version` is a module or some other object but also exports `module.version_tuple` which is just the tuple we're looking for.

tests/agent_unittests/test_package_version_utils.py
```diff
@@ -150,3 +150,21 @@ def test_version_caching(monkeypatch):
...
+        def version(self):
+            return "1.2.3"
+
+    monkeypatch.setattr(pytest, "version", FakeModule.version, raising=False)
+    monkeypatch.setattr(
+        sys.modules["importlib"].metadata, "version", lambda x: "1.2.3", raising=False  # pylint: disable=E1101
+    )
+
+    version = get_package_version("pytest")
+    assert version == "1.2.3"
```
TimPansino, line 170: To go with the suggestion above, can we get a test that does almost this exact thing but instead sets `version_tuple` as an actual tuple while `version` is a property?

tests/agent_unittests/test_package_version_utils.py
```diff
@@ -150,3 +150,21 @@ def test_version_caching(monkeypatch):
     del sys.modules["mymodule"]
     version = get_package_version("mymodule")
     assert version not in NULL_VERSIONS, version
+
+
+@SKIP_IF_NOT_IMPORTLIB_METADATA
```
hmstepanek, line 155: Instead of only running this against importlib (though I realize we are going to drop 3.7 soon in which case it would only use importlib) I think we should just do a more inclusive assertion at the bottom like the rest of the tests and not replace the importlib version-just let importlib or pkg_resources find the existing pytest version:
```
    assert version not in NULL_VERSIONS, version
```

hmstepanek (commented):
Just had one suggestion regarding the tests. This fix looks good!

newrelic/common/package_version_utils.py
```diff
@@ -81,17 +81,14 @@ def _get_package_version(name):
 
                 # In certain cases like importlib_metadata.version, version is a callable
                 # function.
```
TimPansino, line 83:
```suggestion
```
This comment should also get deleted now

tests/agent_unittests/test_package_version_utils.py
```diff
@@ -150,3 +150,43 @@ def test_version_caching(monkeypatch):
...
+    class FakeModule:
+        @property
+        def version(self):
+            return "1.2.3"
+
+    monkeypatch.setattr(pytest, "version", FakeModule.version, raising=False)
+    monkeypatch.setattr(pytest, "version_tuple", (1, 2, 3), raising=False)
+    monkeypatch.setattr(
+        sys.modules["importlib"].metadata, "version", lambda x: "1.2.3", raising=False  # pylint: disable=E1101
+    )
```
TimPansino, line 189:
```suggestion
```
Delete that and let the assertion test that `version_tuple` was used.

tests/agent_unittests/test_package_version_utils.py
hmstepanek: Could also add to this assert isinstance(version, str) or something to make sure it's not a property like it was before.

hmstepanek (approved):

TimPansino (approved):

PR #1266, opened 2024-11-29
fix(NewRelicContextFormatter): add cause and context to stack traces

sweco:
# Overview
`NewRelicContextFormatter` logs stack traces without causes which makes many errors hard to debug.

Let's take a look at the following code:

```python
try:
    try:
        raise ValueError("First exception")
    except:
        raise ValueError("Second exception")
except Exception as e:
    logging.error("Encountered error", exc_info=e)
```

Before the PR, the exception message looked as follows:

```
  File ".../newrelic_exception_cause.py", line 60, in <module>
    raise ValueError("Second exception")
```

After the PR, it shows more details on the first cause:

```
  File ".../newrelic_exception_cause.py", line 58, in <module>
    raise ValueError("First exception")
ValueError: First exception
During handling of the above exception, another exception occurred:
  File ".../newrelic_exception_cause.py", line 60, in <module>
    raise ValueError("Second exception")
ValueError: Second exception
```

# Related Github Issue
No Github issue created.

# Testing
Tests were updated to test a particular case where exception has a cause.

### DIFF
--- a/newrelic/api/log.py
+++ b/newrelic/api/log.py
@@ -16,7 +16,7 @@ import json
 import logging
 import re
 import warnings
-from traceback import format_tb
+from traceback import format_exception
 
 from newrelic.api.application import application_instance
 from newrelic.api.time_trace import get_linking_metadata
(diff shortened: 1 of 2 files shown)


TimPansino:
We'll give this a review after the US holiday season is over, we're going to avoid the risk of changing anyone's logs during this time period.

TimPansino (approved):

PR #1281, opened 2025-01-07
Remove extra spacing artifacts

lrafeei:
This PR removes instances of strings that are separated by a space unintentionally.

Example:

`text = "This is not" "supposed to happen"`


lrafeei:
This PR is moot now that the Ruff megalinter has already fixed the concatenated string issue

PR #1288, opened 2025-01-22
Refactor agent control config.

umaannamalai:
This PR updates configuration naming/ default behavior in agent control

- Removal of the `NEW_RELIC_AGENT_CONTROL_FLEET_ID` env var check
- Addition of check of `NEW_RELIC_AGENT_CONTROL_ENABLED` to determine whether health check is run (defaulted to False)
- Addition of default value for health delivery location (file:///newrelic/apm/health) 
- Explicit int typecasting of reporting frequency


newrelic/core/agent_control_health.py
```diff
@@ -130,13 +125,23 @@ def __init__(self):
...
+    def health_delivery_location(self):
+        health_file_location = os.environ.get(
+            "NEW_RELIC_AGENT_CONTROL_HEALTH_DELIVERY_LOCATION", "file:///newrelic/apm/health"
+        )
+        # Explicitly check if the variable was set to an empty string so we can assign the default
+        if not health_file_location:
+            health_file_location = "file:///newrelic/apm/health"
 
-        return is_valid_file_delivery_location(health_file_location)
+        return health_file_location
```
TimPansino, line 144: What about this instead for a one-liner that doesn't use an if statement?
```
health_file_location = os.environ.get("NEW_RELIC_AGENT_CONTROL_HEALTH_DELIVERY_LOCATION", "") or "file:///newrelic/apm/health"
```

newrelic/core/agent_control_health.py
```diff
@@ -130,13 +125,23 @@ def __init__(self):
...
     def health_check_enabled(self):
-        fleet_id_present = os.environ.get("NEW_RELIC_AGENT_CONTROL_FLEET_ID", None)
-        if not fleet_id_present:
+        # Default to False - this must be explicitly set to True by the sidecar/ operator to enable health check
+        agent_control_enabled = os.environ.get("NEW_RELIC_AGENT_CONTROL_ENABLED", False)
+        if not agent_control_enabled:
             return False
 
-        health_file_location = os.environ.get("NEW_RELIC_AGENT_CONTROL_HEALTH_DELIVERY_LOCATION", None)
+        return is_valid_file_delivery_location(self.health_delivery_location)
```
TimPansino, line 133: Can we use `_environ_as_bool()` for this? This currently just checks if it's an empty string or not. It would be nice to be able to explicitly disable this by setting it to "false" or "0" etc. If we do that currently we're likely to get confused why it's still enabled.

```python
        from newrelic.core.config import _environ_as_bool

        agent_control_enabled = _environ_as_bool("NEW_RELIC_AGENT_CONTROL_ENABLED", False)
        if not agent_control_enabled:
            return False

        return is_valid_file_delivery_location(self.health_delivery_location)
```

newrelic/core/agent_control_health.py
```diff
@@ -216,7 +215,15 @@ def agent_control_health_instance():
...
 def agent_control_healthcheck_loop():
-    reporting_frequency = os.environ.get("NEW_RELIC_AGENT_CONTROL_HEALTH_FREQUENCY", 5)
+    try:
+        reporting_frequency = int(os.environ.get("NEW_RELIC_AGENT_CONTROL_HEALTH_FREQUENCY", 5))
+        # If we have an invalid integer value for frequency, default back to 5
+        if reporting_frequency <= 0:
+            reporting_frequency = 5
+    except Exception:
+        # If we run into an exception when int typecasting, default back to 5
+        reporting_frequency = 5
```
TimPansino, line 225: This one could probably be `_environ_as_int()` with a default of 5, and then run this check afterwards.

```
        if reporting_frequency <= 0:
            reporting_frequency = 5
```

TimPansino (commented):
Couple suggestions but overall looks fine

TimPansino (approved):

Issue #1284, opened 2025-01-10
Partially incompatible with structlog configuration


james-certn:
Using structlog's ["Rendering Using structlog-based Formatters Within `logging`"](https://www.structlog.org/en/stable/standard-library.html#rendering-using-structlog-based-formatters-within-logging) steps, new relic is unable to extract `"event"` -> `"message"`

**Description**
Log entries in New Relic are missing the `"message"` value

**Expected Behavior**
Log entries in New Relic have the `"message"` value populated correctly

**Troubleshooting**
Diagnosing the issue, I noticed:
- newrelic appends [a `new_relic_event_consumer` processor](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/hooks/logger_structlog.py#L41) to the structlog processors list, and ensures it is at the tail end
  - it does this [in `wrap__process_event(...)`](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/hooks/logger_structlog.py#L90)
- `new_relic_event_consumer` inspects the type of the `event` parameter:
  - if str-like it directly copies it to `message`
  - if dict-like it copies the `"event"` value from the dict into `message`
- this works for the 'simplified' structlog setup, but not for the 'ambitious' (structlog's wording, not mine!) setup (linked at the top)
  - in this setup, the last processor when configurating must be `structlog.stdlib.ProcessorFormatter.wrap_for_formatter`
  - `wrap_for_formatter` expects to _always be the last_ processor prior to the `ProcessorFormatter` handling its output
    - as such, `wrap_for_formatter` knows it can _safely change the shape of the data_ it returns to be compatible with `ProcessorFormatter`
    - this format is a `tuple` (specifically, the `(*args, **kwargs})` that `ProcessorFormatter` expects to receive
  - however, because newrelic ensures its `new_relic_event_consumer` is always at the tail end of the processors list, it ends up:
    - receiving a `tuple` of `(*args, **kwargs})`
    - [skips over it, because it doesn't know what to do with that data type](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/hooks/logger_structlog.py#L57-L59)
    - returns the `tuple` (as it should), which successfully gets passed to `ProcessorFormatter`

**Steps to Reproduce**
1. configure structlog in ["Rendering Using structlog-based Formatters Within `logging`"](https://www.structlog.org/en/stable/standard-library.html#rendering-using-structlog-based-formatters-within-logging) fashion
[...]

PR #1289, opened 2025-01-24
Structlog processor formatter

lrafeei:
This PR provides [a fix](https://github.com/newrelic/newrelic-python-agent/issues/1284) for the case where structlog's [ProcessorFormatter](https://www.structlog.org/en/stable/api.html#structlog.stdlib.ProcessorFormatter) is used as a [logging.Formatter](https://docs.python.org/3/library/logging.html#logging.Formatter) for both logging as well as structlog log entries.

For more information about this functionality: https://www.structlog.org/en/stable/standard-library.html#rendering-using-structlog-based-formatters-within-logging

### DIFF
--- a/newrelic/hooks/logger_structlog.py
+++ b/newrelic/hooks/logger_structlog.py
@@ -58,9 +58,6 @@ def new_relic_event_consumer(logger, level, event):
             _event = event[0][0]
             message = original_message = _event.get("event")
             event_attrs = {k: v for k, v in _event.items() if k != "event"}
-            # event_attrs = {
-            #     k: v for extra, log_attrs in event[1].items() if extra == "extra" for k, v in log_attrs.items()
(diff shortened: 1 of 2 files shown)


newrelic/hooks/logger_structlog.py
```diff
@@ -64,6 +68,8 @@ def new_relic_event_consumer(logger, level, event):
                 event = message
             elif isinstance(event, dict) and "event" in event:
                 event["event"] = message
+            elif isinstance(event, tuple) and event[0][0] and "event" in event[0][0]:
+                event[0][0]["event"] = message
```
TimPansino, line 72: Let's throw this into a try/except block to prevent crashes, make this an else and then an if underneath.

TimPansino (approved):

PR #1291, opened 2025-01-27
Add max three app name status for agent control health check

umaannamalai:
This PR adds a new status code to the agent that sets health check reporting to unhealthy when a user has configured more than 3 app names.


hmstepanek (approved):
There's one linter issue but otherwise looks good to me:
```
newrelic/core/agent_control_health.py:68:1: E302 expected 2 blank lines, found 1
```

PR #1283, opened 2025-01-08
Add aws entity support for Kinesis

hmstepanek:
# Overview
Add aws entity support for Kinesis.

PR #1300, opened 2025-02-03
AWS Kinesis Delivery Streams (Firehose) Entity Mapping

TimPansino:
# Overview

* Add instrumentation and tests for AWS Firehose.
* Correct AWS Kinesis Instrumentation naming when no stream target is present.
* Tweak boto3/botocore in tox file.
* Fix S3Transfer test missing absolute path.

# Related Github Issue

Currently depends on #1283 but will be rebased.

### DIFF
--- a/newrelic/hooks/external_botocore.py
+++ b/newrelic/hooks/external_botocore.py
@@ -1293,6 +1293,12 @@ CUSTOM_TRACE_POINTS = {
         extract_agent_attrs=extract_firehose_agent_attrs,
         library="Firehose",
     ),
+    ("firehose", "put_record"): aws_function_trace(
+        "put_record", extract_firehose, extract_agent_attrs=extract_firehose_agent_attrs, library="Firehose"
+    ),
+    ("firehose", "put_record_batch"): aws_function_trace(
+        "put_record_batch", extract_firehose, extract_agent_attrs=extract_firehose_agent_attrs, library="Firehose"
+    ),
     ("firehose", "start_delivery_stream_encryption"): aws_function_trace(
         "start_delivery_stream_encryption",
         extract_firehose,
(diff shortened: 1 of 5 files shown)


newrelic/hooks/external_botocore.py
```diff
@@ -1249,6 +1273,54 @@ def wrap_serialize_to_request(wrapped, instance, args, kwargs):
...
+    ("firehose", "tag_delivery_stream"): aws_function_trace(
+        "tag_delivery_stream", extract_firehose, extract_agent_attrs=extract_firehose_agent_attrs, library="Firehose"
+    ),
+    ("firehose", "untag_delivery_stream"): aws_function_trace(
+        "untag_delivery_stream", extract_firehose, extract_agent_attrs=extract_firehose_agent_attrs, library="Firehose"
+    ),
+    ("firehose", "update_destination"): aws_function_trace(
+        "update_destination", extract_firehose, extract_agent_attrs=extract_firehose_agent_attrs, library="Firehose"
+    ),
+    ("firehose", "put_record"): aws_message_trace(
```
hmstepanek, line 1318: According to the [spec](https://source.datanerd.us/agents/agent-specs/blob/main/implementation_guides/aws-sdk.md#kinesis-firehose) we aren't supposed to create message traces for these. I was also inclined to add them as message traces but we aren't supposed to-see comment [here](https://source.datanerd.us/agents/agent-specs/pull/706/files#r826578).

hmstepanek (changes requested):
Other than removing the message traces it looks good.

newrelic/hooks/external_botocore.py
TimPansino: That seems bizarre but we can go with that for now

hmstepanek (approved):
The firehose looks good. It looks like I didn't handle those cases where the ARN can be either the data stream or the consumer. You don't have to add that here but I think we should do that. I'll leave that up to you as to when we do that (either here or in a follow up)-but in those cases where it's either-or we still need to call extract function(s)-perhaps different extract functions.

PR #1304, opened 2025-02-12
Ruff Format

TimPansino:
# Overview

* Run `ruff format` on the entire repo.


hmstepanek (approved):
This looks good to me. I notice there's a couple comments against the PR that I don't really want to see on future PRs. Could we disable these (can be in a follow up):
* ANN201 Missing return type annotation for public function `main`
* D103 Missing docstring in public function
* ANN001 Missing type annotation for function argument `f`
* D100 Missing docstring in public module

TimPansino:
> This looks good to me. I notice there's a couple comments against the PR that I don't really want to see on future PRs. Could we disable these (can be in a follow up):
> 
> * ANN201 Missing return type annotation for public function `main`
> * D103 Missing docstring in public function
> * ANN001 Missing type annotation for function argument `f`
> * D100 Missing docstring in public module

Yeah I was planning to disable `flake8-annotations` and `pydocstyle` rules completely for now. If we ever do a pass at conforming to documentation standards then those might be useful.

hmstepanek (approved):

Issue #1204, opened 2024-09-03
Support Elasticsearch async client


jairhenrique:
### Is your feature request related to a problem? Please describe.
The asynchronous client for elasticsearch does not put the database and APM metrics related to the tool. This data is only presented if we use the synchronous version.

### Priority
Really Want

PR #1309, opened 2025-02-20
Adds async Elasticsearch support

fagnerzulin:
# Overview
This Pull Request instruments the Elasticsearch python library with asynchronous implementation. Basically, I reused what was already implemented for the synchronous version, however, I ran into some issues with the unit tests in the following files:

- tests/datastore_elasticsearch/test_async_connection.py
- tests/datastore_elasticsearch/test_async_mget.py
- tests/datastore_elasticsearch/test_async_multiple_dbs.py

If anyone can help me understand what’s going on, I would be grateful.

# Related Github Issue
https://github.com/newrelic/newrelic-python-agent/issues/1204

# Testing
exercise your changes.

### DIFF
--- a/newrelic/hooks/datastore_elasticsearch.py
+++ b/newrelic/hooks/datastore_elasticsearch.py
@@ -11,8 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from common.async_wrapper import coroutine_wrapper
 
-from newrelic.api.datastore_trace import DatastoreTrace
+from newrelic.api.datastore_trace import DatastoreTrace, DatastoreTraceWrapper
 from newrelic.api.transaction import current_transaction
 from newrelic.common.object_wrapper import function_wrapper, wrap_function_wrapper
 from newrelic.common.package_version_utils import get_package_version_tuple
@@ -47,7 +48,7 @@ def _extract_args_allocation_explain_index(
     include_yes_decisions=None,
     index=None,
     *args,
-    **kwargs
+    **kwargs,
 ):
     return _index_name(index)
 
@@ -106,6 +107,12 @@ def instrument_es_methods(module, _class, client_methods, prefix=None):
             wrap_elasticsearch_client_method(module, _class, method_name, arg_extractor, prefix)
 
 
+def instrument_async_es_methods(module, _class, client_methods, prefix=None):
+    for method_name, arg_extractor in client_methods:
+        if hasattr(getattr(module, _class), method_name):
+            wrap_async_elasticsearch_client_method(module, _class, method_name, arg_extractor, prefix)
+
+
 def wrap_elasticsearch_client_method(module, class_name, method_name, arg_extractor, prefix=None):
     def _nr_wrapper_Elasticsearch_method_(wrapped, instance, args, kwargs):
         transaction = current_transaction()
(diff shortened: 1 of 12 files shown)


newrelic/hooks/datastore_elasticsearch.py
```diff
@@ -145,6 +152,42 @@ def _nr_wrapper_Elasticsearch_method_(wrapped, instance, args, kwargs):
...
+            index = arg_extractor(*args, **kwargs)
+
+        if prefix:
+            operation = f"{prefix}.{method_name}"
+        else:
+            operation = method_name
+
+        transaction._nr_datastore_instance_info = (None, None, None)
+
+        dt = DatastoreTrace(product="Elasticsearch", target=index, operation=operation, source=async_wrapper)
```
TimPansino, line 174: The source argument here should be `source=wrapped`, that's for reporting the code's file name and line number for the Code Level Metrics feature. The `async_wrapper` code isn't needed here. The actual wrapper needs to be a coroutine that awaits `wrapped()` (probably, will check that's always true).

newrelic/hooks/datastore_elasticsearch.py
```diff
@@ -596,21 +766,49 @@ def _bind_params(host="localhost", port=9200, *args, **kwargs):
...
+    def _bind_params(host="localhost", port=9200, *args, **kwargs):
+        return host, port
+
+    host, port = _bind_params(*args, **kwargs)
+    port = str(port)
+    instance._nr_host_port = (host, port)
+
+    return wrapped(*args, **kwargs)
+
+
```
TimPansino, line 781: Seems like this wrapper is identical to `_nr_Connection__init__wrapper`, we can reuse the original instead.

newrelic/hooks/datastore_elasticsearch.py
```diff
@@ -596,21 +766,49 @@ def _bind_params(host="localhost", port=9200, *args, **kwargs):
...
     instance._nr_host_port = (instance.host, str(instance.port))
     return result
 
 
+def async_BaseNode__init__wrapper(wrapped, instance, args, kwargs):
+    result = wrapped(*args, **kwargs)
+    instance._nr_host_port = (instance.host, str(instance.port))
+    return result
+
+
```
TimPansino, line 801: This is also identical to the original wrapper, we can reuse it instead.

newrelic/hooks/datastore_elasticsearch.py
```diff
@@ -636,6 +834,31 @@ def _nr_get_connection_wrapper(wrapped, instance, args, kwargs):
...
+    try:
+        tracer_settings = transaction.settings.datastore_tracer
+
+        if tracer_settings.instance_reporting.enabled:
+            host, port_path_or_id = conn._nr_host_port
+            instance_info = (host, port_path_or_id, None)
+    except Exception:
+        instance_info = ("unknown", "unknown", None)
+
+    transaction._nr_datastore_instance_info = instance_info
```
TimPansino, line 857: This wrapper is mostly right, but unfortunately it doesn't work in async contexts. When connecting to multiple databases at once there's a race condition that gets introduced if we store instance_info on the transaction object directly. We'll have to instead work out a different design where we stash it elsewhere, or call `current_trace()` and carefully insert it.

TimPansino (commented):
Reviewed just the instrumentation so far but these are the issues I immediately see. I'll take a pass at fixing them and redesigning the instance_info section for async, and we'll see if that fixes the rest of the tests.

newrelic/hooks/datastore_elasticsearch.py
fagnerzulin: Ok, I fixed it.

newrelic/hooks/datastore_elasticsearch.py
fagnerzulin: Resolved!

newrelic/hooks/datastore_elasticsearch.py
fagnerzulin: Resolved!

newrelic/hooks/datastore_elasticsearch.py
fagnerzulin: Do you have any suggestions on how to do this?

TimPansino:
Update from our side, I'm going to take a pass at this next week. We had a bit of crunch to finish a few things by the end of this quarter and I haven't been able to spend any time on this.

TimPansino:
Closing in favor of #1362

PR #1329, opened 2025-03-24
MySQL Updates

TimPansino:
# Overview

* Update MySQLdb instrumentation to work with the Python 3 fork [mysqlclient](https://pypi.org/project/mysqlclient/).
  * Adjust wrappers to now support cursors which didn't exist previously.
  * Restore deleted tests from the Python 2 iteration of the package.
  * Update all tests to follow new usage patterns with cursors.
* Update instrumentation for PyMySQL to support v1.0+, and drop support for older v0.10 signatures.
  * Previously this didn't conform well to DBAPI2 and had very custom implementations. This has changed, and now connections and cursors are properly separated making old instrumentation incompatible with new versions. Update all instrumentation to work with the new format, and drop support for old code from 2020.


hmstepanek (changes requested):
I'm looking at the connect implementation for mysqlclient and I'm seeing duplicated `database` and `db`, `password` and `passwd` input params-do we handle both those cases? I only see testing for `db` and `passwd`.

https://github.com/PyMySQL/mysqlclient/blob/main/src/MySQLdb/connections.py#L62-L68

PR #1331, opened 2025-03-24
Enable ruff Formatter in CI

TimPansino:
# Overview

* Enable `ruff format` as part of megalinter automatically in CI.
* Update megalinter to v8.5.0.
* Fix existing formatting and linting issues.


tests/external_aiobotocore/test_bedrock_chat_completion.py
```diff
@@ -19,6 +19,7 @@
 import botocore.eventstream
 import botocore.exceptions
 import pytest
+from conftest import BOTOCORE_VERSION  # pylint: disable=E0611
```
hmstepanek, line 22: Can we drop the pylint disable on this import?

hmstepanek (approved):
I think this looks good assuming the tests pass CI.

tests/external_aiobotocore/test_bedrock_chat_completion.py
TimPansino: I'll round up all the pylint and other weird noqa markers in a big pass at some point

PR #1345, opened 2025-04-02
Add Gemini Embeddings

umaannamalai:
This PR adds instrumentation + tests for sync and async embeddings in `google-genai`.

### DIFF
--- a/newrelic/hooks/mlmodel_gemini.py
+++ b/newrelic/hooks/mlmodel_gemini.py
@@ -1,0 +1,204 @@
+# Copyright 2010 New Relic, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import sys
+import traceback
+import uuid
+
+from newrelic.api.function_trace import FunctionTrace
+from newrelic.api.time_trace import get_trace_linking_metadata
+from newrelic.api.transaction import current_transaction
+from newrelic.common.object_wrapper import wrap_function_wrapper
+from newrelic.common.package_version_utils import get_package_version
+from newrelic.core.config import global_settings
+
+GEMINI_VERSION = get_package_version("google-genai")
+EXCEPTION_HANDLING_FAILURE_LOG_MESSAGE = "Exception occurred in Gemini instrumentation: While reporting an exception in Gemini, another exception occurred. Report this issue to New Relic Support.\n%s"
+RECORD_EVENTS_FAILURE_LOG_MESSAGE = "Exception occurred in Gemini instrumentation: Failed to record LLM events. Please report this issue to New Relic Support.\n%s"
+
+_logger = logging.getLogger(__name__)
+
+
+def wrap_embed_content_sync(wrapped, instance, args, kwargs):
+    transaction = current_transaction()
+    if not transaction:
+        return wrapped(*args, **kwargs)
+
+    settings = transaction.settings if transaction.settings is not None else global_settings()
+    if not settings.ai_monitoring.enabled:
+        return wrapped(*args, **kwargs)
+
+    # Framework metric also used for entity tagging in the UI
+    transaction.add_ml_model_info("Gemini", GEMINI_VERSION)
+    transaction._add_agent_attribute("llm", True)
+
+    # Obtain attributes to be stored on embedding events regardless of whether we hit an error
+    embedding_id = str(uuid.uuid4())
+
+    ft = FunctionTrace(name=wrapped.__name__, group="Llm/embedding/Gemini")
+    ft.__enter__()
+    linking_metadata = get_trace_linking_metadata()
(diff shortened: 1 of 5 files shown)


newrelic/hooks/mlmodel_gemini.py
```diff
@@ -0,0 +1,204 @@
...
+        model = kwargs.get("model")
+
+        notice_error_attributes = {
+            "http.statusCode": getattr(exc, "code", None),
+            "error.message": getattr(exc, "message", None),
+            "error.code": getattr(exc, "status", None),  # ex: 'NOT_FOUND'
+            "embedding_id": embedding_id,
+        }
+    except Exception:
+        _logger.warning(EXCEPTION_HANDLING_FAILURE_LOG_MESSAGE, traceback.format_exception(*sys.exc_info()))
```
hmstepanek, line 118: Is there a difference between this and `_logger.warning(EXCEPTION_HANDLING_FAILURE_LOG_MESSAGE, exc_info=True)`?

newrelic/hooks/mlmodel_gemini.py
```diff
@@ -0,0 +1,204 @@
...
+
+        full_embedding_response_dict.update(_get_llm_attributes(transaction))
+
+        transaction.record_custom_event("LlmEmbedding", full_embedding_response_dict)
+
+    except Exception:
+        _logger.warning(RECORD_EVENTS_FAILURE_LOG_MESSAGE, traceback.format_exception(*sys.exc_info()))
+
+
+def _get_llm_attributes(transaction):
```
hmstepanek, line 187: Random thought: should we make this a helper function that can be used by all llm implementations?

hmstepanek (commented):
Main question was about the `contents` value and I had one suggestion about a refactor but totally don't have to do that in this PR.

newrelic/hooks/mlmodel_gemini.py
umaannamalai: Yes, I had the same thought! I was planning to do this in a separate PR after the core instrumentation is in for Gemini.

newrelic/hooks/mlmodel_gemini.py
umaannamalai: Tested this and they are functionally equivalent so I'll swap to use `exc_info=True`

newrelic/hooks/mlmodel_gemini.py
```diff
@@ -0,0 +1,204 @@
...
+                else None
+            ),
+            "request.model": model,
+            "vendor": "gemini",
+            "ingest_source": "Python",
+            "duration": ft.duration * 1000,
+            "error": True,
+        }
+        if settings.ai_monitoring.record_content.enabled:
+            error_embedding_dict["contents"] = embedding_content
```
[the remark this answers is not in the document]
umaannamalai: I had changed this to match the terminology used within Gemini but it may make sense to keep it consistent with our other instrumentations so I will change this back.

newrelic/hooks/mlmodel_gemini.py
```diff
@@ -0,0 +1,204 @@
...
+        _logger.warning(RECORD_EVENTS_FAILURE_LOG_MESSAGE, traceback.format_exception(*sys.exc_info()))
+
+
+def _record_embedding_success(transaction, embedding_id, linking_metadata, kwargs, ft):
+    settings = transaction.settings if transaction.settings is not None else global_settings()
+    span_id = linking_metadata.get("span.id")
+    trace_id = linking_metadata.get("trace.id")
+    try:
+        embedding_content = kwargs.get("contents")
+        embedding_content = str(embedding_content) if embedding_content else ""
```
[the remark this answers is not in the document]
umaannamalai: Yeah, it seems that you can pass a single string or a list of strings (ex: https://github.com/googleapis/python-genai/blob/main/google/genai/models.py#L4028). My thought was to typecast to string in case it is a list so that we send the full input. 

In terms of embedding_content being `None`,  I was thinking more along the lines of there being a failure to grab the input from kwargs in the previous line: 
`embedding_content = kwargs.get("contents")`
I was thinking down the road if they change the structure or argument names within `kwargs` then we may not get anything back if we try searching for `contents` within the request so this was to safeguard against that case.

tests/testing_support/ml_testing_utils.py
```diff
@@ -44,6 +44,8 @@ def events_sans_content(event):
             del _event[1]["input"]
         elif "content" in _event[1]:
             del _event[1]["content"]
+        elif "contents" in _event[1]:
```
hmstepanek, line 47: Given the change back to `input` I don't think we need this change anymore:
```suggestion
```

tests/testing_support/ml_testing_utils.py
```diff
@@ -44,6 +44,8 @@ def events_sans_content(event):
...
+            del _event[1]["contents"]
```
hmstepanek, line 48:
```suggestion
```

newrelic/hooks/mlmodel_gemini.py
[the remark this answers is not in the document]
hmstepanek: Ok that makes sense. In cases like that where the argument name might change, I think it's good to consider if it did-how would we catch it? Is this covering up that issue or exposing it so we can catch it quickly and fix it? Also, if this might be a list-how does that impact the token counting (might be something worth asking the AI team)? Would it be better to expand our spec for the token counting function to accept a list and cast it to a string after when recording the input into the event instead?

hmstepanek (changes requested):
Had one minor fixup and some continued discussion on the contents parameter. If you want to merge this and then investigate that further with the AI team I think that's fine since it's going to a feature branch anyway.

PR #1348, opened 2025-04-07
Async Elasticsearch

TimPansino:
# Overview

* Continuing #1309 from @fagnerzulin 
  * Rebased with main post formatter/linter changes, required commit history rewrite.
  * Misc formatting and linter changes from the new tools
* Fixup testing issues


TimPansino:
Having trouble with `aiohttp` throwing weird errors around the timeout context manager, but the instance info issues are solved. Going to clean out unnecessary code before I push that up, and then try to fix the errors. I'll be back to this next week.

PR #1352, opened 2025-04-11
Enable pylint

TimPansino:
# Overview

* Enable `pylint` rules in `ruff`.
  * Disable `pylint` recommendations for now, potentially re-enable later.
  * Disable rules for global statements and redefined loop variables.
    * Too many of these to tackle at once, and most are intentional design choices for better or worse.
  * Fix all other `pylint` rule violations.

### DIFF
--- a/newrelic/common/system_info.py
+++ b/newrelic/common/system_info.py
@@ -349,7 +349,6 @@ def gethostname(use_dyno_names=False, dyno_shorten_prefixes=()):
     """
 
     global _nr_cached_hostname
-    global _nr_cached_hostname_lock
 
     if _nr_cached_hostname:
         return _nr_cached_hostname
@@ -371,7 +370,6 @@ def getips():
     """
 
     global _nr_cached_ip_address
-    global _nr_cached_ipaddress_lock
 
     if _nr_cached_ip_address is not None:
         return _nr_cached_ip_address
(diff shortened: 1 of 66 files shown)


newrelic/common/system_info.py
```diff
@@ -349,7 +349,6 @@ def gethostname(use_dyno_names=False, dyno_shorten_prefixes=()):
     """
 
     global _nr_cached_hostname
-    global _nr_cached_hostname_lock
```
hmstepanek, line 352: Why remove these?

newrelic/hooks/messagebroker_confluentkafka.py
```diff
@@ -76,7 +76,7 @@ def wrap_Producer_produce(wrapped, instance, args, kwargs):
         except Exception as error:
             # Unwrap kafka errors
             while hasattr(error, "exception"):
-                error = error.exception  # pylint: disable=E1101
+                error = error.exception
```
hmstepanek, line 79: I have reviewed to here-will review the rest tomorrow.

newrelic/common/system_info.py
TimPansino: [PLW0602 - global-variable-not-assigned](https://docs.astral.sh/ruff/rules/global-variable-not-assigned/)

There's no need to make this variable global when it's never assigned to. It already reads the globally scoped variable when attempting to use the lock.

hmstepanek (approved):

PR #1353, opened 2025-04-11
Enable flake8-comprehensions

TimPansino:
# Overview

* Enable `flake8-comprehensions` linter in `ruff`.
* Fix all rule violations for `flake8-comprehensions`.


tests/datastore_firestore/test_query.py
```diff
@@ -173,9 +173,9 @@ def exercise_collection_group(client, collection, patch_partition_queries):
     def _exercise_collection_group():
         collection_group = client.collection_group(collection.id)
         assert len(collection_group.get())
-        assert len([d for d in collection_group.stream()])
+        assert len(list(collection_group.stream()))
```
hmstepanek, line 176: Can we just use the thruthiness of lists here instead of calling len on all of these where we aren't checking for a specific length?
```suggestion
        assert list(collection_group.stream())
```

TimPansino: Fixed with the pylint PR

newrelic/api/time_trace.py
```diff
@@ -68,7 +68,7 @@ def _is_leaf(self):
         return self.child_count == len(self.children)
 
     def __repr__(self):
-        return f"<{self.__class__.__name__} object at 0x{id(self):x} {dict(name=getattr(self, 'name', None))}>"
+        return f"<{self.__class__.__name__} object at 0x{id(self):x} { {'name': getattr(self, 'name', None)} }>"
```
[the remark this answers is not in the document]
TimPansino: If you want in a separate change sometime you can pull all those out onto a separate line and assign it to a variable to make it more readable.

hmstepanek (approved):

PR #1354, opened 2025-04-11
Enable pyflakes

TimPansino:
# Overview

* Enable `pyflakes` linter in `ruff`.
* Fix all `pyflakes` violations.
* Correct issues found in `openai`, `kombu`, and `psycopg2` instrumentation (noticed due to linter violations).


newrelic/hooks/database_psycopg2.py
```diff
@@ -63,6 +63,10 @@ class ConnectionSaveParamsWrapper(DBAPI2ConnectionWrapper):
 
     def __enter__(self):
         transaction = current_transaction()
+        if not transaction:
```
hmstepanek, line 66: It's surprising to me that this has been wrong for so long but I don't actually see in the existing code where it is gated by this check higher up so this seems correct to me.

hmstepanek (approved):

Issue #1347, opened 2025-04-07
`TypeError` from inside the `inspect` module after upgrading to 10.8.1


renanvieira:
## Description
After upgrading to 10.8.1, the app started to raise `TypeError` from the inspect module when calling the celery task function with `delay` function.

The message error: 
```
missing a required argument: 'body'
```

This only occurs with celery task functions when calling `.delay()`

- 10.8.0 because of the import bug introduced, I'm not sure if the error happens in this version, given the changelog, its likely it was introduced here.
- 10.7.0 works perfectly.

## Expected Behavior
I expected that newrelic would track the background task successfully,

## [Troubleshooting](https://forum.newrelic.com/s/hubtopic/aAX8W0000008bSoWAI/troubleshooting-frameworks) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results
Redacted stack trace:
```python
  File "<module>/celery.py", line 54, in <background_task>
    <celery_task>.delay(
  File "celery/app/task.py", line 444, in delay
    return self.apply_async(args, kwargs)
  File "celery/app/task.py", line 594, in apply_async
    return app.send_task(
  File "newrelic/hooks/application_celery.py", line 202, in wrap_Celery_send_task
    return wrapped(*args, **kwargs)
  File "celery/app/base.py", line 801, in send_task
    amqp.send_task_message(P, name, message, **options)
  File "celery/app/amqp.py", line 518, in send_task_message
    ret = producer.publish(
  File "newrelic/hooks/messagebroker_kombu.py", line 51, in wrap_Producer_publish
    bound_args = bind_args(wrapped, args, kwargs)
  File "newrelic/common/signature.py", line 20, in bind_args
    bound_args = Signature.from_callable(func).bind(*args, **kwargs)
  File "inspect.py", line 3277, in bind
    return self._bind(args, kwargs)
  File "inspect.py", line 3190, in _bind
    raise TypeError(msg) from None
```

## Steps to Reproduce
1. Install the Python Agent 10.8.1
2. Call a celery task function using the `.delay()` function

## Your Environment
- Python 3.12.9
- Celery 5.4.0
- Newrelic 10.8.1

## Additional context
N/A

m4rcelofs:
Seems related: https://github.com/newrelic/newrelic-python-agent/issues/1339#issuecomment-2779298484

hmstepanek:
[...]

PR #1359, opened 2025-04-17
Fix kombu crash when using sentry & setting parsing

hmstepanek:
# Overview
Because Sentry applies a proxy wrapper around the `Producer.publish` method, when we tried to apply the `bind_args`, it did not pickup the wrapped signature correctly-the instance was missing and so the value for the `body` was set as the value of `self`. Instead use a hardcoded bind to avoid this issue.

Fix missing parsing of config file instrumentation settings.

This is a fix for https://github.com/newrelic/newrelic-python-agent/issues/1347.


newrelic/hooks/messagebroker_kombu.py
```diff
@@ -42,13 +42,57 @@
...
+
+
 def wrap_Producer_publish(wrapped, instance, args, kwargs):
     transaction = current_transaction()
 
     if transaction is None:
         return wrapped(*args, **kwargs)
 
-    bound_args = bind_args(wrapped, args, kwargs)
+    bound_args = bind_publish(*args, **kwargs)
```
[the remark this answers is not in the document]
hmstepanek: Just to note-doesn't look like we can use this because it doesn't have an `__wrapped__` or other attribute to trigger unwrapping against.

TimPansino, line 95:
```suggestion
    try:
        bound_args = bind_publish(*args, **kwargs)
    except Exception:
        bound_args = None
    
    if not bound_args:
        return wrapped(*args, **kwargs)
```

I'm worried this will crash on other versions of Kombu since we only test the latest. Can we wrap it in a try/except clause like so?

This could use an exception message logged to debug as well.

newrelic/hooks/messagebroker_kombu.py
```diff
@@ -42,13 +42,64 @@
...
 
     if transaction is None:
         return wrapped(*args, **kwargs)
 
-    bound_args = bind_args(wrapped, args, kwargs)
+    try:
+        bound_args = bind_publish(*args, **kwargs)
+    except Exception:
+        _logger.debug(
+            "Unable to bind publish arguments. Report this issue to New Relic support.", record_exception=True
```
TimPansino, line 99:
```suggestion
            "Unable to bind arguments for kombu.messaging.Producer.publish. Report this issue to New Relic support.", record_exception=True
```

Make this slightly more specific.

TimPansino (approved):

tox.ini
```diff
@@ -165,8 +165,7 @@ envlist =
     python-template_jinja2-py37-jinja2030103,
     python-template_mako-{py37,py38,py39,py310,py311,py312,py313},
     rabbitmq-messagebroker_pika-{py37,py38,py39,py310,py311,py312,py313,pypy310}-pikalatest,
-    ;; Comment out Kombu until we can root cause the crash.
-    ; rabbitmq-messagebroker_kombu-{py38,py39,py310,py311,py312,py313,pypy310}-kombulatest,
+    rabbitmq-messagebroker_kombu-{py38,py39,py310,py311,py312,py313,pypy310}-{kombulatest,kombu050204},
```
TimPansino, line 168:
```suggestion
    rabbitmq-messagebroker_kombu-{py38,py39,py310,py311,py312,py313,pypy310}-kombulatest,
    rabbitmq-messagebroker_kombu-{py38,py39,py310,py311,pypy310}-kombu050204,
```

Python 3.11 and higher were unsupported on this version, but seems it works fine on Python 3.11. 3.12 and 3.13 are failing so we should remove those.

TimPansino (approved):

PR #1362, opened 2025-04-21
Async Elasticsearch Instrumentation

TimPansino:
# Overview

* Continuing #1309 from @fagnerzulin 
  * Rebased with main post formatter/linter changes, required commit history rewrite.
  * Misc formatting and linter changes from the new tools
* Fixup testing issues


TimPansino:
@fagnerzulin That was more rabbit holes than I was expecting but I believe I've fixed every last issue with this, so it should be out in the next release.

tests/datastore_elasticsearch/test_async_mget.py
```diff
@@ -0,0 +1,150 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+try:
+    from elastic_transport import RoundRobinSelector
+    from elasticsearch import AsyncElasticsearch
+except ImportError:
+    from elasticsearch._async.client import AsyncElasticsearch
```
hmstepanek, line 21: Can we put a comment here explaining this try-catch?

tests/datastore_elasticsearch/test_async_transport.py
```diff
@@ -0,0 +1,75 @@
...
+    from elastic_transport._node._http_httpx import HttpxAsyncHttpNode
+
+
+IS_V8 = ES_VERSION >= (8,)
+IS_V7 = ES_VERSION >= (7,) and ES_VERSION < (8, 0)
+IS_BELOW_V7 = ES_VERSION < (7,)
+
+RUN_IF_V8 = pytest.mark.skipif(IS_V7 or IS_BELOW_V7, reason="Only run for v8+")
+RUN_IF_V7 = pytest.mark.skipif(IS_V8 or IS_BELOW_V7, reason="Only run for v7")
+RUN_IF_BELOW_V7 = pytest.mark.skipif(not IS_BELOW_V7, reason="Only run for versions below v7")
```
hmstepanek, line 39: Are these BELOW_V7's even reachable since we are only running on elasticsearch>=7? (There's a bunch of them though I'm only commenting on one of them.)

tests/datastore_elasticsearch/conftest.py
```diff
@@ -39,8 +40,20 @@
 ES_URL = f"http://{ES_SETTINGS['host']}:{ES_SETTINGS['port']}"
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="function")
```
hmstepanek, line 43: Maybe replace `ES_VERSION = tuple([int(n) for n in get_package_version("elasticsearch").split(".")])` with the following while we're in here?
```
ES_VERSION = get_package_version_tuple("elasticsearch")
```

hmstepanek (changes requested):
Just a couple things-otherwise looks good.

hmstepanek (approved):

PR #1386, opened 2025-05-12
Fix pymssql Tests

TimPansino:
# Overview

* `pymssql` updated to only support Python 3.9+. Pin older Python versions to a compatible version of `pymssql`.
  * v2.3.2 lists support for Python 3.7 and 3.8, but doesn't have prebuilt wheels and fails to build. v2.3.1 is used instead.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -62,7 +62,8 @@ envlist =
     mongodb3-datastore_pymongo-{py37,py38,py39,py310,py311,py312}-pymongo03,
     mongodb8-datastore_pymongo-{py37,py38,py39,py310,py311,py312,py313,pypy310}-pymongo04,
     mysql-datastore_aiomysql-{py37,py38,py39,py310,py311,py312,py313,pypy310},
-    mssql-datastore_pymssql-{py37,py38,py39,py310,py311,py312,py313},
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -288,7 +289,8 @@ deps =
     datastore_motor-motorlatest: tornado
     datastore_pymongo-pymongo03: pymongo<4.0
     datastore_pymongo-pymongo04: pymongo<5.0
-    datastore_pymssql: pymssql
+    datastore_pymssql-pymssqllatest: pymssql
+    datastore_pymssql-pymssql020301: pymssql==2.3.1
```
lrafeei, line 293: From the release notes, it seems to indicate that support for Python 3.7 and 3.8 was dropped in this version: https://pymssql.readthedocs.io/en/latest/changelog.html#version-2-3-1-2024-08-25-mikhail-terekhov

I was going to suggest pinning this to the previous version instead of this one.  However, if this works, we can keep it as is

TimPansino: They dropped MacOS wheels for Python 3.7 & 3.8 in that version, they dropped the linux ones in the next version.

lrafeei (approved):

PR #1394, opened 2025-05-28
Deduplicate Django Hooks Files

TimPansino:
# Overview

* There was a separate file for Django that includes `_py3` in the name that can now be merged into the original file.


Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-423120

newrelic/hooks/framework_django.py
```diff
@@ -420,6 +416,36 @@ def _nr_wrapper_BaseHandler_get_response_(wrapped, instance, args, kwargs):
     return response
 
 
+async def _nr_wrapper_BaseHandler_get_response_async_(wrapped, instance, args, kwargs):
+    response = await wrapped(*args, **kwargs)
+
+    if current_transaction() is None:
+        return response
+
+    request = _bind_get_response(*args, **kwargs)
```
umaannamalai, line 425: Should we swap this to a `bind_args` call and drop the `_bind_get_response` function definition in this file?

TimPansino: I'm inclined to leave it this way since the simple bind function is working still and is much lighter weight than the full bind_args function is.

umaannamalai (approved):

PR #1395, opened 2025-05-28
Azure function instrumentation

lrafeei:
This PR provides instrumentation for Azure Functions with HTTP triggers


newrelic/common/utilization.py
```diff
@@ -207,6 +207,47 @@ class AzureUtilization(CommonUtilization):
...
+
+    @staticmethod
+    def fetch():
+        cloud_region = os.environ.get("REGION_NAME")
+        website_owner_name = os.environ.get("WEBSITE_OWNER_NAME")
+        azure_function_app_name = os.environ.get("WEBSITE_SITE_NAME")
+
+        if all((cloud_region, website_owner_name, azure_function_app_name)):
+            if website_owner_name.endswith("-Linux"):
+                resource_group_name = re.search(r"\+([a-zA-z0-9\-]+)-[a-zA-Z0-9]+(?:-Linux)", website_owner_name).group(
```
TimPansino:
```suggestion
                resource_group_name = re.search(r"\+([a-zA-Z0-9\-]+)-[a-zA-Z0-9]+(?:-Linux)", website_owner_name).group(
```

newrelic/common/utilization.py
```diff
@@ -207,6 +207,47 @@ class AzureUtilization(CommonUtilization):
...
+                    1
+                )
+            else:
+                resource_group_name = re.search(r"\+([a-zA-z0-9\-]+)-[a-zA-Z0-9]+", website_owner_name).group(1)
```
TimPansino:
```suggestion
                resource_group_name = re.search(r"\+([a-zA-Z0-9\-]+)-[a-zA-Z0-9]+", website_owner_name).group(1)
```

newrelic/core/config.py
```diff
@@ -904,6 +904,14 @@ def default_otlp_host(host):
...
 )
 
+# celeryev is the monitoring queue for rabbitmq which we do not need to monitor-it just makes a lot of noise.
+_settings.instrumentation.kombu.ignored_exchanges = parse_space_separated_into_list(
+    os.environ.get("NEW_RELIC_INSTRUMENTATION_KOMBU_IGNORED_EXCHANGES", "celeryev")
+)
+_settings.instrumentation.kombu.consumer.enabled = _environ_as_bool(
+    "NEW_RELIC_INSTRUMENTATION_KOMBU_CONSUMER_ENABLED", default=False
+)
+
```
TimPansino: Is this related to azure functions?

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+
+    # Force default registration of the application instance
+    # instead of lazy registration upon the first request
+    app_name = os.environ.get("NEW_RELIC_APP_NAME", os.environ.get("WEBSITE_SITE_NAME", None))
+    application = application_instance(app_name, activate=False)
+    if application and not application.active:
+        application.activate()
+    elif not application:
+        application = application_instance(app_name)
+        application.activate()
```
TimPansino: Why can't this just be `application = application_instance(app_name)`? Doesn't that handle activating and returning the application?

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+    if application and not application.active:
+        application.activate()
+    elif not application:
+        application = application_instance(
+            os.environ.get("NEW_RELIC_APP_NAME", os.environ.get("WEBSITE_SITE_NAME", None))
+        )
+        application.activate()
+
+    http_request = None
+    for key, value in params.items():
```
TimPansino:
```suggestion
    for value in params.values():
```

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+    application = application_instance(
+        os.environ.get("NEW_RELIC_APP_NAME", os.environ.get("WEBSITE_SITE_NAME", None)), activate=False
+    )
+    if application and not application.active:
+        application.activate()
+    elif not application:
+        application = application_instance(
+            os.environ.get("NEW_RELIC_APP_NAME", os.environ.get("WEBSITE_SITE_NAME", None))
+        )
+        application.activate()
```
TimPansino: Same comment as above about activation.

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+
+        website_owner_name = os.environ.get("WEBSITE_OWNER_NAME", None)
+        if not website_owner_name:
+            subscription_id = "Unknown"
+        else:
+            subscription_id = re.search(r"(?:(?!\+).)*", website_owner_name) and re.search(
+                r"(?:(?!\+).)*", website_owner_name
+            ).group(0)
+        if website_owner_name and website_owner_name.endswith("-Linux"):
+            resource_group_name = re.search(r"\+([a-zA-z0-9\-]+)-[a-zA-Z0-9]+(?:-Linux)", website_owner_name).group(
```
TimPansino:
```suggestion
            resource_group_name = re.search(r"\+([a-zA-Z0-9\-]+)-[a-zA-Z0-9]+(?:-Linux)", website_owner_name).group(
```

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+                1
+            )
+        elif website_owner_name:
+            resource_group_name = re.search(r"\+([a-zA-z0-9\-]+)-[a-zA-Z0-9]+", website_owner_name).group(1)
```
TimPansino:
```suggestion
            resource_group_name = re.search(r"\+([a-zA-Z0-9\-]+)-[a-zA-Z0-9]+", website_owner_name).group(1)
```

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+    if application and not application.active:
+        application.activate()
+    elif not application:
+        application = application_instance(
+            os.environ.get("NEW_RELIC_APP_NAME", os.environ.get("WEBSITE_SITE_NAME", None))
+        )
+        application.activate()
+
+    http_request = None
+    for key, value in params.items():
```
TimPansino:
```suggestion
    for value in params.values():
```

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+    application = application_instance(
+        os.environ.get("NEW_RELIC_APP_NAME", os.environ.get("WEBSITE_SITE_NAME", None)), activate=False
+    )
+    if application and not application.active:
+        application.activate()
+    elif not application:
+        application = application_instance(
+            os.environ.get("NEW_RELIC_APP_NAME", os.environ.get("WEBSITE_SITE_NAME", None))
+        )
+        application.activate()
```
TimPansino: Same comment about activation.

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+
+        website_owner_name = os.environ.get("WEBSITE_OWNER_NAME", None)
+        if not website_owner_name:
+            subscription_id = "Unknown"
+        else:
+            subscription_id = re.search(r"(?:(?!\+).)*", website_owner_name) and re.search(
+                r"(?:(?!\+).)*", website_owner_name
+            ).group(0)
+        if website_owner_name and website_owner_name.endswith("-Linux"):
+            resource_group_name = re.search(r"\+([a-zA-z0-9\-]+)-[a-zA-Z0-9]+(?:-Linux)", website_owner_name).group(
```
TimPansino:
```suggestion
            resource_group_name = re.search(r"\+([a-zA-Z0-9\-]+)-[a-zA-Z0-9]+(?:-Linux)", website_owner_name).group(
```

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+                1
+            )
+        elif website_owner_name:
+            resource_group_name = re.search(r"\+([a-zA-z0-9\-]+)-[a-zA-Z0-9]+", website_owner_name).group(1)
```
TimPansino:
```suggestion
            resource_group_name = re.search(r"\+([a-zA-Z0-9\-]+)-[a-zA-Z0-9]+", website_owner_name).group(1)
```

newrelic/common/utilization.py
TimPansino: This particular line is present in 2 different files and at least 6 places, but it always uses the exact same regex. Can we run `re.compile()` on this one time in this module and import it later in the hooks file?

setup.py
```diff
@@ -164,7 +160,7 @@ def build_extension(self, ext):
         "git_describe_command": "git describe --dirty --tags --long --match *.*.*",
         "write_to": "newrelic/version.txt",
     },
-    setup_requires=["setuptools_scm>=3.2,<9"],
+    setup_requires=["setuptools_scm>=3.2,<8.1"],
```
TimPansino: What was this change for?

tests/agent_unittests/test_agent_protocol.py
```diff
@@ -341,26 +350,30 @@ def connect_payload_asserts(
...
+        (True, True, True, True, True, True, True, False, True),
+        (True, True, True, True, True, True, False, True, True),
     ],
 )
-def test_connect(with_aws, with_ecs, with_pcf, with_gcp, with_azure, with_docker, with_kubernetes, with_ip):
-    global AWS, AZURE, GCP, PCF, DOCKER, KUBERNETES, IP_ADDRESS
+def test_connect(
+    with_aws, with_ecs, with_pcf, with_gcp, with_azure, with_azurefunction, with_docker, with_kubernetes, with_ip
+):
+    global AWS, AZURE, AZUREFUNCTION, GCP, PCF, BOOT_ID, DOCKER, KUBERNETES, IP_ADDRESS
```
TimPansino:
```suggestion
    global AWS, AZURE, AZUREFUNCTION, GCP, PCF, DOCKER, KUBERNETES, IP_ADDRESS
```

tests/serverless_azure/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+)
+
+_default_settings = {
+    "package_reporting.enabled": False,  # Turn off package reporting for testing as it causes slow downs.
+    "transaction_tracer.explain_threshold": 0.0,
+    "transaction_tracer.transaction_threshold": 0.0,
+    "transaction_tracer.stack_trace_threshold": 0.0,
+    "debug.log_data_collector_payloads": True,
+    "debug.record_transaction_failure": True,
+    "host": "staging-collector.newrelic.com",
```
TimPansino: Why is this hard coded?

tests/serverless_azure/sample_application/function_app.py
```diff
@@ -0,0 +1,34 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import newrelic.agent
+
+newrelic.agent.initialize()  # Initialize the New Relic agent
+app_name = "Python Agent Test (serverless_azure)"
+application = newrelic.agent.register_application(app_name)
+
+import azure.functions as func
```
TimPansino:
```suggestion
import azure.functions as func  # noqa: E402
```

tox.ini
```diff
@@ -187,6 +188,8 @@ deps =
     py313: legacy-cgi==2.6.1  # cgi was removed from the stdlib in 3.13, and is required for WebTest
 
     # Test Suite Dependencies
+    serverless_azure: azure-functions
+    serverless_azure: requests
```
TimPansino:
```suggestion
    azurefunctions-serverless_azure: azure-functions
    azurefunctions-serverless_azure: requests
```

TimPansino: These are also unsorted, can we sort these lines into the existing structure?

tox.ini
```diff
@@ -446,6 +449,7 @@ setenv =
     framework_tornado: LDFLAGS=-L/usr/local/opt/openssl/lib
     framework_tornado: CPPFLAGS=-I/usr/local/opt/openssl/include
 
+
```
TimPansino:
```suggestion
```

tox.ini
```diff
@@ -462,7 +466,10 @@ commands =
     framework_tornado: pip install --ignore-installed --config-settings="--build-option=--with-openssl" pycurl
     coverage run -m pytest -v []
 
-allowlist_externals={toxinidir}/.github/scripts/*
+allowlist_externals =
+    {toxinidir}/.github/scripts/*
+    cp
+    mkdir
```
TimPansino: What's this change for? I don't see any reference to these commands

tox.ini
```diff
@@ -526,6 +533,7 @@ changedir =
     external_urllib3: tests/external_urllib3
     framework_aiohttp: tests/framework_aiohttp
     framework_ariadne: tests/framework_ariadne
+    serverless_azure: tests/serverless_azure
     framework_bottle: tests/framework_bottle
```
TimPansino: This needs sorted

.github/workflows/tests.yml
```diff
@@ -1466,3 +1467,61 @@
...
+        env:
+          TOX_PARALLEL_NO_SPINNER: 1
+          PY_COLORS: 0
+
+      - name: Upload Coverage Artifacts
+        uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3 # 4.3.1
+        with:
+          name: coverage-${{ github.job }}-${{ strategy.job-index }}
+          path: ./**/.coverage.*
+          retention-days: 1
```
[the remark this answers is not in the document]
TimPansino: These permissions should be good. I would add these in while you're making changes. I've got permissions added to the other test workflows in another PR.

newrelic/api/transaction.py
```diff
@@ -566,6 +566,9 @@ def __exit__(self, exc, value, tb):
             for ml_model, version in self._ml_models:
                 self.record_custom_metric(f"Supportability/Python/ML/{ml_model}/{version}", 1)
 
+        if os.environ.get("FUNCTIONS_WORKER_RUNTIME", None):
+            self.record_custom_metric("Supportability/Python/AzureFunctionMode/Enabled", 1)
```
umaannamalai: Was there a reason why this supportability metric was being recorded once per transaction vs. once when we [connect to the data collector](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/core/application.py#L530)?

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+            response = await wrapped(*args, **kwargs)
+            return response
+
+    return await wrapped(*args, **kwargs)
+
+
+def wrap_dispatcher__run_sync_func(wrapped, instance, args, kwargs):
+    from azure.functions.http import HttpRequest
+
+    def bind_params(invocation_id, context, func, params, *args, **kwargs):
```
umaannamalai: Is it possible to swap all of the function-specific binds in here to use the auto sig `bind_args` function instead just in case function signatures change down the road?

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+# Right now, we only support HTTP, so this function serves to activate
+# the application if not already registered with the collector as well
+# as determining if this invocation was a cold start or not.
+async def wrap_dispatcher__handle__invocation_request(wrapped, instance, args, kwargs):
+    def bind_params(request, *args, **kwargs):
+        return request
+
+    # Force default registration of the application instance
+    # instead of lazy registration upon the first request
+    app_name = os.environ.get("NEW_RELIC_APP_NAME", os.environ.get("WEBSITE_SITE_NAME", None))
```
umaannamalai: Wondering if we could this application activation logic into a separate function since it seems to be used a few times across wrappers.

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+
+        if hasattr(instance, "_nr_cold_start"):
+            cold_start = True
+            # Delete the attribute so that subsequent calls to this
+            # method are noted as not being cold starts
+            del instance._nr_cold_start
+        else:
+            cold_start = False
+
+        website_owner_name = os.environ.get("WEBSITE_OWNER_NAME", None)
```
umaannamalai: Maybe this logic to grab the subscription ID and resource group name could also be pulled out into a separate method since it seems to be reused in the async wrapper as well

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+        application = application_instance(
+            os.environ.get("NEW_RELIC_APP_NAME", os.environ.get("WEBSITE_SITE_NAME", None))
+        )
+        application.activate()
+
+    http_request = None
+    for key, value in params.items():
+        if isinstance(value, HttpRequest):
+            http_request = value
+            url_split = urlparse.urlsplit(http_request.url)
```
umaannamalai: Maybe we could use `getattr` calls here to grab attributes instead of direct access to avoid attribute errors (by potentially also specifying a default value)

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+            "faas.name": faas_name,
+            "faas.trigger": trigger_type,
+            "faas.invocation_id": context.invocation_id,
+        }
+
+        # Only add this attribute if this is a cold start
+        if cold_start:
+            azure_intrinsics.update({"faas.coldStart": True})
+
+        if not transaction:
```
umaannamalai: Could this exit condition be moved further up before all of the parsing logic to grab data for the intrinsic attributes? To avoid running that code if we don't have a transaction in the first place

newrelic/hooks/serverless_azure.py
```diff
@@ -0,0 +1,298 @@
...
+    return response
+
+
+def wrap_httpresponse__init__(wrapped, instance, args, kwargs):
+    def bind_params(body=b"", status_code=None, headers=None, *args, **kwargs):
+        return status_code, headers
+
+    transaction = current_transaction()
+    if not transaction:
+        response = wrapped(*args, **kwargs)
```
umaannamalai: Can we return `wrapped` directly instead of storing it in a variable first?

tests/serverless_azure/test_application.py
```diff
@@ -0,0 +1,60 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import requests
+import subprocess
+import socket
+import time
+
```
umaannamalai: Could we expand testing here to validate the new Azure intrinsic agent attributes and transaction metrics?

tests/serverless_azure/sample_application/function_app.py
```diff
@@ -0,0 +1,34 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import newrelic.agent
+
+newrelic.agent.initialize()  # Initialize the New Relic agent
```
umaannamalai: Is this manual agent initialization required? The `collector_agent_registration_fixture` does an `initialize` call also.

newrelic/hooks/serverless_azure.py
lrafeei: I think this was the result of trying to get this to work in case the user set the app name in a config file, but I have since found out that the config file does not play nicely with Azure Function Apps anyway.  Removed.

newrelic/core/config.py
lrafeei: I think these lines somehow got duplicated during the merging from main process...

tests/serverless_azure/test_application.py
lrafeei: This is the part that is a bit iffy; to test this I have to use the subprocess module which uses a different process, and the validators only work with the main process (it is the same issue that gunicorn tests have)

tests/serverless_azure/sample_application/function_app.py
lrafeei: This is a weird situation in that the collector_agent_registration fixture initializes the agent in a different process than the one we are testing in.

tox.ini
lrafeei: Whoops, this is from an old commit that should have disappeared

tests/agent_unittests/test_agent_protocol.py
lrafeei: Is that no longer a test we are running?  I assume all of these need to be taken out?  Or just from the list of globals?

tests/serverless_azure/sample_application/function_app.py
TimPansino: This is fixed

tests/serverless_azure/test_application.py
lrafeei: This has been fixed.

.github/workflows/tests.yml
```diff
@@ -24,6 +24,10 @@ on:
   schedule:
     - cron: "0 15 * * *"
 
+permissions:
+  contents: read
+  actions: read
```
TimPansino: Does this need the actions permission? I thought it only needed `contents: read` to function.

lrafeei: I *thought* it needed both but I can just take it out and see

setup.py
TimPansino:
```suggestion
    setup_requires=["setuptools_scm>=3.2,<9"],
```

.github/workflows/tests.yml
TimPansino:
```suggestion
```

newrelic/hooks/framework_azurefunctions.py
```diff
@@ -0,0 +1,254 @@
...
+    app_name = os.environ.setdefault("NEW_RELIC_APP_NAME", os.getenv("WEBSITE_SITE_NAME", ""))
+
+    return application_instance(app_name)
+
+
+def intrinsics_populator(application, context):
+    # For now, only HTTP triggers are supported
+    trigger_type = "http"
+
+    website_owner_name = os.environ.get("WEBSITE_OWNER_NAME", None)
```
umaannamalai, line 37: Perhaps we could throw this whole block into a try-except and log a warning given it looks like there are a few potential failure points here. Or we could ideally minimize the direct attribute access like `context.function_name` and `application.name` and use `getattr` calls instead

newrelic/hooks/framework_azurefunctions.py
```diff
@@ -0,0 +1,254 @@
...
+    # Otherwise, create appropriate type of Transaction in the future
+    if http_request:
+        transaction = WebTransaction(
+            application=application,
+            name=context.function_name,
+            group="AzureFunction",
+            scheme=scheme,
+            host=host,
+            port=port,
+            request_method=http_request.method,
```
umaannamalai: This should be fine since you're checking the instance type of the `http_request` but just want to ensure safe access of the attributes on it to avoid crashes.

newrelic/hooks/framework_azurefunctions.py
lrafeei: Good point--Done!

TimPansino (approved):

PR #1412, opened 2025-06-24
Autogen Agent Instrumentation

umaannamalai:
This PR adds instrumentation for autogen assistant agents and teams.


newrelic/hooks/mlmodel_autogen.py
```diff
@@ -32,6 +58,126 @@ async def wrap_from_server_params(wrapped, instance, args, kwargs):
...
+
+    return llm_metadata_dict
+
+
+def _extract_tool_output(return_val, tool_name):
+    try:
+        output = getattr(return_val[1], "content", None)
+        return output
+    except Exception:
+        _logger.warning(f"Unable to parse tool output value from {tool_name}. Omitting output from LlmTool event.")
```
hmstepanek:
```suggestion
        _logger.warning("Unable to parse tool output value from %s. Omitting output from LlmTool event.", tool_name)
```

newrelic/hooks/mlmodel_autogen.py
```diff
@@ -32,6 +58,126 @@ async def wrap_from_server_params(wrapped, instance, args, kwargs):
...
+        return None
+
+
+def _construct_base_tool_event_dict(bound_args, tool_call_data, tool_id, transaction, settings):
+    try:
+        input = getattr(tool_call_data, "arguments", None)
```
hmstepanek:
```suggestion
        _input = getattr(tool_call_data, "arguments", None)
```

newrelic/hooks/mlmodel_autogen.py
```diff
@@ -32,6 +58,126 @@ async def wrap_from_server_params(wrapped, instance, args, kwargs):
...
+        tool_input = str(input) if input else None
```
hmstepanek:
```suggestion
        tool_input = str(_input) if _input else None
```

newrelic/hooks/mlmodel_autogen.py
```diff
@@ -32,6 +58,126 @@ async def wrap_from_server_params(wrapped, instance, args, kwargs):
...
+
+    return tool_event_dict
+
+
+async def wrap__execute_tool_call(wrapped, instance, args, kwargs):
+    transaction = current_transaction()
+    if not transaction:
+        return await wrapped(*args, **kwargs)
+
+    settings = transaction.settings if transaction.settings is not None else global_settings()
```
hmstepanek:
```suggestion
    settings = transaction.settings or global_settings()
```

newrelic/hooks/mlmodel_autogen.py
```diff
@@ -32,6 +58,126 @@ async def wrap_from_server_params(wrapped, instance, args, kwargs):
...
+        ft.notice_error(attributes={"tool_id": tool_id})
+        ft.__exit__(*sys.exc_info())
+        # If we hit an exception, append the error attribute and duration from the exited function trace
+        tool_event_dict.update({"duration": ft.duration * 1000, "error": True})
+        transaction.record_custom_event("LlmTool", tool_event_dict)
+        raise
+
+    ft.__exit__(None, None, None)
+
+    if not return_val:
```
hmstepanek: Just double checking the logic here is what we actually want: if there is no return value from the tool call we don't capture any tool event info? We don't want to even capture a partial amount of data from the tool call?

tests/mlmodel_autogen/test_assistant_agent.py
```diff
@@ -0,0 +1,256 @@
...
+            "ingest_source": "Python",
+            "error": True,
+            "duration": None,
+        },
+    )
+]
+
+
+# Example tool for testing purposes
+def add_exclamation(input: str) -> str:
```
hmstepanek:
```suggestion
def add_exclamation(message: str) -> str:
```

tests/mlmodel_autogen/test_assistant_agent.py
```diff
@@ -0,0 +1,256 @@
...
+    return f"{input}!"
```
hmstepanek:
```suggestion
    return f"{message}!"
```

newrelic/hooks/mlmodel_autogen.py
umaannamalai: This is how we have done it in our other AI instrumentations though I still do see value in recording the event even when there is no return value. I initially left this to be consistent with OpenAI. However, I am okay to remove that conditional since this instrumentation doesn't have as a heavy of a reliance on the return value.

hmstepanek (approved):

PR #1419, opened 2025-07-01
Core tracing prototype options

hmstepanek:
# Overview
Prototype of Core Tracing
* Add two new configuration settings:
  * `distributed_tracing.drop_inprocess_spans.enabled` - defaulted to False (Only reports entry and exit spans.)
  * `distributed_tracing.unique_spans.enabled` - defaulted to False (Only reports unique exit spans and entry spans.)
  * `distributed_tracing.minimize_attributes.enabled` - defaulted to False (Drops all agent attributes that are not entity synthesis attributes from spans)
    **Note exit spans are defined as spans that contain entity synthesis agent attributes.**
* Report additional metrics:
  * Time to process dropping spans and attributes. Note: attributes are dropped after attr processing intentionally to make time reporting easier.
  * Approximate amount of bytes or "cost" of core tracing/dropping in-process spans.
* Double adaptive sampling rate from default of 10 to 20 when Core Tracing is enabled.

Linking ErrorTraces, slow sql etc to distributed traces is not supported in the prototype.

### DIFF
--- a/newrelic/core/node_mixin.py
+++ b/newrelic/core/node_mixin.py
@@ -192,4 +192,4 @@ class DatastoreNodeMixin(GenericNodeMixin):
         except Exception:
             pass
 
-        return super().span_event(settings, i_attrs=base_attrs, parent_guid=parent_guid, attr_class=attr_class, *args, **kwargs)
+        return super().span_event(settings, base_attrs=i_attrs, parent_guid=parent_guid, attr_class=attr_class, *args, **kwargs)


newrelic/core/node_mixin.py
```diff
@@ -49,37 +52,75 @@ def get_trace_segment_params(self, settings, params=None):
         _params["exclusive_duration_millis"] = 1000.0 * self.exclusive
         return _params
 
-    def span_event(self, settings, base_attrs=None, parent_guid=None, attr_class=dict):
+    def span_event(self, settings, base_attrs=None, parent_guid=None, attr_class=dict, ct_exit_spans=None, ct_processing_time=0):
```
hmstepanek:
```suggestion
    def span_event(self, settings, base_attrs=None, parent_guid=None, attr_class=dict, ct_exit_spans=None, ct_processing_time=None):
```

newrelic/packages/requirements.txt
```diff
@@ -6,3 +6,4 @@ opentelemetry_proto==1.32.1
 urllib3==1.26.19
 wrapt==1.16.0
 asgiref==3.6.0  # We only vendor asgiref.compatibility.py
+objsize==0.7.1
```
hmstepanek:
```suggestion
```

PR #1423, opened 2025-07-07
oracledb Package Instrumentation

### LINKED ISSUE #1420: Instrument python-oracledb as cx_Oracle is obsolete
## Summary

From official Oracle Documentation [cx_Oracle](https://oracle.github.io/python-cx_Oracle/) was bsoleted in favour of [python-oracledb](https://oracle.github.io/python-oracledb/)

## Desired Behavior

New python-oracledb instrumented by default

## Additional context

- Latest Django 5.0 [drops support for cx_Oracle](https://docs.djangoproject.com/en/5.2/ref/databases/#oracle-notes) in favor of python-oracledb

TimPansino:
# Overview

* Add instrumentation for [oracledb](https://pypi.org/project/oracledb/).
  * Instruments both sync and async connections through DBAPI2 when using `connect()` or `connect_async()`.
  * Attempts to parse instance info from multiple sources, falling back to `"unknown"`. 

# Related

Closes #1420

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -67,7 +67,9 @@ envlist =
     mysql-datastore_mysql-mysqllatest-{py37,py38,py39,py310,py311,py312,py313},
     mysql-datastore_mysqldb-{py38,py39,py310,py311,py312,py313},
     mysql-datastore_pymysql-{py37,py38,py39,py310,py311,py312,py313,pypy310},
-    oracledb-datastore_oracledb-{py39,py310,py311,py312,py313},
+    oracledb-datastore_oracledb-{py39,py310,py311,py312,py313}-oracledblatest,
+    oracledb-datastore_oracledb-{py39,py313}-oracledb02,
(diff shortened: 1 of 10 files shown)


tox.ini
```diff
@@ -279,6 +280,7 @@ deps =
     datastore_mysql: protobuf<4
     # mysqlclient is the Python 3 replacement for MySQLdb
     datastore_mysqldb: mysqlclient
+    datastore_oracledb: oracledb
```
hmstepanek: Do we want to support/test on older versions at all?

hmstepanek (approved):
Looks good, just had one question about whether we want to support/test on any older versions than latest.

tox.ini
TimPansino: That's fair, I'll add some older ones to tox.

hmstepanek (approved):

Issue #1045, opened 2024-01-24
Error in NewRelic on publishing to kafka in Python


mishikaraj:
**Description**
 While publishing message to apache kafka , getting error with Newrelic Message Transaction  - **Error 'MessageTransaction' object has no attribute 'destination_name'**
**code** 
```
producer = KafkaProducer(
            bootstrap_servers=bootstrap_servers,
            value_serializer=lambda x: json.dumps(x).encode("utf-8"),
        )
producer.send(topic, value=event_data).get()
```
**Error Stack Trace**
```
File "/Users/mishika/supply-metrics/app/kafka_manager/base_consumer_v2.py", line 55, in produce
    self.producer.send(topic, value=event_data, key=key).get()
  File "/Users/mishika/miniconda3/envs/myenv/lib/python3.10/site-packages/newrelic/hooks/messagebroker_kafkapython.py", line 68, in wrap_KafkaProducer_send
    return wrapped(
  File "/Users/mishika/miniconda3/envs/myenv/lib/python3.10/site-packages/kafka/producer/kafka.py", line 581, in send
    value_bytes = self._serialize(
  File "/Users/mishika/miniconda3/envs/myenv/lib/python3.10/site-packages/kafka/producer/kafka.py", line 714, in _serialize
    return f(data)
  File "/Users/mishika/miniconda3/envs/myenv/lib/python3.10/site-packages/newrelic/hooks/messagebroker_kafkapython.py", line 203, in _wrap_serializer
    topic = transaction.destination_name
AttributeError: 'MessageTransaction' object has no attribute 'destination_name'
```
**newrelic.ini file**
```
[newrelic]
log_level = debug
high_security = false
transaction_tracer.enabled = true
transaction_tracer.transaction_threshold = apdex_f
transaction_tracer.record_sql = obfuscated
transaction_tracer.stack_trace_threshold = 0.5
transaction_tracer.explain_enabled = true
transaction_tracer.explain_threshold = 0.5
transaction_tracer.function_trace =
error_collector.enabled = true
error_collector.ignore_errors = pycommon.exceptions.common_exceptions:DuplicateError rest_framework.exceptions:ValidationError rest_framework.exceptions:NotFound rest_framework.exceptions:ParseError
browser_monitoring.auto_instrument = true
thread_profiler.enabled = true
distributed_tracing.enabled = false
app_name = 
monitor_mode = true
license_key =  
```

**Expected Behavior**
Expected message should have sent to kafka topic

[...]

PR #1425, opened 2025-07-15
Add destination_name attr to all MessageTransactions

### LINKED ISSUE #1363: 'MessageTransaction' object has no attribute 'destination_name'
upgrading agent version from 8.8.0 to 10.8.1 and got error saying 'MessageTransaction' object has no attribute 'destination_name'
this i am getting while creating a celery (5.2.7) task. it does not happen on every celery task publish. not able to reproduce it in staging though

hmstepanek:
# Overview
Fix a bug in messagebroker instrumentations where the serialize methods were looking for the `destination_name` attribute on the MessageTransaction and it did not exist.

# Related Github Issue
Fixes #1363 and #1045.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ pip install newrelic
 Start](https://docs.newrelic.com/docs/agents/python-agent/getting-started/python-agent-quick-start).)
 
 1. Generate the agent configuration file with your [license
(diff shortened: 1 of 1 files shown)


TimPansino (approved):

JestemStefan:
Thank you for fixing this!

When we can expect this to be released?

PR #1428, opened 2025-07-23
Add Converse API non-streaming support.

umaannamalai:
This PR adds support for the AWS Bedrock `Converse` operation.

### DIFF
--- a/newrelic/hooks/datastore_elasticsearch.py
+++ b/newrelic/hooks/datastore_elasticsearch.py
@@ -270,7 +270,6 @@ _elasticsearch_client_methods_v8 = (
     ("msearch_template", _extract_args_search_templates_index),
     ("mtermvectors", _extract_args_index),
     ("open_point_in_time", _extract_args_index),
-    ("options", None),
     ("ping", None),
     ("put_script", None),
     ("rank_eval", _extract_args_requests_index),
--- a/tests/datastore_elasticsearch/test_async_elasticsearch.py
+++ b/tests/datastore_elasticsearch/test_async_elasticsearch.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from conftest import ES_SETTINGS, IS_V8_OR_ABOVE
+from conftest import ES_SETTINGS, IS_V8_OR_ABOVE, RUN_IF_V8_OR_ABOVE
 from elasticsearch._async import client
 from testing_support.fixture.event_loop import event_loop as loop
 from testing_support.fixtures import override_application_settings
@@ -215,3 +215,20 @@ def test_async_elasticsearch_operation_enabled_empty_transaction_settings(async_
 
 def test_async_elasticsearch_no_transaction(async_client, loop):
     loop.run_until_complete(_exercise_es(async_client))
+
+
+@RUN_IF_V8_OR_ABOVE
+@background_task()
+def test_async_elasticsearch_options_no_crash(async_client, loop):
+    """Test that the options method on the async client doesn't cause a crash when run with the agent"""
+
+    async def _test():
+        client_with_auth = async_client.options(basic_auth=("username", "password"))
+        assert client_with_auth is not None
+        assert client_with_auth != async_client
+
+        # If options was instrumented, this would cause a crash since the first call would return an unexpected coroutine
+        client_chained = async_client.options(basic_auth=("user", "pass")).options(request_timeout=60)
+        assert client_chained is not None
+
+    loop.run_until_complete(_test())
--- a/tests/datastore_elasticsearch/test_async_instrumented_methods.py
+++ b/tests/datastore_elasticsearch/test_async_instrumented_methods.py
@@ -75,7 +75,7 @@ def test_method_on_async_client_datastore_trace_inputs(
 
 def _test_methods_wrapped(_object, ignored_methods=None):
     if not ignored_methods:
-        ignored_methods = {"perform_request", "transport"}
+        ignored_methods = {"perform_request", "transport", "options"}
 
     def is_wrapped(m):
         return hasattr(getattr(_object, m), "__wrapped__")
--- a/tests/datastore_elasticsearch/test_instrumented_methods.py
+++ b/tests/datastore_elasticsearch/test_instrumented_methods.py
@@ -84,7 +84,7 @@ def test_method_on_client_datastore_trace_inputs(client, sub_module, method, arg
 
 def _test_methods_wrapped(_object, ignored_methods=None):
     if not ignored_methods:
-        ignored_methods = {"perform_request", "transport"}
+        ignored_methods = {"perform_request", "transport", "options"}
 
     def is_wrapped(m):
         return hasattr(getattr(_object, m), "__wrapped__")


tests/external_botocore/conftest.py
```diff
@@ -54,6 +55,11 @@
 BEDROCK_AUDIT_LOG_FILE = Path(__file__).parent / "bedrock_audit.log"
 BEDROCK_AUDIT_LOG_CONTENTS = {}
 
+BEDROCK_CONVERSE_AUDIT_LOG_FILE = os.path.join(
```
hmstepanek: Due to the work Tim is doing to support windows you'll need to fix these lint errors to use [pathlib](https://docs.python.org/3/library/pathlib.html) instead.

newrelic/hooks/external_botocore.py
```diff
@@ -725,6 +751,113 @@ def _wrap_bedrock_runtime_invoke_model(wrapped, instance, args, kwargs):
...
+                exc, False, model, span_id, trace_id, request_extractor, {}, ft, transaction, kwargs, is_converse=True
+            )
+            raise
+
+        if not response or response_streaming and not settings.ai_monitoring.streaming.enabled:
+            ft.__exit__(None, None, None)
+            return response
+
+        # Let the instrumentation of make_api_call in the aioboto3 client handle it if we have an async case
+        if inspect.iscoroutine(response):
```
hmstepanek, line 815: Do we have a test for this case? We should probably add one if we don't.

newrelic/hooks/external_botocore.py
```diff
@@ -534,10 +536,23 @@ def extract_bedrock_cohere_model_streaming_response(response_body, bedrock_attrs
 
 
 def handle_bedrock_exception(
-    exc, is_embedding, model, span_id, trace_id, request_extractor, request_body, ft, transaction
+    exc, is_embedding, model, span_id, trace_id, request_extractor, request_body, ft, transaction, kwargs, is_converse
 ):
     try:
         bedrock_attrs = {"model": model, "span_id": span_id, "trace_id": trace_id}
+        if is_converse:
```
hmstepanek, line 542: A couple thoughts on the design of this:

I wonder if there's a way you could move this if is_converse block into a request_extractor? Have you explored that at all/do you think that's worth exploring?

OR

Maybe converse should not use this extractor concept at all since it doesn't need to.

hmstepanek (changes requested):
A couple things that need addressing but overall looks good. I'm going to try to run this manually next week to verify but these are just my thoughts from looking at the code changes.

hmstepanek (commented):
Just ran this locally and it appears to be behaving as expected so all good from that side of things!

newrelic/hooks/external_botocore.py
umaannamalai: I went ahead and added instrumentation into the aiobotocore hooks file to cover async clients and there are now tests associated with that which should cover this.

tests/external_botocore/conftest.py
```diff
@@ -103,6 +107,58 @@ def bedrock_server():
...
+        wrap_function_wrapper(
+            "botocore.eventstream", "EventStreamBuffer.add_data", wrap_botocore_eventstream_add_data_converse
+        )
+        yield client  # Run tests
+
+        # Write responses to audit log
+        bedrock_audit_log_contents = dict(
+            sorted(BEDROCK_CONVERSE_AUDIT_LOG_CONTENTS.items(), key=lambda i: (i[1][1], i[0]))
+        )
+        with open(BEDROCK_CONVERSE_AUDIT_LOG_FILE, "w") as audit_log_fp:
```
hmstepanek:
```suggestion
        with BEDROCK_CONVERSE_AUDIT_LOG_FILE.open("w", encoding="utf-8") as audit_log_fp:
```

newrelic/hooks/external_aiobotocore.py
```diff
@@ -97,23 +98,36 @@ async def wrap_client__make_api_call(wrapped, instance, args, kwargs):
...
-        model = args[1].get("modelId")
-        request_body = args[1].get("body")
-        is_embedding = "embed" in model
-    else:
-        model = ""
-        request_body = None
-        is_embedding = False
+    try:
+        bedrock_args = args[1] if len(args) >= 2 else {}
+        model = bedrock_args.get("modelId") if bedrock_args else ""
```
hmstepanek: Can this just be None otherwise? It looks like the line below handles it being None.
```suggestion
        model = bedrock_args.get("modelId")  # This will return None if it's not in there already.
```

newrelic/hooks/external_aiobotocore.py
```diff
@@ -97,23 +98,36 @@ async def wrap_client__make_api_call(wrapped, instance, args, kwargs):
...
+        is_embedding = "embed" in model if model else False
+
+        if is_converse:
+            request_body = {}
+        else:
+            request_body = bedrock_args.get("body") if bedrock_args else ""
```
hmstepanek:
```suggestion
            request_body = bedrock_args.get("body")
```

newrelic/hooks/external_botocore.py
```diff
@@ -42,7 +42,7 @@
 _logger = logging.getLogger(__name__)
 
 EXCEPTION_HANDLING_FAILURE_LOG_MESSAGE = "Exception occurred in botocore instrumentation for AWS Bedrock: While reporting an exception in botocore, another exception occurred. Report this issue to New Relic Support.\n%s"
-REQUEST_EXTACTOR_FAILURE_LOG_MESSAGE = "Exception occurred in botocore instrumentation for AWS Bedrock: Failed to extract request information. Report this issue to New Relic Support.\n%s"
+REQUEST_EXTRACTOR_FAILURE_LOG_MESSAGE = "Exception occurred in botocore instrumentation for AWS Bedrock: Failed to extract request information. Report this issue to New Relic Support.\n%s"
```
hmstepanek: I think the `\n%s` is leftover here from before it was using `exec_info=True`
```suggestion
REQUEST_EXTRACTOR_FAILURE_LOG_MESSAGE = "Exception occurred in botocore instrumentation for AWS Bedrock: Failed to extract request information. Report this issue to New Relic Support."
```

newrelic/hooks/external_botocore.py
```diff
@@ -660,6 +677,7 @@ def _wrap_bedrock_runtime_invoke_model(wrapped, instance, args, kwargs):
         instance._nr_ft = ft
         instance._nr_response_streaming = response_streaming
         instance._nr_settings = settings
+        instance._nr_is_converse = False
```
hmstepanek: Had a thought here-instead of adding this to the existing instrumentation points when it's False-can we just check for if the attr is present like we do with `__wrapped__`? So add it when it's  true and then instead of checking for it being True just check for if it's present?

hmstepanek (changes requested):
A couple small things but otherwise looks good!

newrelic/hooks/external_aiobotocore.py
```diff
@@ -97,23 +98,33 @@ async def wrap_client__make_api_call(wrapped, instance, args, kwargs):
...
-        model = args[1].get("modelId")
-        request_body = args[1].get("body")
-        is_embedding = "embed" in model
-    else:
-        model = ""
-        request_body = None
-        is_embedding = False
+    try:
+        bedrock_args = args[1] if len(args) >= 2 else {}
+        model = bedrock_args.get("modelId")  # This will return None if it's not in there already
```
hmstepanek:
```suggestion
        model = bedrock_args.get("modelId")
```

hmstepanek (dismissed):
Looks good-just one item to remove the comment.

hmstepanek (approved):

TimPansino (approved):

TimPansino (approved):

PR #1429, opened 2025-07-23
Celery (re)instrumentation

lrafeei:
This PR picks new instrumentation points for Celery support to allow for Custom Task classes.

### DIFF
--- a/newrelic/config.py
+++ b/newrelic/config.py
@@ -4013,17 +4013,8 @@ def _process_module_builtin_defaults():
     # _process_module_definition("celery.app.task", "newrelic.hooks.application_celery", "instrument_celery_app_task")
     _process_module_definition("celery.app.trace", "newrelic.hooks.application_celery", "instrument_celery_app_trace")
     _process_module_definition("celery.worker", "newrelic.hooks.application_celery", "instrument_celery_worker")
-    # _process_module_definition(
-    #     "celery.concurrency.processes", "newrelic.hooks.application_celery", "instrument_celery_worker"
-    # )
     _process_module_definition(
         "celery.concurrency.prefork", "newrelic.hooks.application_celery", "instrument_celery_worker"
-    )
-    # _process_module_definition(
-    #     "celery.concurrency.asynpool", "newrelic.hooks.application_celery", "instrument_celery_worker"
-    # )
-    _process_module_definition(
-        "billiard.process", "newrelic.hooks.application_celery", "instrument_billiard_process"
     )
 
     _process_module_definition("billiard.pool", "newrelic.hooks.application_celery", "instrument_billiard_pool")
--- a/newrelic/config.py
+++ b/newrelic/config.py
@@ -4007,18 +4007,25 @@ def _process_module_builtin_defaults():
         "instrument_rest_framework_decorators",
     )
 
-    _process_module_definition("celery.task.base", "newrelic.hooks.application_celery", "instrument_celery_app_task")
-    _process_module_definition("celery.app.task", "newrelic.hooks.application_celery", "instrument_celery_app_task")
+    _process_module_definition("celery.app.base", "newrelic.hooks.application_celery", "instrument_celery_app_base")
+    _process_module_definition("celery.local", "newrelic.hooks.application_celery", "instrument_celery_local")
+    # _process_module_definition("celery.task.base", "newrelic.hooks.application_celery", "instrument_celery_app_task")
+    # _process_module_definition("celery.app.task", "newrelic.hooks.application_celery", "instrument_celery_app_task")
     _process_module_definition("celery.app.trace", "newrelic.hooks.application_celery", "instrument_celery_app_trace")
     _process_module_definition("celery.worker", "newrelic.hooks.application_celery", "instrument_celery_worker")
-    _process_module_definition(
-        "celery.concurrency.processes", "newrelic.hooks.application_celery", "instrument_celery_worker"
-    )
+    # _process_module_definition(
+    #     "celery.concurrency.processes", "newrelic.hooks.application_celery", "instrument_celery_worker"
+    # )
     _process_module_definition(
         "celery.concurrency.prefork", "newrelic.hooks.application_celery", "instrument_celery_worker"
     )
+    # _process_module_definition(
+    #     "celery.concurrency.asynpool", "newrelic.hooks.application_celery", "instrument_celery_worker"
+    # )
+    _process_module_definition(
+        "billiard.process", "newrelic.hooks.application_celery", "instrument_billiard_process"
+    )
 
-    _process_module_definition("celery.app.base", "newrelic.hooks.application_celery", "instrument_celery_app_base")
     _process_module_definition("billiard.pool", "newrelic.hooks.application_celery", "instrument_billiard_pool")
 
     _process_module_definition("flup.server.cgi", "newrelic.hooks.adapter_flup", "instrument_flup_server_cgi")
--- a/newrelic/config.py
+++ b/newrelic/config.py
@@ -4007,7 +4007,7 @@ def _process_module_builtin_defaults():
         "instrument_rest_framework_decorators",
     )
 
-    _process_module_definition("celery.app.base", "newrelic.hooks.application_celery", "instrument_celery_app_base")
+    # _process_module_definition("celery.app.base", "newrelic.hooks.application_celery", "instrument_celery_app_base")
     _process_module_definition("celery.local", "newrelic.hooks.application_celery", "instrument_celery_local")
     # _process_module_definition("celery.task.base", "newrelic.hooks.application_celery", "instrument_celery_app_task")
     # _process_module_definition("celery.app.task", "newrelic.hooks.application_celery", "instrument_celery_app_task")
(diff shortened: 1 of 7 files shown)


tests/application_celery/_target_application.py
```diff
@@ -27,11 +26,37 @@
...
+
+
+class CustomCeleryTaskWithRun(Task):
+    def __call__(self, *args, **kwargs):
+        transaction = current_transaction()
+        if transaction:
+            transaction.add_custom_attribute("custom_task_attribute", "Called with run")
+        return self.run(*args, **kwargs)
+
+
```
TimPansino, line 64: Nitpick, can we move these (and the tasks that use them) to the bottom of the file so the test app's functionality escalates in complexity rather than starting with the most confusing things?

newrelic/hooks/application_celery.py
```diff
@@ -63,6 +64,94 @@ def task_info(instance, *args, **kwargs):
...
+    class TaskWrapper(FunctionWrapper):
+        def run(self, *args, **kwargs):
+            return self.__call__(*args, **kwargs)
+
+    try:
+        bound_args = bind_args(wrapped, args, kwargs)
+        task = bound_args.get("task", None)
+
+        task = TaskWrapper(task, wrap_task_call)
+        task.__module__ = wrapped.__module__  # Ensure module is set for monkeypatching detection
```
TimPansino: I'm not sure I understand what this is doing, and the comment is insufficient.

From inspecting execution, `task.__module__ = 'celery.app.builtins'` and `wrapped.__module__ = 'celery.app.trace'`. 

"Ensure the module is set" doesn't really clarify what it's being set to, and "for monkeypatching detection" is confusing. What is doing the detecting and why is that important?

I disabled this line and the tests pass 100%, so if there's some reason to be doing this we need a test for it.

newrelic/config.py
```diff
@@ -4010,18 +4010,13 @@ def _process_module_builtin_defaults():
         "instrument_rest_framework_decorators",
     )
 
-    _process_module_definition("celery.task.base", "newrelic.hooks.application_celery", "instrument_celery_app_task")
-    _process_module_definition("celery.app.task", "newrelic.hooks.application_celery", "instrument_celery_app_task")
```
TimPansino, line 4014: I presume with removing these hooks we don't support _some_ older versions of Celery outside our support window, but let's be sure to call that out in the release notes as I'm sure some customers out there use ancient versions of Celery.

tests/application_celery/test_distributed_tracing.py
```diff
@@ -12,52 +12,104 @@
...
-    name="test_distributed_tracing:test_celery_task_distributed_tracing_disabled", background_task=True
-)
-@validate_transaction_count(2)
-@background_task()
-def test_celery_task_distributed_tracing_disabled():
-    result = add.apply_async((1, 2))
-    result = result.get()
-    assert result == 3
+@pytest.mark.parametrize("dt_enabled", [True, False])
+def test_celery_task_distributed_tracing_inside_background_task(dt_enabled):
```
TimPansino: The names of all these tests in this file have gotten so long they don't actually describe what they do. Can we shorten all of these and clarify what we're actually testing? Something like this maybe to trim out as many words as possible.

```suggestion
def test_distributed_tracing_inside_txn_apply_async(dt_enabled):
```

tests/application_celery/test_distributed_tracing.py
```diff
@@ -12,52 +12,104 @@
...
+        result = add.apply_async((1, 2))
+        result = result.get()
+        assert result == 3
+
+    _test()
+
+
+# In this case, the background task creating the transaction
+# has not generated a distributed trace header, so the Celery
+# task will not have a distributed trace header to accept.
```
TimPansino: This comment and the test don't make sense to me.

This test is for Celery's `.apply()` method which is immediate and local execution of the task no?

So the reason there's no distributed tracing headers to accept is because this isn't a distributed system, and Celery isn't handling the transaction?

We should validate that no DT headers are accepted for this then with forgone metrics by setting it to `None` like the above tests do when DT is disabled.

TimPansino: If all of that is true, why is the test after this seeing DT headers and accepting them? Isn't that also local exection? Where is it even getting these headers from to accept?

tests/application_celery/test_distributed_tracing.py
```diff
@@ -12,52 +12,104 @@
...
+        group="Celery",
+        rollup_metrics=[
+            ("Supportability/TraceContext/Create/Success", 1 if dt_enabled else None),
+            ("Supportability/DistributedTrace/CreatePayload/Success", 1 if dt_enabled else None),
+        ],
+        background_task=True,
+    )
+    @validate_transaction_count(1)  # In the same process, so only one transaction
+    def _test():
+        result = add.apply((1, 2))
```
TimPansino, line 136: Additional thought, can't you call tasks directly like functions? If that's not different execution-wise from calling `.apply()`, shouldn't we test that here as well the same as we test `.apply()`?

tests/application_celery/test_task_methods.py
```diff
@@ -13,31 +13,14 @@
...
-FORGONE_TASK_METRICS = [("Function/_target_application.add", None), ("Function/_target_application.tsum", None)]
 
-
-@pytest.fixture(scope="module", autouse=True, params=[False, True], ids=["unpatched", "patched"])
-def with_worker_optimizations(request, celery_worker_available):
-    if request.param:
-        celery.app.trace.setup_worker_optimizations(celery_worker_available.app)
-
-    yield request.param
-    celery.app.trace.reset_worker_optimizations()
```
TimPansino, line 31: This was a major bug in our instrumentation, I don't like getting rid of the regression tests for it. Even if it no longer affects our instrumentation _today_, it's very likely that Celery rewrites something in the future and meta-programming is very dangerous to not to continue to validate.

Random thought but should we just move this to `conftest.py` so it applies to all tests in this suite and not just this file? Change the scope to `"session"` instead of `"module"` though.

tests/application_celery/test_task_methods.py
```diff
@@ -13,31 +13,14 @@
...
-
-
-@validate_transaction_metrics(
-    name="_target_application.add",
-    group="Celery",
-    scoped_metrics=FORGONE_TASK_METRICS,
-    rollup_metrics=FORGONE_TASK_METRICS,
-    background_task=True,
-)
```
TimPansino: All of these validators have the `FORGONE_TASK_METRICS` removed but that was a valid part of the test. It's testing that we don't double wrap tasks with both a transaction and function trace. I'd like to leave that in to prevent regressions.

tests/application_celery/test_wrappers.py
```diff
@@ -28,17 +25,7 @@ def test_task_wrapping_detection():
...
-    is_instrumented = lambda: isinstance(celery.app.task.BaseTask.__call__, _NRBoundFunctionWrapper)
-
-    celery.app.trace.reset_worker_optimizations()
-    assert is_instrumented(), "Instrumentation not initially applied."
-
-    celery.app.trace.setup_worker_optimizations(celery_worker_available.app)
-    assert is_instrumented(), "setup_worker_optimizations removed instrumentation."
-
-    celery.app.trace.reset_worker_optimizations()
-    assert is_instrumented(), "reset_worker_optimizations removed instrumentation."
```
TimPansino: I think this test still serves a purpose in preventing regressions if instrumentation changes in the future. Maybe leave a comment about the history of why this test is in there with meta-programming removing instrumentation in the past and just leave this test in there? You can change the `is_instrumented` callable to check if `add.__call__` is a function wrapper like the other test.

tests/application_celery/test_distributed_tracing.py
lrafeei: Good point, I meant to add a direct call test as well.  The execution path is different from `.apply()` even though they both are executed on the main process.

tests/application_celery/test_distributed_tracing.py
lrafeei: I've added a separate test for this, where a task is called within a transaction (background task) and `insert_distributed_trace_headers()`

Basically, `delay`/`apply_async` create worker processes and `apply`/`__call__` use the main process to execute.  In the case of a task being called in a function that is wrapped with the background_task decorator, any tasks called with `delay`/`apply_async` will create a worker process to execute the task and will no longer be in the context of the transaction.  Conversely, any tasks called with `apply`/`__call__` will execute in the main process and will be a function trace of the existing transaction.

tests/application_celery/test_task_methods.py
lrafeei: That is fair, I can add this test back in and in `confest.py`.  Since we are no longer instrumenting `BaseTask`/`Task`, I did remove the instrumentation for those functions, though.

newrelic/hooks/application_celery.py
lrafeei: Defining the TaskWrapper where the `run()` method just calls the existing `__call__` method covers the case that the overriding of the `__module__` attribute covers.  I believe this was a way to optimize functionality.

Using the scenario where the `__call__` method has not been overridden and the `run()` method has been defined:
Without overriding the `__module__` attribute, the task tracer will simply use the `run()` method and bypass the `__call__` function.  Then, the `TaskWrapper` class will reroute that back to the `__call__` method.  When overriding the `__module__` attribute, the task tracer will see that the `__call__` method has been overridden (which it does by traversing the MRO), and use that.

lrafeei: That being said, whoops.  That is supposed to be `task.__module__ = wrap_task_call.__module__` (so wrapper function instead of wrapped function)

I figured it would be faster execution to keep that line in there, but I don't have strong feelings about removing it either.

TimPansino (approved):

PR #1431, opened 2025-07-23
Change Paths to use pathlib.Path

TimPansino:
# Overview

To make paths more portable for Windows support:

* Update all paths in the agent from strings and `os.path` to use `pathlib.Path`.
* Enable `flake8-pathlib` linter to ensure no regressions.

### DIFF
--- a/newrelic/bootstrap/sitecustomize.py
+++ b/newrelic/bootstrap/sitecustomize.py
@@ -16,6 +16,7 @@ import os
 import sys
 import time
 from importlib.machinery import PathFinder
+from pathlib import Path
 
 # Define some debug logging routines to help sort out things when this
 # all doesn't work as expected.
@@ -42,7 +43,7 @@ def del_sys_path_entry(path):
 
 log_message("New Relic Bootstrap (%s)", __file__)
 
-log_message("working_directory = %r", os.getcwd())
+log_message("working_directory = %r", str(Path.cwd()))
 
 log_message("sys.prefix = %r", os.path.normpath(sys.prefix))
 
@@ -74,7 +75,7 @@ for name in sorted(os.environ.keys()):
 # imp module to find the module, excluding the bootstrap directory from
 # the search, and then load what was found.
 
-boot_directory = os.path.dirname(__file__)
+boot_directory = str(Path(__file__).parent)
 log_message("boot_directory = %r", boot_directory)
 
 del_sys_path_entry(boot_directory)
@@ -160,7 +161,7 @@ if k8s_operator_enabled or azure_operator_enabled or (python_prefix_matches and 
             # 'newrelic' module to reduce chance that will cause any issues.
             # If it is a buildout created script, it will replace the whole
             # sys.path again later anyway.
-            root_directory = os.path.dirname(os.path.dirname(boot_directory))
+            root_directory = str(Path(boot_directory).parent.parent)
             log_message("root_directory = %r", root_directory)
 
             new_relic_path = root_directory
--- a/newrelic/common/system_info.py
+++ b/newrelic/common/system_info.py
@@ -24,6 +24,7 @@ import socket
 import subprocess
 import sys
 import threading
+from pathlib import Path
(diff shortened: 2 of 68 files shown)


newrelic/bootstrap/sitecustomize.py
```diff
@@ -160,7 +161,7 @@ def del_sys_path_entry(path):
             # 'newrelic' module to reduce chance that will cause any issues.
             # If it is a buildout created script, it will replace the whole
             # sys.path again later anyway.
-            root_directory = os.path.dirname(os.path.dirname(boot_directory))
+            root_directory = str(Path(boot_directory).parent.parent)
```
hmstepanek, line 165: This is actually cleaner code than it was before which is nice. 😊

newrelic/common/system_info.py
```diff
@@ -74,7 +75,7 @@ def logical_processor_count():
     # looking at the devices corresponding to the available CPUs.
 
     try:
-        pseudoDevices = os.listdir("/devices/pseudo/")
+        pseudoDevices = Path("/devices/pseudo/").iterdir()
```
hmstepanek, line 78: I don't think these locations make sense on windows but I assume we'll address that later.

newrelic/console.py
```diff
@@ -446,15 +447,15 @@ def __thread_run(self):
...
+                Path(self.__listener_socket).unlink()
             except Exception:
                 pass
 
             listener = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
             listener.bind(self.__listener_socket)
 
             atexit.register(self.__socket_cleanup, self.__listener_socket)
-            os.chmod(self.__listener_socket, 0o600)
+            Path(self.__listener_socket).chmod(0o600)
```
hmstepanek, line 458: Does this work on windows?

tests/agent_unittests/test_check_environment.py
```diff
@@ -24,15 +24,14 @@
...
-        uwsgi_dir = os.path.join(temp_dir, "uwsgi")
-        init_file = os.path.join(uwsgi_dir, "__init__.py")
-        os.makedirs(uwsgi_dir)
-        with open(init_file, "w") as f:
-            for key, value in content.items():
-                f.write(f"{key} = {value}")
+        uwsgi_dir = temp_dir / "uwsgi"
+        init_file = uwsgi_dir / "__init__.py"
+        uwsgi_dir.mkdir(parents=True)
+        with init_file.open("w") as f:
```
hmstepanek, line 31: Is the default mode when opening a file the same between `os.open` and `pathlib.open`?

hmstepanek (approved):
Had a couple questions, otherwise looks good!

tests/agent_unittests/test_check_environment.py
TimPansino: Yes, they're both read-only by default.

* https://docs.python.org/3/library/pathlib.html#pathlib.Path.open
* https://docs.python.org/3/library/functions.html#open

newrelic/console.py
TimPansino: Sounds like no, but it'll ignore it. Can try and fix that when we get to trying to run this on Windows if it's necessary to fix.

https://docs.python.org/3.13/library/os.html#os.chmod

newrelic/common/system_info.py
TimPansino: No but according to the commend above in this file, that line is not reachable as Windows is already handled. That block is only for checking on Solaris systems I think based on the other comment.


```
    # The multiprocessing module provides support for Windows,
    # BSD systems (including MacOS X) and systems which support
    # the POSIX API for querying the number of CPUs.

    try:
        return multiprocessing.cpu_count()
    except NotImplementedError:
        pass
```

```
    # Assuming that Solaris will support POSIX API for querying
    # the number of CPUs. Just in case though, work it out by
    # looking at the devices corresponding to the available CPUs.

    try:
        pseudoDevices = os.listdir("/devices/pseudo/")
```

PR #1432, opened 2025-07-24
Add Workflow Permissions

TimPansino:
# Overview

* Add default permissions to all workflows to satisfy security scanning.
* Remove statuses permissions from Megalinter
  * Also removes status reporter, a deprecated and buggy feature.


hmstepanek (approved):
Looks good.

PR #1439, opened 2025-07-31
Add Windows Test Runners for All Core Test Suites

TimPansino:
# Overview

* Add Windows test runners to Github Actions.
  * This requires manually installs of Python using the setup-python action, which is a bit of a regression from our nice CI image. Hopefully this isn't a big deal for just the core suites, if it is we may have to explore a Windows docker image for the CI.
* Add Windows test environments for all core test suites in tox.
* Initial Windows support limited to Python 3.13


tox.ini
```diff
@@ -44,6 +44,19 @@ setupdir = {toxinidir}
 ; Fail tests when interpreters are missing.
 skip_missing_interpreters = false
 envlist =
+    # Core Features Tests run on Linux, MacOS, and Windows
+    python-agent_features-{py37,py38,py39,py310,py311,py312,py313}-{with,without}_extensions,
```
hmstepanek: Didn't we want to only run on 3.13?

tests/agent_unittests/test_encoding_utils.py
```diff
@@ -64,19 +65,25 @@ def _generator():
...
-        (10.0, "10.0"),
-        ("my_string", '"my_string"'),
-        (b"my_bytes", '"my_bytes"'),
-        ({"id": 1, "name": "test", "NoneType": None}, '{"id":1,"name":"test","NoneType":null}'),
-        (_generator(), "[1,2,3]"),
-        (tuple(range(4, 7)), "[4,5,6]"),
-        (Path("test/path/file.txt"), '"test/path/file.txt"'),
-    ],
-    ids=["int", "float", "str", "bytes", "dict", "generator", "iterable", "Path"],
+    "input_,expected", JSON_ENCODE_TESTS, ids=["int", "float", "str", "bytes", "dict", "generator", "iterable", "Path"]
```
hmstepanek: It's a little weird to not put the test ids along side the test values-maybe move those out into a constant too next to the JSON_ENCODE_TESTS as well.

tests/agent_unittests/test_agent_protocol.py
```diff
@@ -522,7 +522,8 @@ def test_audit_logging():
     protocol = AgentProtocol(settings, client_cls=HttpClientRecorder)
     protocol.send("preconnect")
 
-    with Path(f.name).open() as f:
+    audit_log_path = Path(f.name)
+    with audit_log_path.open() as f:
```
hmstepanek, line 526: Should this have an encoding on it?

tests/agent_features/test_async_generator_trace.py
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import functools
-import sys
```
hmstepanek, line 16: I don't understand how we are still finding these. Must have been missed in the original linted PR.

tests/agent_features/test_agent_control_health_check.py
```diff
@@ -41,8 +38,29 @@ def get_health_file_contents(tmp_path):
...
+    monkeypatch.setenv("NEW_RELIC_AGENT_CONTROL_ENABLED", "True")
+    monkeypatch.setenv("NEW_RELIC_AGENT_CONTROL_HEALTH_DELIVERY_LOCATION", file_uri)
+
+    agent_control_instance = agent_control_health_instance()
+    assert not agent_control_instance.health_delivery_location_is_valid
+
+
+@pytest.mark.skipif(sys.platform != "win32", reason="Only valid for Windows")
+@pytest.mark.parametrize("leading_slash", [True, False], ids=["leading_slash", "no_leading_slash"])
+def test_inconsistent_paths_on_windows(monkeypatch, tmp_path, leading_slash):
```
hmstepanek, line 52: I feel like you went over this in mobbing but I can't remember now. Is this just because windows allows both `///` and `//` so it must handle both?

newrelic/core/agent_control_health.py
```diff
@@ -135,16 +101,87 @@ def health_check_enabled(self):
...
     @property
     def health_delivery_location(self):
-        # Set a default file path if env var is not set or set to an empty string
-        health_file_location = (
+        file_uri = (
             os.environ.get("NEW_RELIC_AGENT_CONTROL_HEALTH_DELIVERY_LOCATION", "") or "file:///newrelic/apm/health"
         )
 
-        return health_file_location
+        # Return from cache if already parsed
```
hmstepanek, line 112: I'm gonna have to look at this part a little more in depth later.

hmstepanek (commented):
Posting this now so I can get this feedback to you-I'd like to take a closer look at the healthcheck section changes when I have some more time.

tox.ini
TimPansino: Confusing change, but the `python-` prefix now denotes default python runners with no dependencies, which is linux. The envs with `windows-` as a prefix are for running on windows runners specifically. The line you're looking at is a linux only runner set, there's a windows env for agent features on just Python 3.13 down below.

tests/agent_unittests/test_agent_protocol.py
TimPansino: No, the audit log file is being written to in the actual agent code in the [preferred encoding](https://docs.python.org/3/library/functions.html#open) as returned by `locale.getencoding()`. We read it back in using the same default encoding.

The other tests that needed encoding specified are pre-written files from Mac/Linux machines that encode unicode characters in UTF-8, which is not the default for Windows and produces errors.

tests/agent_features/test_async_generator_trace.py
TimPansino: I turned off the rule for unused imports in the tests folder because of fixtures commonly causing that to trip and get removed unless we mark them with `# noqa` markers. Unfortunately, that means we don't catch unused imports in tests files but they're not run in production so I wasn't too concerned with it.

newrelic/core/agent_control_health.py
TimPansino: The health stuff is a significant enough change that I could pull it out into a separate PR and mark these tests as XFAIL for now if you want.

tests/agent_features/test_agent_control_health_check.py
TimPansino: There are inconsistencies with handling the root path on windows that caused a ton of headaches due to mixing which library was creating vs parsing the file URIs.

A path for the file `C:\\Users\\Reli\\myapp\\main.py` will end up represented as a URI with the `file://` but both of these are valid for after the prefix:

`c:/user/reli/myapp/main.py` and `/c:/user/reli/myapp/main.py`.

It's hard to spot when it's `file://c:/user/reli/myapp/main.py` vs `file:///c:/user/reli/myapp/main.py` but the extra slash  causes SOME parsers to fail due to interpreting the leading slash as "relative to the root", but there's no folder named `c:` under the root, since that's a drive letter. It's effectively the same as looking for `C:\\C:`. So the Path object produced from parsing that doesn't exist.

That being said, it SHOULD be valid on Windows as a URI but not every parser interprets it correctly. The parser with correct behavior is once again `pathlib`, but the URI parser wasn't added until Python 3.13. Since we're now not aiming to support lower versions on Windows yet, I could comment out that whole section and call [Path.from_uri()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.from_uri) on 3.13+ to sidestep this whole thing.

tests/agent_unittests/test_encoding_utils.py
TimPansino: Fixed

newrelic/core/agent_control_health.py
hmstepanek: That's fine-I'm taking a look now. My brain was dead at the time and I thought it would be better to review it later when I could really focus on it.

tests/agent_features/test_async_generator_trace.py
hmstepanek: Ah I see. Makes sense.

PR #1444, opened 2025-08-09
Django middleware filtering settings

lrafeei:
This PR adds three settings related to Django middleware instrumentation:

`instrumentation.django_middleware.enabled`
`instrumentation.django_middleware.exclude`
`instrumentation.django_middleware.include`

Where the exclude/include settings follow similar rules to attribute exclude/include: https://docs.newrelic.com/docs/apm/agents/python-agent/attributes/enabling-disabling-attributes-python/#rule-specific-wins


newrelic/core/config.py
```diff
@@ -644,6 +649,17 @@ def _parse_attributes(s):
     return valid
 
 
+# Called from newrelic.config.py to parse django_middleware lists
+def _parse_django_middleware(s):
```
hmstepanek: Can we just point at this method rather than re-creating the exact same implementation again: `_parse_attributes`? Maybe just rename it to be generic or accept an additional argument for the log message.

newrelic/hooks/framework_django.py
```diff
@@ -1149,17 +1090,124 @@ def _wrapper(wrapped, instance, args, kwargs):
...
+        return True
+
+    # Check if the callable name matches any of the excluded middleware patterns.
+    # If middleware name ends with '*', it is a wildcard
+    deny = False
+    for exclude_middleware in settings.instrumentation.django_middleware.exclude:
+        if exclude_middleware.endswith("*"):
+            name = exclude_middleware.rstrip("*")
+            if callable_name.startswith(name):
+                if not include_logic(callable_name, name):
```
hmstepanek: Cache the value so we aren't computing it twice.
```suggestion
                include = include_logic(callable_name, name)
                if not include:
                    return False
                else:
                    deny |= include
```

newrelic/hooks/framework_django.py
```diff
@@ -1149,17 +1090,124 @@ def _wrapper(wrapped, instance, args, kwargs):
...
+        Where `middleware.*` is less specific than `middleware.parameters.*`
+        but `middleware.parameters.bar*` is more specific than `middleware.parameters.*`
+
+        In this case, we want to make sure to exclude any middleware that
+        matches the `middleware.parameters.bar*` pattern, but include any
+        other middleware that matches the `middleware.parameters.*` pattern.
+        """
+        if len(settings.instrumentation.django_middleware.include) == 0:
+            return True
+        for include_middleware in settings.instrumentation.django_middleware.include:
```
hmstepanek: My brain is fried from trying to understand this...I'll come back to this later and hopefully it will make more sense. In theory I see what you are trying to do here but my brain is not following all of these implementation details.

lrafeei: Fair--it gets a bit dicey.  A lot of the complexity in this logic comes from attempting to sort out a case like this one:

Middleware = django.middleware.common:CommonMiddleware, django.middleware.csrf:CsrfViewMiddleware
exclude = ["django.middleware.c*"],
include = ["django.middleware.*", "django.middleware.co*"]

The `django.middleware.*` in the include list is effectively a no op.  However, in the case where a user has this scenario, this should still filter correctly (based on the more specific include statement, `django.middleware.co*`, relative to the exclude statement `django.middleware.c*`

tests/framework_django/test_wildcard_filters.py
```diff
@@ -0,0 +1,134 @@
...
+    "transaction_tracer.explain_threshold": 0.0,
+    "transaction_tracer.transaction_threshold": 0.0,
+    "transaction_tracer.stack_trace_threshold": 0.0,
+    "debug.log_data_collector_payloads": True,
+    "debug.record_transaction_failure": True,
+    "debug.log_autorum_middleware": True,
+    "instrumentation.django_middleware.enabled": True,
+}
+
+wildcard_exclude_specific_include_settings = (
```
hmstepanek, line 44: Can we get some more tests here:
* there aren't *'d paths.
* there is only an exclude rule
* there is only an include rule
* there are no dots in the paths
* there are only dots in the include path
* there are only dots in the exclude path
* there is a path with a dot (no *) and a path with no dots and only a *
* case sensitive test
There's probably some additional cases I'm not thinking of?

I want to be really thorough when we test these cause there's a lot of complex logic there and I don't want to miss some corner case bug in the implementation. (I don't see any but it might be hard to catch without thorough testing.)

hmstepanek (changes requested):
Ok I mainly just want some more thorough tests for the filtering-otherwise this looks good.

PR #1448, opened 2025-08-13
Add w3c traceparent header support

hmstepanek:
# Overview
Add support for w3c traceparent sampled flag. See agent spec for details.


newrelic/api/transaction.py
```diff
@@ -1081,7 +1098,7 @@ def _create_distributed_trace_data(self):
         if not (account_id and application_id and trusted_account_key and settings.distributed_tracing.enabled):
             return
 
-        self._compute_sampled_and_priority()
+        self._compute_sampled_and_priority(traceparent)
```
hmstepanek:
```suggestion
        self._compute_sampled_and_priority()
```

newrelic/api/transaction.py
```diff
@@ -1004,16 +1005,32 @@ def _update_agent_attributes(self):
...
+        else:  # self._traceparent_sampled is False.
+            config = self.settings.distributed_tracing.sampler.remote_parent_not_sampled
+
+        if config == 'always_on':
+            self._sampled = True
+            self._priority = 2.0
+        elif config == 'always_off':
+            self._sampled = False
+            self._priority = 0
+        else:
```
TimPansino, line 1033: We should probably have a branch for an invalid setting here that logs a warning, and then follows the default path.

tests/agent_features/test_distributed_tracing.py
```diff
@@ -409,3 +411,65 @@ def _test_inbound_dt_payload_acceptance():
...
+        (False, 'always_on', 'default', None, None, True),  # Uses sampling alog.
+        (False, 'always_off', 'default', None, None, True),  # Uses sampling algo.
+        (True, 'default', 'always_on', None, None, True),  # Uses sampling algo.
+        (True, 'default', 'always_off', None, None, True),  # Uses sampling algo.
+        (False, 'default', 'always_on', True, 2, False),  # Always sampled.
+        (False, 'default', 'always_off', False, 0, False),  # Never sampled.
+    )
+)
+def test_distributed_trace_w3cparent_sampling_decision(sampled, remote_parent_sampled, remote_parent_not_sampled, expected_sampled, expected_priority, expected_adaptive_sampling_algo_called, wrap_sampling_algo_compute_sampled_and_priority):
+    required_intrinsics = []
```
TimPansino: I feel like this should potentially be ported to the cross agent tests and standarized.

tests/agent_features/test_distributed_tracing.py
```diff
@@ -409,3 +411,65 @@ def _test_inbound_dt_payload_acceptance():
...
+    global was_called
+    was_called = False
+
+    def _wrap_sampling_algo_compute_sampled_and_priority(wrapped, instance, args, kwargs):
+        global was_called
+        was_called = True
+        return wrapped(*args, **kwargs)
+
+    return _wrap_sampling_algo_compute_sampled_and_priority
+
```
TimPansino: The code here in this fixture and the code that's applying this as a wrapper should be a validator and go in the validators folder. This should be using a transient function wrapper that auto applies and unapplies this wrapping to the `sampling_algo...` function when the test runs.

tests/agent_features/test_distributed_tracing.py
hmstepanek: That's a really good point! It doesn't look like we have any but I'll create a ticket for myself to add some-then maybe we can get rid of this one in leu of the cross agent test.

tests/agent_features/test_distributed_tracing.py
TimPansino: I think you're looking for `validate_function_called`

tests/component_djangorestframework/test_application.py
```diff
@@ -15,7 +15,7 @@
 import django
 import pytest
 import webtest
-from testing_support.fixtures import function_not_called, override_generic_settings
+from testing_support.validators.validate_function_not_called import validate_function_not_called, override_generic_settings
```
TimPansino:
```suggestion
from testing_support.fixtures import override_generic_settings
from testing_support.validators.validate_function_not_called import validate_function_not_called
```

tests/framework_sanic/test_application.py
```diff
@@ -17,7 +17,7 @@
 import pytest
 import sanic
 from testing_support.fixtures import (
-    function_not_called,
+    validate_function_not_called,
```
TimPansino:
```suggestion

from testing_support.fixture.validate_function_not_called import validate_function_not_called
from testing_support.fixtures import (
```

TimPansino (changes requested):
There's a bunch of places where the imports are screwed up now due to moving that validator, I'll stop reviewing there and let you re-check those.

TimPansino (approved):

PR #1451, opened 2025-08-18
Add graphene-django instrumentation

lrafeei:
This PR adds instrumentation for `graphene-django`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1022,7 +1022,7 @@ jobs:
           - 2181:2181
 
       kafka:
-        image: bitnami/kafka:3.6.1
+        image: bitnamilegacy/kafka:3.3.2
         ports:
           - 8080:8080
           - 8082:8082


TimPansino:
First comment is that this should probably be a `component_` and not `framework_` prefix, as this library provides components for `django` and `graphene` and is not a standalone framework.

tests/framework_graphenedjango/urls.py
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from django.urls import path
+from graphene_django.views import GraphQLView
+
+# When New Relic supports async schemas/asgi for graphene-django,
+# we can add the schema as an argument like so:
+# path("graphql", GraphQLView.as_view(graphiql=False, schema=schema), name="graphql_sync")
+# and remove the global schema declaration in settings.py under "GRAPHENE"
```
TimPansino: Is the plan to support this initially without async and to add that later?

newrelic/hooks/framework_graphenedjango.py
```diff
@@ -0,0 +1,102 @@
...
+    except Exception:
+        # Execution finished synchronously, exit immediately.
+        trace.__exit__(*sys.exc_info())
+        raise
+    else:
+        trace.set_transaction_name(priority=14)
+        if isawaitable(result):
+            # Asynchronous implementations
+            # Return a coroutine that handles closing the operation trace
+            return nr_coro_graphql_impl_wrapper(wrapped, trace, ignore_graphql_duplicate_exception, result)
```
TimPansino: Is this something we should be using from an import? I'm worried about drift here. It has the potential to be broken later if we don't copy the implementation into this file.

newrelic/hooks/framework_graphenedjango.py
```diff
@@ -0,0 +1,102 @@
...
+    transaction = current_transaction()
+
+    if not transaction:
+        return wrapped(*args, **kwargs)
+
+    string_response, status_code = wrapped(*args, **kwargs)
+    response = json.loads(string_response)
+    transaction.process_response(status_code, response)
+
+    return string_response, status_code
```
TimPansino: I don't believe this instrumentation is necessary. If this is a view, it should be capturing the response headers and status code in the `django` instrumentation. right?

Tried it locally with this commented out, and it still passed 100% of tests. If this is actually necessary for some reason then we need a test that covers this.

tests/framework_graphenedjango/urls.py
lrafeei: Yes, exactly

newrelic/hooks/framework_graphenedjango.py
lrafeei: You're right.  I even added a validator to some tests after removing this to make sure that this still is populated.  Removed!

PR #1456, opened 2025-08-19
Update MCP instrumentation to check if AIM is enabled.

umaannamalai:
This PR adds checks to verify whether AIM is enabled before creating MCP-related function traces.

### DIFF
--- a/tests/adapter_mcp/test_mcp.py
+++ b/tests/adapter_mcp/test_mcp.py
@@ -13,13 +13,18 @@
 # limitations under the License.
 
 import pytest
+
 from fastmcp.client import Client
 from fastmcp.client.transports import FastMCPTransport
 from fastmcp.server.server import FastMCP
+from mcp.server.fastmcp.tools import ToolManager
+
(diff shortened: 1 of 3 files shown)


tests/adapter_mcp/test_mcp.py
```diff
@@ -114,3 +115,38 @@ async def _test():
...
+
+@disabled_ai_monitoring_settings
+@function_not_called("newrelic.api.function_trace", "FunctionTrace.__enter__")
+@background_task()
+def test_resource_tracing_aim_disabled(loop, fastmcp_server):
+    run_read_resources(loop, fastmcp_server, "greeting://Python")
+
+
+@disabled_ai_monitoring_settings
+@function_not_called("newrelic.api.function_trace", "FunctionTrace.__enter__")
```
TimPansino, line 167: This could also have been achieved with checking for the absence of these metrics, but I suppose this is more heavy handed and prevents drift issues. Probably would be fine either way.

PR #1477, opened 2025-09-02
Remove instrumentation for options method on async ElasticSearch client.

umaannamalai:
This PR removes wrapping for the method `options` on the async ES client. This method does not need to be wrapped in a `DatastoreTrace` since it is focused on modifying the client with custom config rather than conducting any datastore operations/ requests. The previous instrumentation caused a crash because the wrapper was expecting all client methods to be async but `options` is actually a sync method.


TimPansino (approved):

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-459136

PR #1478, opened 2025-09-03
Distributed CI Image Build

TimPansino:
# Overview

* Upgrade CI image builds to be distributed, allowing them to be built without emulation.
* Add a weekly rebuild of the CI image to keep Python versions up to date.
* Add Rust to the toolchain to avoid slowdowns with building Rust wheels (like cryptography).

### DIFF
--- a/.github/workflows/build-ci-image.yml
+++ b/.github/workflows/build-ci-image.yml
@@ -19,6 +19,7 @@ on:
 
 permissions:
   contents: read
+  packages: write
 
 concurrency:
   group: ${{ github.ref || github.run_id }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-ci-image.yml
```diff
@@ -16,21 +16,31 @@ name: Build CI Image
 
 on:
   workflow_dispatch: # Allow manual trigger
+  schedule:
+    - cron: "15 16 * * 0" # At 16:15 UTC on Sunday
```
hmstepanek, line 20: Thank you for the comment!

PR #1479, opened 2025-09-03
Remove old apis

lrafeei:
Removes the following APIs:

- record_exception (renamed to notice_error 4 years ago)

- `lambda_handler` & `LambdaHandlerWrapper`
- `accept_distributed_trace_payload`
- `create_distributed_trace_payload`
- `add_custom_parameter(s)`
- `get_browser_timing_footer`
- `ObjectWrapper`
- `NewRelicLogHandler`

Also:

- Removes `newrelic/api/in_function.py`, `newrelic/api/out_function.py`, `newrelic/api/pre_function.py`, and `newrelic/api/post_function.py`
- Removes `ignore_errors` argument from `error_trace()`


newrelic/api/transaction.py
```diff
@@ -1352,8 +1340,11 @@ def accept_distributed_trace_headers(self, headers, transport_type="HTTP"):
             self._record_supportability("Supportability/TraceContext/Accept/Success")
             return True
         elif distributed_header:
-            distributed_header = ensure_str(distributed_header)
             return self._accept_distributed_trace_payload(distributed_header, transport_type)
+        else:
+            # Do not return anything, but still generate supportability
+            # metric for the lack of payload/distributed_header
+            self._accept_distributed_trace_payload(distributed_header, transport_type)
```
TimPansino, line 1322: Is this change supposed to be in here? It doesn't seem directly related to the deprecations.

tests/cross_agent/test_distributed_tracing.py
```diff
@@ -125,7 +126,13 @@ def target_wsgi_application(environ, start_response):
...
     if outbound_payloads:
         for payload_assertions in outbound_payloads:
-            payload = txn._create_distributed_trace_payload()
+            headers = []
+            txn.insert_distributed_trace_headers(headers)
+            # To revert to the dict format of the payload, use this:
+            payload = json.loads(base64.b64decode([value for key, value in headers if key=="newrelic"][0]).decode("utf-8"))
+            payload_version = payload.get("v")
+            if payload_version and isinstance(payload_version, list):
+                payload["v"] = tuple(payload_version)
```
TimPansino, line 137: Is this actually required? I would think we would want to be running assertions on the payload as it was decoded. Otherwise we might miss some TypeErrors that this would suppress.

newrelic/api/transaction.py
lrafeei: In this case, `accept_distributed_trace_payload` and `accept_distributed_trace_headers` do not have a 1-to-1 mapping in terms of functionality.

`accept_distributed_trace_payload`'s functionality boils down to checking to see if DT has already been accepted, and if not, running `_accept_distributed_trace_payload`.  `_accept_distributed_trace_payload`'s first conditional is checking to see if there is a distributed_header (or payload) and if not, recording a supportability metric and returning False.

With `accept_distributed_trace_headers`, if there is no traceparent or distributed_header/payload, it will return None.  It seems that we still want a way to record a supportability metric to indicate that there was no payload in the headers while not changing the return logic.

tests/cross_agent/test_distributed_tracing.py
lrafeei: Great question!  In `create_distributed_trace_payload`, the payload data is generated from `_create_distributed_trace_data()` and run through `DistributedTracePayload`

`insert_distributed_trace_headers` (which is effectively just `_generate_distributed_trace_headers`) also takes the payload generated from `_create_distributed_trace_data()` and runs it through `DistributedTracePayload` but then runs it through `DistributedTracePayload.http_safe()`.  This encodes the payload with `base64.b64encode()`

All of that is to explain that when it is going through all this extra encoding/decoding and loads/dumps to keep the tests' original functionality, using `json.dumps()` converts tuples to lists (which checks out since JSON only has one list type).  This check is specific to this test and not logic that is applied to the agent itself.

TimPansino (approved):

PR #1485, opened 2025-09-09
Remove capture_params setting

lrafeei:
This PR removes the deprecated `capture_params` setting.  Note: Wait until #1482 has been merged to make review easier

While this setting has been removed, there are two settings that heavily relied upon this setting that were not removed:

- environ setting `newrelic.capture_request_params`
- `capture_request_params` API

Request parameters are a special kind of agent attribute that is prefaced with `request.parameters.`—with settings such as HSM, these need to be disabled.  The analogous settings for `capture_params` are as follows:

capture_params = True:
attributes.enabled = True
attributes.include=["request.parameters.*"]

capture_params = False:
attributes.enabled=True
attributes.exclude=["request.parameters.*"]


newrelic/api/web_transaction.py
```diff
@@ -618,27 +618,30 @@ def __init__(self, application, environ, source=None):
...
+
         # Make sure that if high security mode is enabled that
         # capture of request params is still being disabled.
         # No warning is issued for this in the logs because it
         # is a per request configuration and would create a lot
         # of noise.
 
         if settings.high_security:
-            self.capture_params = False
+            self.settings.attributes.exclude.append("request.parameters.*")
```
TimPansino: This section's behavior seems potentially incorrect with regards to high security. If high security mode is enabled, there shouldn't be any params captured. If it's not, THEN we could add the request params to the include list. Otherwise, we're going to see some conflicting settings for include and exclude where it's unclear which one should win.

Beyond that, this isn't correct as a user can set the include list to have a specific param included, and it will be included despite the exclude list set to `*`. This is because the attribute system prioritizes more specific over general when rules conflict.

This code seems like it's just recreating the behavior of `capture_params` under the hood, when really we should be taking it out altogether. Can we just remove this `_capture_request_params` attribute and replace it with:

1. Checking for high security mode, and not allowing any param captures at all if enabled.
2. Capturing the params and putting them through the attribute filter.

I'm not sure what the defaults are here and we may need to consider them, but that seems like the most sensible behavior.

newrelic/api/web_transaction.py
```diff
@@ -618,27 +618,30 @@ def __init__(self, application, environ, source=None):
...
 
-        # LEGACY: capture_params = False
-        #
-        #    Don't add request parameters at all, which means they will not
-        #    go through the AttributeFilter.
-        if self.capture_params is False:
+        # Don't add request parameters at all, which means
+        # they will not go through the AttributeFilter.
+        if "request.parameters.*" in self.settings.attributes.exclude:
             self._request_params.clear()
```
TimPansino: This is not correct. The above explanation applies, where a user can set all params to be excluded with exclude set to `request.parameters.*`, but then include only specific params with include settings like `request.parameters.my_param`.

The current change breaks this functionality by not capturing them in the first place, so the more specific rules won't win like they're supposed to.

TimPansino (changes requested):
Haven't finished reading the whole thing yet, but I think this might need some redesign so posting these initial comments early.

PR #1489, opened 2025-09-12
Remove Cross Application Tracing (CAT)

lrafeei:
This PR removes Cross Application Tracing (CAT) from the agent. This feature has been fully replaced by Distributed Tracing (DT).


newrelic/hooks/messagebroker_pika.py
```diff
@@ -62,9 +62,7 @@ def _add_consume_rabbitmq_trace(transaction, method, properties, nr_start_time,
 
     # Do not record dt headers in the segment parameters
     if headers:
-        headers.pop(MessageTrace.cat_id_key, None)
-        headers.pop(MessageTrace.cat_transaction_key, None)
-        headers.pop(MessageTrace.cat_distributed_trace_key, None)
+        headers.pop(MessageTrace.distributed_trace_key, None)
```
hmstepanek:
```suggestion
        headers.pop("newrelic", None)
```

newrelic/hooks/messagebroker_pika.py
```diff
@@ -124,9 +122,7 @@ def _nr_wrapper_basic_publish(wrapped, instance, args, kwargs):
     user_headers = properties.headers.copy()
 
     # Do not record dt headers in the segment parameters
-    user_headers.pop(MessageTrace.cat_id_key, None)
-    user_headers.pop(MessageTrace.cat_transaction_key, None)
-    user_headers.pop(MessageTrace.cat_distributed_trace_key, None)
+    user_headers.pop(MessageTrace.distributed_trace_key, None)
```
hmstepanek:
```suggestion
    user_headers.pop("newrelic", None)
```

newrelic/api/header_mixin.py
```diff
@@ -0,0 +1,52 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# HeaderMixin assumes the mixin class also inherits from TimeTrace
+class HeaderMixin:
+    synthetics_key = "X-NewRelic-Synthetics"
+    synthetics_info_key = "X-NewRelic-Synthetics-Info"
+    distributed_trace_key = "newrelic"
```
hmstepanek:
```suggestion
```

newrelic/config.py
```diff
@@ -453,7 +453,7 @@ def _process_configuration(section):
     _process_setting(section, "debug.disable_harvest_until_shutdown", "getboolean", None)
     _process_setting(section, "debug.connect_span_stream_in_developer_mode", "getboolean", None)
     _process_setting(section, "debug.otlp_content_encoding", "get", None)
-    _process_setting(section, "cross_application_tracer.enabled", "getboolean", None)
+    # _process_setting(section, "cross_application_tracer.enabled", "getboolean", None)
```
hmstepanek:
```suggestion
```

newrelic/core/transaction_node.py
```diff
@@ -541,11 +521,6 @@ def error_event_intrinsics(self, error, stats_table):
         intrinsics["transactionName"] = self.path
         intrinsics["spanId"] = error.span_id
```
hmstepanek, line 526:
```suggestion
intrinsics["nr.transactionGuid"] = self.guid
```

newrelic/hooks/application_celery.py
```diff
@@ -127,11 +127,6 @@ def wrapper(wrapped, instance, args, kwargs):
                                         headers.update(dict(dt_headers))
```
hmstepanek, line 127: Start reviewing again at this file.

hmstepanek (changes requested):
Posting a partial review. I'll review the rest later.

newrelic/core/agent_protocol.py
```diff
@@ -522,7 +522,7 @@ def finalize(self):
     @classmethod
     def connect(cls, app_name, linked_applications, environment, settings, client_cls=ServerlessModeClient):
         aws_lambda_metadata = settings.aws_lambda_metadata
-        settings = finalize_application_settings({"cross_application_tracer.enabled": False}, settings)
+        settings = finalize_application_settings()
```
hmstepanek:
```suggestion
        settings = finalize_application_settings(settings=settings)
```

newrelic/hooks/framework_aiohttp.py
```diff
@@ -279,16 +269,13 @@ async def _coro():
             response = await wrapped(*args, **kwargs)
 
             try:
-                trace.process_response_headers(response.headers.items())
+                trace.process_response(status_code=response.status)
```
hmstepanek: I don't see that process_respose_headers called process_response so I think this would just be deleted entirely right?

newrelic/hooks/framework_aiohttp.py
```diff
@@ -279,16 +269,13 @@ async def _coro():
...
             except:
                 pass
 
             return response
-        except Exception as e:
-            try:
-                trace.process_response_headers(e.headers.items())
-            except:
-                pass
+        except Exception:
```
hmstepanek: Same thing here-I think what was happening before was we had to process the headers regardless of whether there was an exception and when there was an exception we processed the headers and then re-raised the error to be caught by the code wrapped above this.

newrelic/hooks/framework_aiohttp.py
```diff
@@ -17,6 +17,7 @@
 
 from newrelic.api.external_trace import ExternalTrace
 from newrelic.api.function_trace import function_trace
+from newrelic.api.time_trace import notice_error
```
hmstepanek:
```suggestion
```

newrelic/hooks/framework_aiohttp.py
```diff
@@ -214,18 +204,18 @@ def items(self):
         return itertools.chain(self.__wrapped__.items(), nr_headers.items())
 
 
-def _nr_aiohttp_add_cat_headers_(wrapped, instance, args, kwargs):
+def _nr_aiohttp_add_dt_headers_(wrapped, instance, args, kwargs):
```
hmstepanek, line 194: I have to admit, I'm really having a hard time understanding the purpose of the previous code in here.

newrelic/hooks/framework_aiohttp.py
```diff
@@ -33,17 +34,6 @@ def aiohttp_version_info():
     return tuple(int(_) for _ in aiohttp.__version__.split(".")[:2])
 
 
-def headers_preserve_casing():
```
hmstepanek, line 36: Wouldn't this always return True?

newrelic/api/web_transaction.py
```diff
@@ -337,16 +333,7 @@ def process_response(self, status_code, response_headers):
...
-            return []
-
-        # Generate CAT response headers
-        try:
-            read_length = int(self._request_headers.get("content-length"))
-        except Exception:
-            read_length = -1
-
-        return self._generate_response_headers(read_length)
+        return []
```
hmstepanek: I think we could just make this function not return anything now.

newrelic/hooks/framework_sanic.py
```diff
@@ -205,9 +205,9 @@ def _nr_sanic_response_parse_headers(wrapped, instance, args, kwargs):
         return wrapped(*args, **kwargs)
 
     # instance is the response object
-    cat_headers = transaction.process_response(str(instance.status), instance.headers.items())
+    dt_headers = transaction.process_response(str(instance.status), instance.headers.items())
 
-    for header_name, header_value in cat_headers:
+    for header_name, header_value in dt_headers:
```
hmstepanek: I think we could get rid of this loop entirely since we aren't adding any response headers anymore.

newrelic/hooks/framework_tornado.py
```diff
@@ -132,9 +132,9 @@ def wrap_write_headers(wrapped, instance, args, kwargs):
 
     if transaction:
         http_status, headers = _bind_response_headers(*args, **kwargs)
-        cat_headers = transaction.process_response(http_status, headers)
+        dt_headers = transaction.process_response(http_status, headers)
 
-        for name, value in cat_headers:
+        for name, value in dt_headers:
```
hmstepanek: Same here-there's probably a lot more places so I'm gonna not mention it again but anywhere where we were adding response headers - DT doesn't add response headers so we should remove these loops.

newrelic/core/transaction_node.py
```diff
@@ -465,23 +458,10 @@ def _add_if_not_empty(key, value):
 
         apdex_perf_zone = self.apdex_perf_zone()
         _add_if_not_empty("apdexPerfZone", apdex_perf_zone)
-        _add_if_not_empty("nr.apdexPerfZone", apdex_perf_zone)
```
hmstepanek, line 469: Why are we removing the `nr.apdexPerfZone`? While this is in the cross agent spec this doesn't seem related to CAT directly? Also we are dropping the nr. prefixed but not the other one which seems weird.

tests/external_httpx/test_client.py
```diff
@@ -17,48 +17,35 @@
...
 from newrelic.api.time_trace import current_trace
 from newrelic.api.transaction import current_transaction
 from newrelic.core.config import global_settings
 
-ENCODING_KEY = "1234567890123456789012345678901234567890"
 SCOPED_METRICS = []
 ROLLUP_METRICS = [("External/all", 2), ("External/allOther", 2)]
 
-CAT_RESPONSE_CODE = None
+DT_RESPONSE_CODE = None
```
hmstepanek: Wondering if we can just name this RESPONSE_CODE.

tests/framework_aiohttp/conftest.py
```diff
@@ -115,15 +115,15 @@ def handler(self):
 def mock_external_http_server():
     response_values = []
 
-    def respond_with_cat_header(self):
+    def respond_with_dt_header(self):
```
hmstepanek: This seems like a weird name as there are no response dt headers

tests/framework_aiohttp/test_client.py
```diff
@@ -52,7 +51,8 @@ def task(loop, method, exc_expected, url):
         assert isinstance(text_list[0], _expected_error_class), text_list[0].__class__
         assert isinstance(text_list[1], _expected_error_class), text_list[1].__class__
     else:
-        assert text_list[0] == text_list[1], text_list
+        assert text_list[0]
```
hmstepanek, line 54: This doesn't seem like the same assertion equivalency here-need to investigate this one a little deeper.

hmstepanek: After looking at this a little closer this seems fine.

newrelic/core/transaction_node.py
```diff
@@ -470,18 +463,6 @@ def _add_if_not_empty(key, value):
...
-            intrinsics["nr.pathHash"] = self.path_hash
-
-            _add_if_not_empty("nr.referringPathHash", self.referring_path_hash)
-            _add_if_not_empty("nr.alternatePathHashes", ",".join(self.alternate_path_hashes))
-            _add_if_not_empty("nr.referringTransactionGuid", self.referring_transaction_guid)
-
-        if self.synthetics_resource_id:
-            intrinsics["nr.guid"] = self.guid
-
         if self.parent_tx:
```
hmstepanek, line 468:
```suggestion
        if self.synthetics_resource_id:
            intrinsics["nr.guid"] = self.guid
        if self.parent_tx:
```

newrelic/hooks/framework_aiohttp.py
TimPansino: Just glancing through this review, but `process_response` should be called for `ExternalTrace`s since it's what sets the `http.statusCode` attribute. 

We can get rid of all parsing of response headers for external calls, but we should definitely be capturing the status code anywhere we weren't already.

newrelic/core/transaction_node.py
lrafeei: This was originally added as a temporary object to have "backwards" compatibility with CAT.  `apdexPerfZone` is intended for DT.

newrelic/api/web_transaction.py
```diff
@@ -329,25 +323,9 @@ def process_response(self, status_code, response_headers):
 
         try:
             self._response_code = int(status_code)
-
-            # If response code is 304 do not insert CAT headers. See:
-            # https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.5
-            if self._response_code == 304:
-                return []
```
TimPansino, line 336: Why is this removed? Read the HTTP spec on 304, and it seems like this is still very much true that we should not insert DT headers on a 304.

tests/framework_tornado/_target_application.py
```diff
@@ -250,8 +220,6 @@ def make_app(custom=False):
         (r"/init", InitializeHandler),
         (r"/html-insertion", HTMLInsertionHandler),
         (r"/bad-get-status", BadGetStatusHandler),
-        (r"/force-cat-response/(\S+)/(\S+)/(\S+)", ProcessCatHeadersHandler),
-        (r"/304-cat-response/(\S+)/(\S+)", ProcessCatHeadersHandler, {"response_code": 304}),
```
TimPansino, line 254: Not sure but this is probably still needed for testing 304 responses if the previous comment about 304 is valid?

TimPansino (changes requested):

newrelic/hooks/framework_aiohttp.py
```diff
@@ -70,18 +60,6 @@ def _is_expected(exc, value, tb):
...
-def _nr_process_response_proxy(response, transaction):
-    nr_headers = transaction.process_response(response.status, response.headers)
-
-    response._headers = HeaderProxy(response.headers, nr_headers)
-
-
-def _nr_process_response(response, transaction):
-    nr_headers = transaction.process_response(response.status, response.headers)
-
-    response._headers.update(nr_headers)
```
TimPansino, line 82: A lot of instrumentation removed from this file, but we still need it. We should be calling `process_response` to capture the status code for web transactions and the headers as agent attributes.

tests/framework_aiohttp/test_client_async_await.py
```diff
@@ -16,7 +16,6 @@
```
TimPansino, line 16: Could probably rename this file to just `test_client.py` now.

TimPansino: Also, we have a `test_server_dt.py` but no corresponding tests for outbound DT headers in the aiohttp client. We should add tests from this, as we've just removed the `test_client_cat.py` file that has the only thing remotely resembling tests for this.

tests/framework_aiohttp/test_client.py
TimPansino: Why was this fine? Shouldn't we be keeping the behavior of these assertions? (Context: https://github.com/newrelic/newrelic-python-agent/pull/1489#discussion_r2377132527)

lrafeei: Basically, we are not testing the equivalency of these two anymore because they will not have equivalent headers, namely, the trace IDs of the two different tasks. (So, `traceparent`, `traceheader`, and `newrelic` headers will all have a dependency on the trace_id, which will be different per task)

newrelic/api/web_transaction.py
TimPansino: It has since occurred to me that this function no longer inserts headers of any kind because DT is request based, and doesn't edit response headers. Ignore all the 304 related comments.

newrelic/hooks/framework_aiohttp.py
```diff
@@ -381,9 +356,6 @@ async def _coro(*_args, **_kwargs):
 
 def instrument_aiohttp_web(module):
     global _nr_process_response
```
TimPansino:
```suggestion
```

newrelic/hooks/framework_aiohttp.py
```diff
@@ -279,16 +263,13 @@ async def _coro():
...
             return response
-        except Exception as e:
-            try:
-                trace.process_response_headers(e.headers.items())
-            except:
-                pass
+        except Exception:
+            notice_error()
 
             raise
```
TimPansino: I don't think we want to call `notice_error` here necessarily, it's not equivalent to what we had before. I think this outer try/except can be deleted and we end up with just the success case.

```
response = await wrapped(*args, **kwargs)

try:
    trace.process_response(status_code=response.status)
except:
    pass

return response
```

TimPansino: @lrafeei Github was fighting me trying to suggest too many things so I pushed it as a commit, see if you agree with everything here. https://github.com/newrelic/newrelic-python-agent/pull/1489/commits/ad3d4e6cf50a627c76c68f1e35dc6aa12bf568e5

tests/testing_support/validators/validate_synthetics_event.py
```diff
@@ -42,8 +42,6 @@ def _flatten(event):
 
                 flat_event = _flatten(event)
 
-                assert "nr.guid" in flat_event, f"name=nr.guid, event={flat_event!r}"
```
TimPansino, line 45: This shouldn't be deleted

newrelic/core/transaction_node.py
TimPansino: Are we confident this isn't used in the UI anywhere?

newrelic/config.py
```diff
@@ -3116,12 +3106,6 @@ def _process_module_builtin_defaults():
 
     _process_module_definition("aiohttp.wsgi", "newrelic.hooks.framework_aiohttp", "instrument_aiohttp_wsgi")
     _process_module_definition("aiohttp.web", "newrelic.hooks.framework_aiohttp", "instrument_aiohttp_web")
-    _process_module_definition(
-        "aiohttp.web_reqrep", "newrelic.hooks.framework_aiohttp", "instrument_aiohttp_web_response"
-    )
-    _process_module_definition(
-        "aiohttp.web_response", "newrelic.hooks.framework_aiohttp", "instrument_aiohttp_web_response"
-    )
```
TimPansino, line 3326: Don't we still need these?

tests/testing_support/fixtures.py
```diff
@@ -811,7 +805,6 @@ def _bind_params(transaction, *args, **kwargs):
                 assert intrinsics["error.class"] == required_attrs["error.class"]
                 assert intrinsics["error.message"].startswith(required_attrs["error.message"])
                 assert intrinsics["error.expected"] == required_attrs["error.expected"]
-                assert intrinsics["nr.transactionGuid"] is not None
```
TimPansino, line 818: As far as I can tell this still exists and we need this line

tests/agent_features/test_attributes_in_action.py
```diff
@@ -165,30 +155,27 @@ def normal_application(request):
 # Tests for checking the presence and format of agent attributes.
 # Test default settings.
 
-_expected_attributes = {"agent": TRACE_ERROR_AGENT_KEYS, "user": ERROR_USER_ATTRS, "intrinsic": ["trip_id"]}
+_expected_attributes = {"agent": TRACE_ERROR_AGENT_KEYS, "user": ERROR_USER_ATTRS, "intrinsic": ["guid"]}
```
TimPansino, line 158: All of these changes where we deleted trip_id are breaking this file, it's supposed to test for the behavior of intrinsics. I replaced them all with guid and that should solve all those issues.

tests/testing_support/fixtures.py
lrafeei: You're right.  Even though this was initially a CAT attribute, it is also used for error events.

newrelic/hooks/framework_aiohttp.py
lrafeei: I don't see any issues with that commit

hmstepanek (approved):

PR #1490, opened 2025-09-12
Add deprecation warning for modules

lrafeei:
This PR adds a section to `newrelic/config.py` that issues a deprecation warning when a currently unsupported framework is still being used by the customer.

### DIFF
--- a/newrelic/config.py
+++ b/newrelic/config.py
@@ -20,8 +20,8 @@ import sys
 import threading
 import time
 import traceback
+from datetime import datetime, timezone
 from pathlib import Path
-from datetime import datetime
 
 import newrelic.api.application
 import newrelic.api.background_task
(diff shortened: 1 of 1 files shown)


newrelic/config.py
```diff
@@ -1105,6 +1106,19 @@ def _module_import_hook(target, module, function):
     def _instrument(target):
         _logger.debug("instrument module %s", ((target, module, function),))
 
+        # Deprecation warning for archived/unsupported modules
+        library_name = target.__package__.split(".")[0]
+        deprecated_modules = {"aioredis": datetime(2022, 2, 22, 0, 0, tzinfo=timezone.utc)}
```
TimPansino: Can we move this line to the module level so it's not constantly redefined?

lrafeei: Fair point--I moved the `deprecated_modules` dict to the module level as a global variable

TimPansino (approved):

PR #1491, opened 2025-09-13
Drop Python 3.7 Support

lrafeei:
This PR removes support for Python 3.7 as well as removing the monolithic C extension for Python 2.7 (which has been unused since New Relic dropped Python 2.7 support in v10.0.0)


TimPansino:
Checked this over and everything looked good except the changes to `setup.py` which was definitely heavy on merge conflicts. I fixed it all and ran it with older Python versions to make sure we're still getting the correct errors.

umaannamalai (approved):

PR #1496, opened 2025-09-15
Add deprecation warning to WSGI environ dict settings

lrafeei:
Add deprecation warnings for the following New Relic specific WSGI environ dictionary settings:

- `newrelic.set_background_task`
- `newrelic.suppress_apdex_metric`
- `newrelic.suppress_transaction_trace`
- `newrelic.capture_request_params`
- `newrelic.disable_browser_autorum`

### DIFF
--- a/newrelic/api/web_transaction.py
+++ b/newrelic/api/web_transaction.py
@@ -87,7 +87,9 @@ def _lookup_environ_setting(environ, name, default=False):
     # Check for deprecated WSGI environ dictionary setting
     if name in DEPRECATED_ENVIRON_DICT:
         warnings.warn(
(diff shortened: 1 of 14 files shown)


newrelic/api/web_transaction.py
```diff
@@ -71,11 +71,27 @@ def _parse_time_stamp(time_stamp):
 TRUE_VALUES = {"on", "true", "1"}
 FALSE_VALUES = {"off", "false", "0"}
 
+DEPRECATED_ENVIRON_DICT = (
```
TimPansino: This isn't a dict anymore, can we rename it?

lrafeei: Whoops, good call.  Done!

PR #1499, opened 2025-09-17
Update deprecated urllib APIs

lrafeei:
Since Python 3.3, `urllib.URLopener()` has been deprecated in favor of `urllib.request.urlopen`.  In Python 3.14, `urllib.URLopener()` will be removed. `urllib.request.urlopen` corresponds to the old `urllib2.urlopen`

For more information: https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen

### DIFF
--- a/tests/external_httplib/test_urllib.py
+++ b/tests/external_httplib/test_urllib.py
@@ -12,14 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
-
 import pytest
 
-try:
-    import urllib.request as urllib
-except:
-    import urllib
+from urllib.request import urlopen as urlopen, urlretrieve
 
 from testing_support.external_fixtures import cache_outgoing_headers, insert_incoming_headers
 from testing_support.fixtures import cat_enabled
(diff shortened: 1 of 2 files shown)


tests/external_httplib/test_urllib.py
```diff
@@ -53,8 +51,8 @@ def test_urlopener_http_request(server, metrics):
     )
     @background_task(name="test_urllib:test_urlopener_http_request")
     def _test():
-        opener = urllib.URLopener()
-        opener.open(f"http://localhost:{server.port}/")
+        with urlopen(f"http://localhost:{server.port}/") as response:
+            assert response.status == 200
```
TimPansino: I don't think any of the changes to this file are correct. Testing `urlopen` should be done in the tests for `urllib2`, which has all of these tests already.

We still want to be testing `urllib` on Python 3.13 and below, since it's real instrumentation customers can still use.

I think the correct change to make here is to mark tests for URLopener with a pytest skip if we're on Python 3.14 or higher. To make that possible, put back the imports to not use `from` so we don't get import errors.

TimPansino (changes requested):

TimPansino (approved):

PR #1504, opened 2025-09-19
Add response token count logic to Gemini instrumentation.

umaannamalai:
This PR updates sync and async Bedrock instrumentation and tests to pull token counts directly from the response object. If a customer has an LLM token counting callback function registered, this will still take priority over the response token count values.

It also removes token count calculations from error cases as the agent will consistently stop reporting token counts when a error was met during the creation of a text generation.


newrelic/hooks/external_botocore.py
```diff
@@ -272,41 +277,56 @@ def create_chat_completion_message_event(
...
     return bedrock_attrs
 
 
 def extract_bedrock_titan_text_model_response(response_body, bedrock_attrs):
     if response_body:
         response_body = json.loads(response_body)
 
+        input_tokens = response_body.get("inputTextTokenCount", 0)
+        completion_tokens = sum(result["tokenCount"] for result in response_body.get("results", []))
+        if not completion_tokens:
```
hmstepanek: Is this line necessary? It looks like `sum([])` is 0 already.

newrelic/hooks/external_botocore.py
```diff
@@ -272,41 +277,56 @@ def create_chat_completion_message_event(
...
+
     return bedrock_attrs
 
 
 def extract_bedrock_titan_text_model_response(response_body, bedrock_attrs):
     if response_body:
         response_body = json.loads(response_body)
 
+        input_tokens = response_body.get("inputTextTokenCount", 0)
+        completion_tokens = sum(result["tokenCount"] for result in response_body.get("results", []))
```
hmstepanek:
```suggestion
        completion_tokens = sum(result.get("tokenCount", 0) for result in response_body.get("results", []))
```

newrelic/hooks/external_botocore.py
```diff
@@ -272,41 +277,56 @@ def create_chat_completion_message_event(
...
+        if not completion_tokens:
+            completion_tokens = 0
+
+        total_tokens = input_tokens + completion_tokens
+
         output_message_list = [
             {"role": "assistant", "content": result["outputText"]} for result in response_body.get("results", [])
```
hmstepanek:
```suggestion
            {"role": "assistant", "content": result.get("outputText")} for result in response_body.get("results", [])
```

newrelic/hooks/external_botocore.py
```diff
@@ -342,10 +351,42 @@ def extract_bedrock_mistral_text_model_streaming_response(response_body, bedrock
...
+
+    bedrock_attrs["input_message_list"] = input_message_list
+    bedrock_attrs["request.max_tokens"] = request_config.get("maxTokenCount")
+    bedrock_attrs["request.temperature"] = request_config.get("temperature")
+
+    return bedrock_attrs
 
-    bedrock_attrs["input"] = request_body.get("inputText")
+
+def extract_bedrock_titan_text_model_streaming_response(response_body, bedrock_attrs):
```
hmstepanek, line 364: Can you for the sake of being able to easily identify what's changed here: re-order these changes in the file such that they end up being diffed against the original function. Right now it's really hard to compare to see what's changed because the diff isn't showing a 1:1 comparison. I know it's a pain but in a larger PR like this with lots of files changed it makes it a ton easier to review.

newrelic/hooks/external_botocore.py
```diff
@@ -1014,23 +1132,28 @@ def handle_embedding_event(transaction, bedrock_attrs):
     model = bedrock_attrs.get("model", None)
     input_ = bedrock_attrs.get("input")
 
+    response_total_tokens = bedrock_attrs.get("response.usage.total_tokens", None)
```
hmstepanek: None is redundant.
```suggestion
    response_total_tokens = bedrock_attrs.get("response.usage.total_tokens")
```

newrelic/hooks/external_botocore.py
```diff
@@ -1055,12 +1179,35 @@ def handle_chat_completion_event(transaction, bedrock_attrs):
     response_id = bedrock_attrs.get("response_id", None)
     model = bedrock_attrs.get("model", None)
 
+    response_prompt_tokens = bedrock_attrs.get("response.usage.prompt_tokens", None)
```
hmstepanek:
```suggestion
    response_prompt_tokens = bedrock_attrs.get("response.usage.prompt_tokens")
```

newrelic/hooks/external_botocore.py
```diff
@@ -1055,12 +1179,35 @@ def handle_chat_completion_event(transaction, bedrock_attrs):
...
+    response_completion_tokens = bedrock_attrs.get("response.usage.completion_tokens", None)
```
hmstepanek:
```suggestion
    response_completion_tokens = bedrock_attrs.get("response.usage.completion_tokens")
```

newrelic/hooks/external_botocore.py
```diff
@@ -1055,12 +1179,35 @@ def handle_chat_completion_event(transaction, bedrock_attrs):
...
+    response_total_tokens = bedrock_attrs.get("response.usage.total_tokens", None)
```
hmstepanek:
```suggestion
    response_total_tokens = bedrock_attrs.get("response.usage.total_tokens")
```

newrelic/hooks/external_botocore.py
```diff
@@ -1055,12 +1179,35 @@ def handle_chat_completion_event(transaction, bedrock_attrs):
...
+
     input_message_list = bedrock_attrs.get("input_message_list", [])
     output_message_list = bedrock_attrs.get("output_message_list", [])
     number_of_messages = (
         len(input_message_list) + len(output_message_list)
     ) or None  # If 0, attribute will be set to None and removed
 
+    input_message_content = " ".join([msg.get("content", "") for msg in input_message_list if msg.get("content")])
```
hmstepanek: I don't think the default values of "" are reachable here given the if msg.get("content").

newrelic/hooks/external_botocore.py
```diff
@@ -224,6 +225,8 @@ def create_chat_completion_message_event(
             "vendor": "bedrock",
             "ingest_source": "Python",
         }
+        if all_token_counts:
+            chat_completion_message_dict["token_count"] = 0
```
hmstepanek, line 229: Is this the value that was previously added in the pipeline? Are we just setting it to 0 as a transition?

umaannamalai: Yeah, this is just there for backwards compatibility. This zero-value attribute is required because it signals the ingest pipeline tokenizer to not run since token counts are already present on the summary event.

PR #1523, opened 2025-09-29
Fix CI Image

TimPansino:
# Overview

Various things were broken in the latest version of the CI image. The following changes have repaired it.

* Fix `PyPy` and `CPython` competing for `python3.10` binary by renaming the `PyPy` binary during install.
* Fix shims for default `Python` version by using `uv python install --default` to create multiple shims instead of just the `python3.13` shim.
* Set `UV_LINK_MODE=copy` to silence warnings around the cache being mounted in Docker.
* Replacing all the deleted development headers from Ubuntu.
  * We may want to remove any unused headers in a future version, but for now putting these back.
* Remove the custom `pycurl` reinstall command in `tornado` tests.
  * This is leftover from when we needed to build these from scratch and is now unnecessary.
* Fixed azure-functions install flow by adding `pip` to the created venvs through `uv_seed`.

# Testing

* [Example Test Run](https://github.com/TimPansino/newrelic-python-agent/actions/runs/18108599889/job/51529210805), this time verified to be using the changed CI image.
  * Coverage fails on the fork due to permissions but that can be ignored.

### DIFF
--- a/.github/containers/Dockerfile
+++ b/.github/containers/Dockerfile
@@ -107,18 +107,20 @@ WORKDIR "${HOME}"
 RUN curl -LsSf https://astral.sh/uv/install.sh | sh
 ENV PATH="${HOME}/.local/bin:${PATH}"
 ENV UV_PYTHON_PREFERENCE="only-managed"
+ENV UV_LINK_MODE="copy"
 
-# Install Python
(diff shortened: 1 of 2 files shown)


lrafeei:
Running it locally, mysql-datastore_aiomysql, mysql-datastore_pymysql, and python-adapter_mcp all worked on pypy310 now.  Could we add those back in?  Or should we just make a separate PR for that?

lrafeei (approved):

PR #1524, opened 2025-09-29
Agent API Benchmarks

TimPansino:
# Overview

* Delete placeholder benchmark file.
* Add benchmark file for all relevant public APIs in `newrelic.agent`.
  * Note: There are a couple TODOs that are marked out, those will need to be fixed in the future to fully cover `newrelic.agent`.
* Tweak behavior of developer mode and OTLP to allow benchmarking `record_dimensional_metric[s]`.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -20,13 +20,13 @@ FROM ghcr.io/newrelic/newrelic-python-agent-ci:${IMAGE_TAG}
 USER root
 ARG UID=1001
 ARG GID=${UID}
-ENV HOME=/home/vscode
+ENV HOME=/root
 RUN mkdir -p ${HOME} && \
     groupadd --gid ${GID} vscode && \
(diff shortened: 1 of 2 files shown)


tests/agent_benchmarks/bench_agent_apis.py
```diff
@@ -0,0 +1,482 @@
...
+    @background_task()
+    def bench_record_custom_event(self):
+        record_custom_event("TestEvent", {"attribute1": "value1", "attribute2": 2})
+
+    @background_task()
+    def bench_record_exception(self):
+        pass
+
+    @background_task()
+    def bench_record_llm_feedback_event(self):
```
umaannamalai, line 267: Was there a reason these were not being run with test data?

TimPansino: Must have lost that commit when rebasing. Fixed this and added the API calls back here.

PR #1532, opened 2025-10-02
Add support for partial granularity type

hmstepanek:
# Overview
This adds support for the concept of partial granularity distributed traces. There are 3 types:
* reduced
* essential
* compact

Add support for the following new configuration options:
```
distributed_tracing
   sampler
       adaptive_sampling_target
       full_granularity
           enabled
           remote_parent_sampled
           remote_parent_not_sampled
       partial_granularity
           enabled
           type
           remote_parent_sampled
           remote_parent_not_sampled
```

### DIFF
--- a/tests/agent_features/test_event_loop_wait_time.py
+++ b/tests/agent_features/test_event_loop_wait_time.py
@@ -27,7 +27,7 @@ from newrelic.api.transaction import current_transaction
 from newrelic.core.trace_cache import trace_cache
 
 
-@pytest.fixture()
+@pytest.fixture
 def event_loop():
     from asyncio import new_event_loop, set_event_loop
 


newrelic/core/node_mixin.py
```diff
@@ -68,18 +68,85 @@ def span_event(self, settings, base_attrs=None, parent_guid=None, attr_class=dic
...
+                current_start_time = ct_exit_spans[span_attrs][0]["timestamp"]
+                current_end_time = ct_exit_spans[span_attrs][0]["timestamp"]/1000 + ct_exit_spans[span_attrs][1]["nr.durations"]
+                new_start_time = i_attrs["timestamp"]
+                new_end_time = i_attrs["timestamp"]/1000 + i_attrs["duration"]
+                set_start_time = min(new_start_time, current_start_time)
+                # If the new span starts after the old span's end time or the new span
+                # ends before the current span starts; add the durations.
+                if current_end_time < new_start_time/1000 or new_end_time < current_start_time/1000:
+                    set_duration = ct_exit_spans[span_attrs][1]["nr.durations"] + i_attrs["duration"]
+                # Otherwise, if the new and old span's overlap in time, use the newest
```
hmstepanek, line 139:
```suggestion
                # Otherwise, if the new and old spans overlap in time, use the newest
```

newrelic/core/config.py
```diff
@@ -1369,9 +1399,29 @@ def finalize_application_settings(server_side_config=None, settings=_settings):
...
+    # The `distributed_tracing.sampler.full_granularity` path takes precedence.
+    # To simplify logic in the code that uses these settings, store the values that
+    # should be used at the `distributed_tracing.sampler.full_granularity` path.
+    if not settings.distributed_tracing.sampler.full_granularity.remote_parent_sampled:
+        settings.distributed_tracing.sampler.full_granularity.remote_parent_sampled = settings.distributed_tracing.sampler.remote_parent_sampled
+    if not settings.distributed_tracing.sampler.full_granularity.remote_parent_not_sampled:
+        settings.distributed_tracing.sampler.full_granularity.remote_parent_not_sampled = settings.distributed_tracing.sampler.remote_parent_not_sampled
+    # Partial granularity tracing is not available in infinite tracing mode.
+    if settings.infinite_tracing.enabled and settings.distributed_tracing.sampler.partial_granularity.enabled:
+        logger.warning(
```
hmstepanek:
```suggestion
        _logger.warning(
```

newrelic/api/transaction.py
```diff
@@ -285,7 +285,7 @@ def __init__(self, application, enabled=None, source=None):
         self.tracestate = ""
         self._priority = None
         self._sampled = None
-        self._traceparent_sampled = None
+        self._remote_parent_sampled = None
```
hmstepanek, line 289:
```suggestion
        # Remote parent sampled is set from the W3C parent header or the Newrelic header if no W3C parent header is present.
        self._remote_parent_sampled = None
```

newrelic/core/config.py
```diff
@@ -1371,9 +1402,34 @@ def finalize_application_settings(server_side_config=None, settings=_settings):
...
+    if not settings.distributed_tracing.sampler.full_granularity.remote_parent_not_sampled:
+        settings.distributed_tracing.sampler.full_granularity.remote_parent_not_sampled = (
+            settings.distributed_tracing.sampler.remote_parent_not_sampled
+        )
+    # Partial granularity tracing is not available in infinite tracing mode.
+    if settings.infinite_tracing.enabled and settings.distributed_tracing.sampler.partial_granularity.enabled:
+        _logger.warning(
+            "Improper configuration. Infinite tracing cannot be enabled at the same time as partial granularity tracing. Setting distributed_tracing.sampler.partial_granularity.enabled=False."
+        )
+        settings.distributed_tracing.sampler.partial_granularity.enabled = False
```
lrafeei: Since these settings aren't configurable in server-side configs, you might be able to just put those settings in `translate_deprecated_settings()` which is used earlier in the settings configuration process (where we are still parsing local settings). This way, there is not a concern about a potential mismatch between global and application settings later.

newrelic/core/config.py
```diff
@@ -1371,9 +1396,20 @@ def finalize_application_settings(server_side_config=None, settings=_settings):
 
     application_settings.attribute_filter = AttributeFilter(flatten_settings(application_settings))
 
+    simplify_distributed_tracing_sampler_granularity_settings(application_settings)
```
lrafeei: I just realized that `simplify_distributed_tracing_sampler_granularity_settings()` should either
1. be moved before the `attribute_filter` application, but right after `apply_server_side_settings()`
2. be removed all together and put the logic from that function in `apply_server_side_settings()` (maybe after the `custom_insights_events.max_attribute_value` override?)

newrelic/core/config.py
```diff
@@ -1371,6 +1407,8 @@ def finalize_application_settings(server_side_config=None, settings=_settings):
...
```
lrafeei: Since the logic for this has now moved to `apply_server_side_settings()`, the call to this function can be removed

lrafeei (approved):

PR #1537, opened 2025-10-03
Fix max_samples_stored settings to still allow server side config to override settings.

lrafeei:
Changing the `*.max_samples_stored` settings from `event_harvest_config.harvest_limits.*` in Python agent v11.0.0 resulted in the server side configurations not being properly applied to the agent, resulting in logs (and other events) being sent by the agent anyway. 

This is not visible to users because the logging team has a block on this.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -229,6 +229,9 @@ jobs:
   windows:
     env:
       TOTAL_GROUPS: 1
+      UV_PYTHON: "3.13"
+      UV_PYTHON_DOWNLOADS: "never"
+      UV_PYTHON_PREFERENCE: "only-system"
 
     strategy:
       fail-fast: false
@@ -249,11 +252,12 @@ jobs:
         with:
           python-version: "3.13"
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # 6.8.0
+
       - name: Install Dependencies
         run: |
-          pip install --upgrade tox
-        env:
-          GROUP_NUMBER: ${{ matrix.group-number }}
+          uv tool install tox --with tox-uv
 
       - name: Get Environments
         id: get-envs
@@ -281,6 +285,9 @@ jobs:
   windows_arm64:
     env:
       TOTAL_GROUPS: 1
+      UV_PYTHON: "3.13"
+      UV_PYTHON_DOWNLOADS: "never"
+      UV_PYTHON_PREFERENCE: "only-system"
 
     strategy:
       fail-fast: false
@@ -301,11 +308,12 @@ jobs:
         with:
           python-version: "3.13"
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # 6.8.0
+
       - name: Install Dependencies
         run: |
-          pip install --upgrade tox
-        env:
-          GROUP_NUMBER: ${{ matrix.group-number }}
+          uv tool install tox --with tox-uv
 
       - name: Get Environments
         id: get-envs


newrelic/config.py
```diff
@@ -638,6 +638,62 @@ def delete_setting(settings_object, name):
...
+
+    Here, the user will still get warnings about deprecated settings
+    being used.  However, the agent will also translate the settings
+    to their deprecated `event_harvest_config` counterparts during
+    the configuration process.
+    """
+
+    cached = dict(cached_settings)
+
+    # breakpoint()
```
TimPansino:
```suggestion
```

newrelic/config.py
```diff
@@ -669,19 +725,7 @@ def translate_deprecated_settings(settings, cached_settings):
     cached = dict(cached_settings)
 
     deprecated_settings_map = [
-        ("transaction_tracer.capture_attributes", "transaction_tracer.attributes.enabled"),
-        ("error_collector.capture_attributes", "error_collector.attributes.enabled"),
-        ("browser_monitoring.capture_attributes", "browser_monitoring.attributes.enabled"),
-        ("analytics_events.capture_attributes", "transaction_events.attributes.enabled"),
-        ("analytics_events.enabled", "transaction_events.enabled"),
-        ("analytics_events.max_samples_stored", "transaction_events.max_samples_stored"),
```
TimPansino, line 677: I think this single setting (`analytics_events.max_samples_stored`) might be a problem, it's currently unhandled. We might need a custom block of code just for it.

tests/agent_features/test_configuration.py
```diff
@@ -405,56 +405,6 @@ def test_delete_setting_parent():
...
-    ),
-    (
-        TSetting("analytics_events.capture_attributes", False, True),
-        TSetting("transaction_events.attributes.enabled", True, True),
-    ),
-    (TSetting("analytics_events.enabled", True, True), TSetting("transaction_events.enabled", False, True)),
-    (TSetting("analytics_events.enabled", False, True), TSetting("transaction_events.enabled", True, True)),
-    (
-        TSetting("analytics_events.max_samples_stored", 1200, 1200),
-        TSetting("transaction_events.max_samples_stored", 9999, 1200),
```
TimPansino, line 452: Some test akin to this will also still be needed. Might not be exactly the same.

tests/agent_features/test_configuration.py
```diff
@@ -498,6 +448,9 @@ def test_delete_setting_parent():
 ]
 
 
+@pytest.mark.skip(
+    "Behavior for the event_harvest_config will be different than typical deprecated settings.  Renable this test once there are other deprecated settings."
+)
```
TimPansino: We should probably have 2 separate test functions here.

1. Testing the event_harvest_config settings are translated correctly if you use either the old or new names. Should check that `analytics_events.max_samples_stored` ends up as `event_harvest_config.harvest_limits.analytic_event_data` as well.
2. Testing the other deprecated settings which currently don't exist.

tests/agent_unittests/test_agent_protocol.py
```diff
@@ -474,6 +474,10 @@ def test_connect(
     assert HttpClientRecorder.STATE == 0
 
 
+def test_connect_logging():
+    pass
```
TimPansino: Is this supposed to be here?

newrelic/config.py
lrafeei: I thought since it was removed, we no longer needed to incorporate the usage of that into the code

tests/agent_unittests/test_agent_protocol.py
lrafeei: Nope!

tests/agent_features/test_configuration.py
lrafeei: Done!

tests/agent_features/test_configuration.py
lrafeei: Since we removed `analytics_events.max_samples_stored`, do we still need to account for it in tests?

PR #1540, opened 2025-10-07
Python 3.14 Support

TimPansino:
# Overview

* Add testing, benchmarking, wheels, and classifiers for CPython 3.14.
* Upgrade PyPy testing version to 3.11.
* Fix `asyncio` hooks for Python 3.14.
* Fix various tests failures on Python 3.14.
* Add not yet functional frameworks' test environments to `tox.ini` as comments.
* Add summary reporter for `tox` tests to `tests.yml` workflow.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -309,7 +309,7 @@ jobs:
             3.14
 
       - name: Install uv
-        uses: astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # 6.8.0
+        uses: astral-sh/setup-uv@3259c6206f993105e3a61b142c2d97bf4b9ef83d # 7.1.0
 
       - name: Install Dependencies
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -374,26 +460,40 @@ jobs:
...
         with:
           name: coverage-${{ github.job }}-${{ strategy.job-index }}
           path: ./**/.coverage.*
           include-hidden-files: true
           if-no-files-found: error
           retention-days: 1
 
+      - name: Upload Results Artifacts
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # 4.6.2
+        if: always()
```
hmstepanek, line 489: Do you even need this if here?

hmstepanek: Just looked this up. Looks like [it's used](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#always) to still run even when canceled so I guess for that reason it makes sense to keep it.

TimPansino: Yes this is intended to always upload the artifacts if they exist so we capture failures and can summarize them.

PR #1544, opened 2025-10-15
Fix Arbitrary Properties of kombu Producer.publish

TimPansino:
# Overview

* Fix a bug where instrumentation for `kombu.messaging.Producer.publish()` was causing properties to accidentally be nested inside of the dictionary returned by `bind_args()`, causing the properties to be in the wrong spot and not read by downstream packages.
  * Elevated these back to the top level of the dictionary to fix the issue.

### DIFF
--- a/tests/datastore_rediscluster/test_uninstrumented_rediscluster_methods.py
+++ b/tests/datastore_rediscluster/test_uninstrumented_rediscluster_methods.py
@@ -138,6 +138,7 @@ IGNORED_METHODS = {
     "result_callbacks",
     "set_default_node",
     "user_on_connect_func",
+    "startup_nodes",
 }
 
 REDIS_MODULES = {"bf", "cf", "cms", "ft", "graph", "json", "tdigest", "topk", "ts", "vset"}


tests/messagebroker_kombu/test_producer.py
```diff
@@ -105,3 +106,24 @@ def test():
             )
 
     test()
+
+
+def test_custom_properties(exchange, queue, events, get_consumer_record, producer):
+    validate_custom_properties_called = []
```
hmstepanek, line 112:
```suggestion
    validate_custom_properties_present = []
```

tests/messagebroker_kombu/test_producer.py
```diff
@@ -105,3 +106,24 @@ def test():
...
+
+    @transient_function_wrapper("kombu.messaging", "Producer._publish")
+    def validate_custom_properties(wrapped, instance, args, kwargs):
+        from newrelic.common.signature import bind_args
+
+        bound_args = bind_args(wrapped, args, kwargs)
+        properties = bound_args["properties"]
+        assert properties.get("custom_property", "") == "baz", "custom_property was deleted by instrumentation."
+        validate_custom_properties_called.append(True)
```
hmstepanek, line 121:
```suggestion
        validate_custom_properties_present.append(True)
```

tests/messagebroker_kombu/test_producer.py
```diff
@@ -105,3 +106,24 @@ def test():
...
+
+    @background_task()
+    @validate_custom_properties
+    def test():
+        producer.publish({"foo": 123}, exchange=exchange, routing_key="bar", declare=[queue], custom_property="baz")
+
+    test()
+    assert validate_custom_properties_called, "validate_custom_properties was not called."
```
hmstepanek, line 129:
```suggestion
    assert validate_custom_properties_present, "custom properties were not passed by new relic wrapper."
```

hmstepanek (changes requested):
I just had some suggestions for the naming of some things in the test-otherwise looks good.

tests/messagebroker_kombu/test_producer.py
TimPansino: This isn't quite right. `validate_custom_properties_called` is not a flag for validating that the properties are actually present. It's to validate that the function `validate_custom_properties` was actually called. It's verifying that the transient function wrapper is actually applied and functioning correctly, where otherwise this would silently pass if the transient function wrapper was being overwritten or unapplied somehow.

This is line inside `validate_custom_properties` is the one that actually does the validation of the custom properties being added.

```python
assert properties.get("custom_property", "") == "baz", "custom_property was deleted by instrumentation."
```

hmstepanek: Oh ok I see what you are doing now. Ok-that's fine then.

hmstepanek (approved):

PR #1553, opened 2025-10-23
fix(aiomysql): avoid wrapping pooled connections multiple times

canonrock16:
# Overview
- prevent repeatedly wrapping aiomysql pooled connections with `AsyncConnectionWrapper`
- avoid building deep ObjectProxy chains that led to `RecursionError: maximum recursion depth exceeded`
- add a regression unit test (`tests/agent_unittests/test_database_aiomysql.py`) that simulates a pool returning the same connection twice

# Related Github Issue
None.
Related forum thread: https://support.newrelic.com/s/hubtopic/aAXPh000000EHIjOAO/recursionerror-when-new-relic-python-agent-instruments-aiomysql-still-happening

# Testing
- tox -c tox.ini -e python-agent_unittests-py312-without_extensions
All tests passed including the new test I added

### DIFF
--- a/newrelic/hooks/database_aiomysql.py
+++ b/newrelic/hooks/database_aiomysql.py
@@ -79,7 +79,7 @@ def wrap_pool__acquire(dbapi2_module):
             connection = await wrapped(*args, **kwargs)
             connection_kwargs = getattr(instance, "_conn_kwargs", {})
 
-            if isinstance(connection, AsyncConnectionWrapper):
+            if hasattr(connection, "__wrapped__"):
                 return connection
 
             return AsyncConnectionWrapper(connection, dbapi2_module, (((), connection_kwargs)))


canonrock16:
Reproduction details

Agent 10.17.0 / aiomysql 0.2.0 / Python 3.12
Applied load against the service with hey; during the run the service began returning 500s and CloudWatch logs filled with RecursionError: maximum recursion depth exceeded
Full write-up: https://support.newrelic.com/s/hubtopic/aAXPh000000EHIjOAO/recursionerror-when-new-relic-python-agent-instruments-aiomysql-still-happening

Sample log :
```
2025-10-22T08:52:10.099Z RecursionError: maximum recursion depth exceeded
2025-10-22T08:52:10.184Z ERROR:root:unexpected error
File "<internal_path>/server.py", line 105, in wrap_error
return await func(*args, **kwargs)
...
File "<internal_path>/newrelic/hooks/database_aiomysql.py", line 79, in _wrap_pool__acquire
connection = await wrapped(*args, **kwargs)
File "<internal_path>/aiomysql/pool.py", line 171, in _fill_free_pool
self._loop.time() - conn.last_usage > self._recycle)
File "<internal_path>/newrelic/common/object_wrapper.py", line 70, in getattr
return super().getattr(name)
[previous line repeated 322 more times]
RecursionError: maximum recursion depth exceeded
```
After this patch, the same load no longer produces RecursionErrors and the service stays healthy.

canonrock16:
> I'll handle moving the tests for this to the right place and making sure they're working, thanks for finding this though.
> 
> Really curious how `Pool._acquire()` is maintaining the wrappers such that they're preserved for future calls. I'll have to read the codebase again and look into it.

For reference, aiomysql’s pool implementation reuses the same connection object pulled from the free queue:

- `_acquire`: https://github.com/aio-libs/aiomysql/blob/v0.2.0/aiomysql/pool.py#L118-L138
- `_fill_free_pool`: https://github.com/aio-libs/aiomysql/blob/v0.2.0/aiomysql/pool.py#L140-L196

`_acquire()` pops a connection from `self._free`, and `_fill_free_pool()` rotates existing entries unless they’re closed or expired. As a result, the same instance that was returned to the pool is handed back on the next `acquire()`, so the agent’s wrapper stays attached unless we detect it and avoid wrapping again.

TimPansino (approved):

PR #1563, opened 2025-10-30
Add Strands tools and agents instrumentation.

umaannamalai:
This PR adds instrumentation for Strands tools and agents. It also includes tests for single agent use cases.

### DIFF
--- a/newrelic/hooks/mlmodel_strands.py
+++ b/newrelic/hooks/mlmodel_strands.py
@@ -37,7 +37,6 @@ AGENT_EVENT_FAILURE_LOG_MESSAGE = "Exception occurred in Strands instrumentation
 TOOL_EXTRACTOR_FAILURE_LOG_MESSAGE = "Exception occurred in Strands instrumentation: Failed to extract tool information. If the issue persists, report this issue to New Relic support.\n"
 
 
-
 def wrap_agent__call__(wrapped, instance, args, kwargs):
     trace = current_trace()
     if not trace:


newrelic/api/error_trace.py
```diff
@@ -43,17 +44,25 @@ def __exit__(self, exc, value, tb):
...
-        parent = current_trace()
+def ErrorTraceWrapper(wrapped, ignore=None, expected=None, status_code=None, async_wrapper=None):
+    def literal_wrapper(wrapped, instance, args, kwargs):
+        wrapper = async_wrapper if async_wrapper is not None else get_async_wrapper(wrapped)
+        if not wrapper:
+            parent = current_trace()
+            if not parent:
+                return wrapped(*args, **kwargs)
+        else:
+            parent = None
```
hmstepanek, line 60: Regardless of wether there is a wrapper or not shouldn't we still set the parent to the current trace if there is a current trace?

newrelic/api/error_trace.py
```diff
@@ -43,17 +44,25 @@ def __exit__(self, exc, value, tb):
...
 
-        if parent is None:
-            return wrapped(*args, **kwargs)
+        trace = ErrorTrace(ignore, expected, status_code, parent=parent)
+
+        if wrapper:
```
hmstepanek, line 64: I think we need some comments here explaining what is going on cause I'm not following this logic in here.

newrelic/hooks/mlmodel_strands.py
```diff
@@ -0,0 +1,474 @@
...
+    linking_metadata = get_trace_linking_metadata()
+    agent_id = str(uuid.uuid4())
+
+    try:
+        return_val = wrapped(*args, **kwargs)
+    except Exception:
+        raise
+
+    # For streaming responses, wrap with proxy and attach metadata
+    proxied_return_val = AsyncGeneratorProxy(
```
hmstepanek: Should we put this code in a try-catch?

newrelic/hooks/mlmodel_strands.py
```diff
@@ -0,0 +1,474 @@
...
+
+        try:
+            strands_attrs = getattr(self, "_nr_strands_attrs", {})
+
+            # If there are no strands attrs exit early as there's no data to record.
+            if not strands_attrs:
+                return
+
+            agent_name = strands_attrs.get("agent_name", "agent")
+            agent_id = strands_attrs.get("agent_id", None)
```
hmstepanek: I think this is the same as this:
```suggestion
            agent_id = strands_attrs.get("agent_id")
```

newrelic/hooks/mlmodel_strands.py
```diff
@@ -0,0 +1,474 @@
...
+        finally:
+            # Clear cached data to prevent memory leaks and duplicate reporting
+            if hasattr(self, "_nr_strands_attrs"):
+                self._nr_strands_attrs.clear()
+
+
+def _construct_base_tool_event_dict(strands_attrs, tool_results, transaction, linking_metadata):
+    try:
+        try:
+            tool_output = tool_results[-1]["content"][0] if tool_results else None
```
hmstepanek, line 186: I wonder if by combining these under the same try-catch we might be missing cases where there's an error but no content?

newrelic/hooks/mlmodel_strands.py
```diff
@@ -0,0 +1,474 @@
...
+            error = tool_results[-1]["status"] == "error"
+        except Exception:
+            tool_output = None
+            error = False
+            _logger.warning(TOOL_OUTPUT_FAILURE_LOG_MESSAGE, exc_info=True)
+
+        tool_name = strands_attrs.get("tool_name", "tool")
+        tool_id = strands_attrs.get("tool_id", None)
```
hmstepanek: Again, the get(key, None) seems redundant as get() on a dictionary already returns None if it can't find the key.

newrelic/hooks/mlmodel_strands.py
```diff
@@ -0,0 +1,474 @@
...
+            "id": agent_id,
+            "name": agent_name,
+            "span_id": linking_metadata.get("span.id"),
+            "trace_id": linking_metadata.get("trace.id"),
+            "vendor": "strands",
+            "ingest_source": "Python",
+        }
+        agent_event_dict.update(_get_llm_metadata(transaction))
+    except Exception:
+        agent_event_dict = {}
```
hmstepanek, line 240: Should we have a log message here as well?

newrelic/hooks/mlmodel_strands.py
```diff
@@ -0,0 +1,474 @@
...
+
+            # Notice the error on the function trace
+            self._nr_ft.notice_error(attributes={"tool_id": tool_id})
+            self._nr_ft.__exit__(*sys.exc_info())
+
+            # Create error event
+            tool_event_dict = _construct_base_tool_event_dict(
+                strands_attrs, tool_results, transaction, linking_metadata
+            )
+            tool_event_dict.update({"duration": self._nr_ft.duration * 1000})
```
hmstepanek: Would it be faster to do this instead for a single key-value pair?
```
tool_event_dict["duration"] = self._nr_ft.duration * 1000
```

newrelic/hooks/mlmodel_strands.py
```diff
@@ -0,0 +1,474 @@
...
+    settings = transaction.settings or global_settings()
+    if not settings.ai_monitoring.enabled:
+        return wrapped(*args, **kwargs)
+
+    # Framework metric also used for entity tagging in the UI
+    transaction.add_ml_model_info("Strands", STRANDS_VERSION)
+    transaction._add_agent_attribute("llm", True)
+
+    # Grab tool data
+    bound_args = bind_args(wrapped, args, kwargs)
```
hmstepanek: Maybe we should surround this in a try-catch?

newrelic/hooks/mlmodel_strands.py
```diff
@@ -0,0 +1,474 @@
...
+    linking_metadata = get_trace_linking_metadata()
+    tool_id = str(uuid.uuid4())
+
+    try:
+        return_val = wrapped(*args, **kwargs)
+    except Exception:
+        raise
+
+    # For streaming responses, wrap with proxy and attach metadata
+    proxied_return_val = AsyncGeneratorProxy(
```
hmstepanek: Same thing here-perhaps another try-catch?

newrelic/api/error_trace.py
umaannamalai: The idea here was that the if we had an async wrapper, then the trace context management would happen directly in the wrapper (ex: https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/common/async_wrapper.py#L25). This has been the same approach we take for other trace types such as[ ExternalTraces](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/api/external_trace.py#L101) and [FunctionTraces](https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/api/function_trace.py#L140)

newrelic/hooks/mlmodel_strands.py
umaannamalai: The inner try-except handles the case where there is no content accessible. The outer try-except was meant to handle any other errors that come up when trying to grab attribute data in the case that there was or wasn't content. If there wasn't content, we set safeguards and keep going without raising an error. So technically that would cover cases with and without content?

hmstepanek (approved):

PR #1564, opened 2025-10-31
Fix notice_error logic for non-iterable exceptions.

umaannamalai:
In some cases, caught exceptions passed to notice_error() are singular class errors and not iterables like the notice_error() logic currently expects. This PR updates logic to check if the errors passed to the API exist and are fully populated iterables before grabbing `sys.exc_info()` instead.


newrelic/api/time_trace.py
```diff
@@ -360,16 +360,23 @@ def _observe_exception(self, exc_info=None, ignore=None, expected=None, status_c
...
-        # Pull from sys.exc_info if no exception is passed
-        if not error or None in error:
+        # Pull from sys.exc_info() if no exception is passed
+        # Check that the error exists and that it is a fully populated iterable
+        if not error or none_in_error(error) or (error and not error_is_iterable(error)):
             error = sys.exc_info()
 
             # If no exception to report, exit
-            if not error or None in error:
+            if not error or none_in_error(error) or (error and not error_is_iterable(error)):
```
TimPansino, line 379: This line should only ever deal with the output of [sys.exc_info](https://docs.python.org/3/library/sys.html#sys.exc_info), which is guaranteed to be a tuple of `(None, None, None)` or `(exc, val, tb)`. This line is adding a bunch of unnecessary complexity when the original code here should have been fine.

newrelic/api/time_trace.py
```diff
@@ -360,16 +360,23 @@ def _observe_exception(self, exc_info=None, ignore=None, expected=None, status_c
...
+
         attributes = attributes if attributes is not None else {}
 
         # If no exception details provided, use current exception.
 
-        # Pull from sys.exc_info if no exception is passed
-        if not error or None in error:
+        # Pull from sys.exc_info() if no exception is passed
+        # Check that the error exists and that it is a fully populated iterable
+        if not error or none_in_error(error) or (error and not error_is_iterable(error)):
```
TimPansino, line 375: If the intent of this change is to make it possible to pass in an instance of BaseException to be able to report it then this change is overcomplicated and unfortunately looks to be introducing a bug.

Take for example this code:

```python
try:
    raise ValueError()
except ValueError as exc1:
    pass

try:
    raise TypeError()
except TypeError as exc2:
    notice_error(exc1)
```

Previously the exception instance was not allowed as an input, and would have caused an error that would crash.

With this PR's changes instead, it silently reports the wrong exception. The user is trying to pass in an exception to report, but as written this will trigger a pull from `sys.exc_info()` which will yield back the tuple for `exc2`, instead of `exc1` which was passed in.

There's also a lot of complexity in these `if` statements now that I don't care for, as well as a bunch of new function calls that add overhead. Instead, I propose we construct a tuple in the form we expect from the exception instance if and only if the error passed in is an instance of BaseException and it has a non None `__traceback__` attribute.

PR #1565, opened 2025-10-31
Bedrock Converse Streaming Support

TimPansino:
# Overview

* Add support for `BedrockRuntime.converse_stream` in `botocore` and `boto3`.

### DIFF
--- a/newrelic/hooks/external_botocore.py
+++ b/newrelic/hooks/external_botocore.py
@@ -970,11 +970,6 @@ class BedrockRecordEventMixin:
         if "messageStop" in event:
             bedrock_attrs["response.choices.finish_reason"] = (event.get("messageStop") or {}).get("stopReason", "")
 
-        # TODO: Is this also subject to the content_block_stop behavior from Langchain?
-        # If so, that would preclude us from ever capturing the messageStop event with the stopReason.
-        # if "contentBlockStop" in event:
-        #     self.record_events_on_stop_iteration(transaction)
-
 
 class EventStreamWrapper(ObjectProxy):
     def __iter__(self):


newrelic/hooks/external_botocore.py
```diff
@@ -850,35 +861,129 @@ def extract_bedrock_converse_attrs(kwargs, response, response_headers, model, sp
...
+                if not error_attributes:
+                    return
+
+                error_attributes = bedrock_error_attributes(exc, error_attributes)
+                notice_error_attributes = {
+                    "http.statusCode": error_attributes.get("http.statusCode"),
+                    "error.message": error_attributes.get("error.message"),
+                    "error.code": error_attributes.get("error.code"),
+                }
+                notice_error_attributes.update({"completion_id": str(uuid.uuid4())})
```
umaannamalai:
```suggestion
                notice_error_attributes["completion_id"] = str(uuid.uuid4())})
```
Since this is just one pair being added to the dict, would direct access be more performant?

newrelic/hooks/external_botocore.py
```diff
@@ -850,35 +861,129 @@ def extract_bedrock_converse_attrs(kwargs, response, response_headers, model, sp
...
+
+            content = ((event.get("contentBlockDelta") or {}).get("delta") or {}).get("text", "")
+            if "output_message_list" not in bedrock_attrs:
+                bedrock_attrs["output_message_list"] = [{"role": "assistant", "content": ""}]
+            bedrock_attrs["output_message_list"][0]["content"] += content
+
+        if "messageStop" in event:
+            bedrock_attrs["response.choices.finish_reason"] = (event.get("messageStop") or {}).get("stopReason", "")
+
+        # TODO: Is this also subject to the content_block_stop behavior from Langchain?
```
umaannamalai: Flagging this for discussion

newrelic/hooks/external_botocore.py
umaannamalai, line 1: Unrelated to your changes but I noticed around line 762 the following code is missing a () after `global_settings`:

`settings = transaction.settings or global_settings`

Could you add that in while you're in here?

newrelic/hooks/external_botocore.py
```diff
@@ -840,6 +850,7 @@ def _wrap_bedrock_runtime_converse(wrapped, instance, args, kwargs):
 
 def extract_bedrock_converse_attrs(kwargs, response, response_headers, model, span_id, trace_id):
     input_message_list = []
```
umaannamalai, line 852:
```suggestion
        input_message_list = []
    output_message_list = None
    # If a system message is supplied, it is under its own key in kwargs rather than with the other input messages
    if "system" in kwargs.keys():
        input_message_list.extend({"role": "system", "content": result["text"]} for result in kwargs.get("system", []))

    # kwargs["messages"] can hold multiple requests and responses to maintain conversation history
    # We grab the last message (the newest request) in the list each time, so we don't duplicate recorded data
    _input_messages = kwargs.get("messages", [])
    _input_messages = _input_messages and (_input_messages[-1] or {})
    _input_messages = _input_messages.get("content", [])
    input_message_list.extend(
        [{"role": "user", "content": result["text"]} for result in _input_messages if "text" in result]
    )
```
These changes were on the strands/ converse branch to prevent `KeyErrors` so we may want to port them into this branch.

newrelic/hooks/external_botocore.py
TimPansino: This was removed after further review of the framework code as not necessary.

PR #1571, opened 2025-11-04
Add support for BaseException instances as arguments to notice_error

TimPansino:
# Overview

* Add ability to supply an instance of BaseException as an argument to notice_error and have it be unpacked into an exception tuple before processing.
* Add tests to ensure the behavior reports the correct exception even when sys.exc_info() would return a different one.

# Related Issues

Rework of #1564


umaannamalai (approved):
Tested the change and it looks good!

PR #1580, opened 2025-11-14
 LangChain: Fix message timestamps, add default role assignment, and Bedrock support

sgoel-nr:
## Summary
This PR addresses several bugs related to LangChain integration in the New Relic Python Agent, specifically focusing on message timestamp handling and role assignments for chat completion events.

### Message Timestamp Handling
- Fixed request message timestamp preservation for LangChain chat completion messages
- Ensured proper timestamp propagation to `LlmChatCompletionMessage` events
- Corrected request timestamp handling for `LlmChatCompletionSummary` table entries

### Role Assignment
- Implemented default role assignment for input/output messages within LangChain
- Ensures consistent role metadata across chat completion events

### Bedrock Integration
- Bedrock methods are properly integrated with the LangChain instrumentation

### DIFF
--- a/newrelic/hooks/external_botocore.py
+++ b/newrelic/hooks/external_botocore.py
@@ -705,6 +705,7 @@ def wrap_bedrock_runtime_invoke_model(response_streaming=False):
         instance._nr_ft = ft
         instance._nr_response_streaming = response_streaming
         instance._nr_settings = settings
+        instance._nr_request_timestamp = request_timestamp
 
         # Add a bedrock flag to instance so we can determine when make_api_call instrumentation is hit from non-Bedrock paths and bypass it if so
         instance._nr_is_bedrock = True
@@ -834,6 +835,7 @@ def wrap_bedrock_runtime_converse(response_streaming=False):
         instance._nr_ft = ft
         instance._nr_response_streaming = response_streaming
         instance._nr_settings = settings
+        instance._nr_request_timestamp = request_timestamp
         instance._nr_is_converse = True
 
         # Add a bedrock flag to instance so we can determine when make_api_call instrumentation is hit from non-Bedrock paths and bypass it if so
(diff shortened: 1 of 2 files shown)


newrelic/hooks/mlmodel_gemini.py
```diff
@@ -221,6 +222,8 @@ def wrap_generate_content_sync(wrapped, instance, args, kwargs):
     if not settings.ai_monitoring.enabled:
         return wrapped(*args, **kwargs)
 
+    kwargs["timestamp"] = int(1000.0 * time.time())
```
umaannamalai: We will want to avoid actually inserting the timestamp attribute directly into the `kwargs` dictionary. `kwargs` contains the arguments that are actually passed to Gemini's `generate_content` function call so we cannot directly alter this with our own attribute. We may just need to handle this in a similar way to the `completion_id` where we create it in the generate content wrapper and then pass it through to `_record_generation_error/ success`

newrelic/hooks/external_botocore.py
```diff
@@ -881,6 +927,7 @@ def __iter__(self):
 class GeneratorProxy(ObjectProxy):
```
umaannamalai: The streaming/ generator proxy logic in this file has changed with the addition of Converse streaming. We will want to rebase this file to pull in the new code from the repo's main branch and store the request timestamp appropriately.

newrelic/hooks/external_botocore.py
```diff
@@ -1091,6 +1139,7 @@ def handle_chat_completion_event(transaction, bedrock_attrs):
         "response.number_of_messages": number_of_messages,
         "response.choices.finish_reason": bedrock_attrs.get("response.choices.finish_reason", None),
         "error": bedrock_attrs.get("error", None),
+        "timestamp": request_timestamp or None,
```
umaannamalai, line 1186: Does the UI rely on the accurate timestamp being present on the summary event or just the message events? I just wanted to confirm whether it is necessary to introduce this logic on the summary events (unless this is just for consistency with the corresponding message event timestamps).

sgoel-nr: Yes, UI logic is dependent on the timestamp for both summary and message event. In AI Response view, it uses the AI completion messages. In model inventor page, it a join of both completion and summary messages.

newrelic/hooks/external_botocore.py
sgoel-nr: I did the rebase and updated the logic accordingly

PR #1587, opened 2025-11-19
Initial Hybrid Agent Trace implementation

lrafeei:
This PR contains a barebones tracing implementation.  Scaffolding exists for many features, but `SpanKind.INTERNAL`/`BackgroundTasks`/`FunctionTrace` are the only implementations tested in this branch.

Future PRs include (and will likely be in this order):

- Distributed tracing/context management
- Change enabled setting name
- WSGI PR
- ASGI PR
- Scope attributes for instrumentation frameworks
- Tracer include/exclude settings
- Datastore PR
- MessageQueues PR
- gRPC PR
- Status implementation
- Events/logging
- Links
- Metrics update
- Configuration setup
- Sampler integration between NR and OTel

### DIFF
--- a/newrelic/api/opentelemetry.py
+++ b/newrelic/api/opentelemetry.py
@@ -102,6 +102,7 @@ class Span(otel_api_trace.Span):
                 self.otel_parent,
                 current_nr_trace,
             )
+            return
 
         if nr_trace_type == FunctionTrace:
             trace_kwargs = {"name": self.name, "params": self.attributes, "parent": self.nr_parent}


newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,490 @@
...
+            nr_transaction or current_transaction()
+        )  # This attribute is purely to prevent garbage collection
+        self.nr_trace = None
+        self.instrumenting_module = instrumenting_module
+
+        self.nr_parent = None
+        current_nr_trace = current_trace()
+        if (
+            not self.otel_parent
+            or (self.otel_parent and self.otel_parent.span_id == int(current_nr_trace.guid, 16))
```
hmstepanek: Is the casting necessary here?
Could the current_nr_trace come back as None and cause a crash here?

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,490 @@
...
+            #    to the newly created one.
+            # 3. Only a Sentinel Trace exists so far while still having a
+            #    remote parent. From OTel's end, this will be represented
+            #    as a `NonRecordingSpan` (and be seen as `None` at this
+            #    point). This covers cases where span is remote.
+            self.nr_parent = current_nr_trace
+        else:
+            # Not sure if there is a usecase where we could get in here
+            # but for debug purposes, we will raise an error
+            _logger.debug("Otel span and NR trace do not match nor correspond to a remote span")
```
hmstepanek: Seems like this should be a warning and not just happen in debug:
```suggestion
            _logger.warning("OpenTelemetry span (%s) and NR trace (%s) do not match nor correspond to a remote span.", self.otel_parent, current_nr_trace)
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,490 @@
...
+            _logger.debug("otel span: %s\nnewrelic trace: %s", self.otel_parent, current_nr_trace)
```
```suggestion
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,490 @@
...
+            raise ValueError("Unexpected span parent scenario encountered")
```
hmstepanek: Is this error caught somewhere?

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,490 @@
...
+
+        if nr_trace_type == FunctionTrace:
```
hmstepanek, line 107: I have a lot of questions about these. Mainly; shouldn't we be pulling all the attributes out of the OTEL attr list and mapping the here into the appropriate NR equivalents unless there isn't a spot for them in which case they are supposed to become user attributes according to the spec?

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,490 @@
...
+            trace_kwargs = {"name": self.name, "params": self.attributes, "parent": self.nr_parent}
+            self.nr_trace = nr_trace_type(**trace_kwargs)
+        elif nr_trace_type == DatastoreTrace:
+            trace_kwargs = {
+                "product": self.instrumenting_module,
+                "target": None,
```
hmstepanek, line 113: It looks like `target` maps to OTEL's `db.name`.

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+        # can also mean the following:
+        # 1. There was not a context passed in that explicitly has sampling disabled.
+        #   This flag would be found in the traceparent or traceparent and tracespan headers.
+        # 2. Transaction was not created where DT headers are accepted during __init__
+        # Therefore, we will treat a value of `None` as `True` for now.
+
+        if self.otel_parent:
+            return bool(self.otel_parent.trace_flags)
+        else:
+            return bool(self.nr_transaction and (self.nr_transaction.sampled or (self.nr_transaction._sampled is None)))
```
hmstepanek: The property just returns self._sampled so self.nr_transaction._sampled and self.nr_transaction.sampled are the same.
```suggestion
            return bool(self.nr_transaction and (self.nr_transaction.sampled or (self.nr_transaction.sampled is None)))
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+    def _is_sampled(self):
+        # Uses NR to determine if the trace is sampled
+
+        # transaction.sampled can be None, True, False.
+        # If None, this has not been computed by NR which
+        # can also mean the following:
+        # 1. There was not a context passed in that explicitly has sampling disabled.
+        #   This flag would be found in the traceparent or traceparent and tracespan headers.
+        # 2. Transaction was not created where DT headers are accepted during __init__
+        # Therefore, we will treat a value of `None` as `True` for now.
```
hmstepanek: Why treat sampled=None as sampled rather than not sampled?

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+
+        if self.otel_parent:
+            return bool(self.otel_parent.trace_flags)
+        else:
+            return bool(self.nr_transaction and (self.nr_transaction.sampled or (self.nr_transaction._sampled is None)))
+
+    def _is_remote(self):
+        # Remote span denotes if propagated from a remote parent
+        if self.otel_parent and self.otel_parent.is_remote:
```
hmstepanek: Could probably just simplify this to be:
return bool(self.otel_parent and self.otel_parent.is_remote)

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+            return True
+        return False
+
+    def get_span_context(self):
+        if not (hasattr(self, "nr_trace") and self.nr_trace):
```
hmstepanek: This is hard to follow-maybe do this instead?
```suggestion
        if not getattr(self, "nr_trace", False):
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+            return otel_api_trace.INVALID_SPAN_CONTEXT
+
+        otel_tracestate_headers = None
+
+        return otel_api_trace.SpanContext(
+            trace_id=int(self.nr_transaction._trace_id, 16),
```
hmstepanek: Use the public property method to access the trace id (that goes for all of these-I think I saw this issue in a couple other places as well with other attributes that had property methods defined).
```suggestion
            trace_id=int(self.nr_transaction.trace_id, 16),
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+
+    def set_attribute(self, key, value):
+        self.attributes[key] = value
+
+    def set_attributes(self, attributes):
+        for key, value in attributes.items():
+            self.set_attribute(key, value)
+
+    def _set_attributes_in_nr(self, otel_attributes=None):
+        if not (otel_attributes and hasattr(self, "nr_trace") and self.nr_trace):
```
hmstepanek, line 189:
```suggestion
        if not (otel_attributes and getattr(self, "nr_trace", False):
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+            return
+        for key, value in otel_attributes.items():
+            self.nr_trace.add_custom_attribute(key, value)
+
+    def add_event(self, name, attributes=None, timestamp=None):
+        # TODO: Not implemented yet.
+        # We can implement this as a log event
+        pass
```
```suggestion
        raise NotImplementedError("TODO: We can implement this as a log event.")
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+
+    def add_link(self, context=None, attributes=None):
+        # TODO: Not implemented yet.
+        pass
```
```suggestion
    raise NotImplementedError("Not implemented yet.")
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+    def update_name(self, name):
+        # Sentinel, MessageTrace, DatastoreTrace, and ExternalTrace
+        # types do not have a name attribute
+        self._name = name
+        if hasattr(self, "nr_trace") and hasattr(self.nr_trace, "name"):
+            self.nr_trace.name = self._name
+
+    def is_recording(self):
+        return self._is_sampled() and not (
+            hasattr(self, "nr_trace") and hasattr(self.nr_trace, "end_time") and self.nr_trace.end_time
```
```suggestion
            getattr(getattr(self, "nr_trace"),  None), "end_time", None)
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+        )
+
+    def set_status(self, status, description=None):
+        # TODO: not implemented yet
+        pass
```
```suggestion
       raise NotImplementedError("Not implemented yet")
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+
+    def record_exception(self, exception, attributes=None, timestamp=None, escaped=False):
+        if not hasattr(self, "nr_trace"):
+            if exception:
+                notice_error((type(exception), exception, exception.__traceback__))
```
hmstepanek: Should we be passing attributes here too?

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+            else:
+                notice_error(sys.exc_info(), attributes=attributes)
+        else:
+            self.nr_trace.notice_error((type(exception), exception, exception.__traceback__), attributes=attributes)
```
hmstepanek: What if there's no exception here?

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+    def is_recording(self):
+        return self._is_sampled() and not (
+            hasattr(self, "nr_trace") and hasattr(self.nr_trace, "end_time") and self.nr_trace.end_time
+        )
+
+    def set_status(self, status, description=None):
+        # TODO: not implemented yet
+        pass
+
+    def record_exception(self, exception, attributes=None, timestamp=None, escaped=False):
```
hmstepanek, line 217: Should we be doing something with timestamp and escaped here?

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,486 @@
...
+            else:
+                notice_error(sys.exc_info(), attributes=attributes)
+        else:
+            self.nr_trace.notice_error((type(exception), exception, exception.__traceback__), attributes=attributes)
+
+    def end(self, end_time=None, *args, **kwargs):
+        # We will ignore the end_time parameter and use NR's end_time
+
+        # Check to see if New Relic trace ever existed or,
+        # if it does, that trace has already ended
```
hmstepanek, line 229: Maybe store this value and simplify the logic rather than calling hasattr a bunch:
```
nr_trace = hasattr(self, "nr_trace", None)
if not nr_trace or nr_trace and getattr(nr_trace, "end_time", None):
    return

**Review by hmstepanek — CHANGES_REQUESTED** [review]
Just reviewed this one class so far but wanted to get you the feedback now rather than waiting until next week.

File: newrelic/api/opentelemetry.py (hunk 5 above)
Comment (lrafeei) [reply newrelic/api/opentelemetry.py:107]: That part will come in a separate PR.  For now, the attributes are added as custom attributes, but the actual attribute mapping per SpanKind will come in separate PRs.  Of course, most of the time, the attributes are not added during initialization, but rather throughout the span's operation, so I will be adding logic to add these when the span is ending.

File: newrelic/api/opentelemetry.py (hunk 4 above)
Comment (lrafeei) [reply newrelic/api/opentelemetry.py]: In this case, I cannot think of a scenario where this would happen, so I wanted to keep this here for debug purposes.

File: newrelic/api/opentelemetry.py (hunk 8 above)
Comment (lrafeei) [reply newrelic/api/opentelemetry.py]: I chose this because Opentelemetry evaluates this toward the beginning of the span cycle whereas we do this toward the end (most of the time).  `sampled=True` or `sampled=False` are explicitly known states, but when Opentelemetry requests this information and `sampled=None`, this is effectively only because NR has not gotten to it yet.  This was one of the things I was hoping to sync on re: adaptive sampler integration

File: newrelic/api/opentelemetry.py (hunk 19 above)
Comment (lrafeei) [reply newrelic/api/opentelemetry.py:217]: It does not seem like we will do anything with the escaped argument, and the timestamp argument will be used for the events/log events (which I have not implemented yet)

File: newrelic/api/opentelemetry.py (hunk 6 above)
Comment (lrafeei) [reply newrelic/api/opentelemetry.py:113]: It does!  The issue with the OTel attributes is that the `db.name` attribute is usually not set during the creation of the span, but rather later on in the span's operation.

File: newrelic/api/opentelemetry.py (hunk 8 above)
Comment (hmstepanek) [reply newrelic/api/opentelemetry.py]: Ok I have more questions but maybe lets just talk in person about this one more.

File: newrelic/api/opentelemetry.py [hunk 21]
```diff
@@ -0,0 +1,476 @@
...
+            }
+            self.nr_trace = nr_trace_type(**trace_kwargs)
+        else:
+            # TODO: Still need to implement GraphQLOperationTrace and GraphQLResolverTrace
+            trace_kwargs = {"name": self.name, "params": self.attributes, "parent": self.nr_parent}
+            self.nr_trace = nr_trace_type(**trace_kwargs)
+
+        self.nr_trace.__enter__()
+
+    def _is_sampled(self):
```
hmstepanek: Are these `_is_sampled` and `_is_remote` functions part of the otel api? I don't see them listed and if they aren't we should make these names more Pythonic (aka: `sampled`, `remote`). The prefixing of `is` comes from javascript (maybe other languages do this too but not generally Python).

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,476 @@
...
+#   "wsgi": WSGI Web Application
+#   "asgi": ASGI Web Application
+#   "other": Background Task
+#   "message": Message Queue
+# Traces:
+#   "other": Function Trace
+#   "db": Database/Datastore
+#   "message": Message Queue
+#
+INSTRUMENTING_MODULE_TYPE = {
```
hmstepanek: Checking my understanding: this is a mapping of modules to span/transaction types for modules in OTEL that don't have attributes added on span creation that would otherwise help us distinguish what type of instrumentation it is.

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,476 @@
...
+        **kwargs,
+    ):
+        # Retrieve parent span
+        parent_span_context = otel_api_trace.get_current_span(context).get_span_context()
+
+        if parent_span_context is None or not parent_span_context.is_valid:
+            parent_span_context = None
+
+        nr_trace_type = FunctionTrace
+        transaction = current_transaction()
```
hmstepanek: Wondering if it would be wise to use this transaction to make sure we don't have one that already exists and is active before we create a new one? I see one section uses the transaction and the other section doesn't-maybe I'm just having trouble following the logic here.

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,476 @@
...
+    def get_tracer(
+        self,
+        instrumenting_module_name="Default",
+        instrumenting_library_version=None,
+        schema_url=None,
+        attributes=None,
+        *args,
+        **kwargs,
+    ):
+        return Tracer(resource=self._resource, instrumentation_library=instrumenting_module_name)
```
hmstepanek: Should we be passing args and kwargs here?

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -0,0 +1,171 @@
...
+
+def wrap_set_tracer_provider(wrapped, instance, args, kwargs):
+    settings = global_settings()
+
+    if not settings.otel_bridge.enabled:
+        return wrapped(*args, **kwargs)
+
+    global _TRACER_PROVIDER
+    bound_args = bind_args(wrapped, args, kwargs)
+    tracer_provider = bound_args.get("tracer_provider", None)
```
```suggestion
    tracer_provider = bound_args.get("tracer_provider")
```

hmstepanek: Maybe move these calls into the if statement to save time?

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -0,0 +1,171 @@
...
+    # This needs to act as a singleton, like the agent instance.
+    # We should initialize the agent here as well, if there is
+    # not an instance already.
+    application = application_instance(activate=False)
+    if not application or (application and not application.active):
+        application_instance().activate()
+
+    global _TRACER_PROVIDER
+
+    hybrid_agent_tracer_provider = TracerProvider("hybrid_agent_tracer_provider")
```
hmstepanek: Again, move this into the if statement so we only create it if we need to.

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -0,0 +1,171 @@
...
+
+    # Do not allow the wrapper to continue if
+    # the Hybrid Agent setting is not enabled
+    settings = global_settings()
+
+    if not settings.otel_bridge.enabled:
+        return wrapped(*args, **kwargs)
+
+    bound_args = bind_args(wrapped, args, kwargs)
+    context_carrier = bound_args.get("context_carrier", None)
```
```suggestion
    context_carrier = bound_args.get("context_carrier")
```

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
```diff
@@ -0,0 +1,115 @@
...
+from testing_support.validators.validate_transaction_metrics import validate_transaction_metrics
+
+from newrelic.api.application import application_instance
+from newrelic.api.background_task import BackgroundTask
+from newrelic.api.external_trace import ExternalTrace
+from newrelic.api.function_trace import FunctionTrace
+from newrelic.api.time_trace import current_trace
+from newrelic.api.transaction import current_transaction
+
+tracer = otel_api_trace.get_tracer(__name__)
```
hmstepanek: Should this be a fixture with a more global context?

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+# Does not create segment without a transaction
+@validate_transaction_count(0)
+def test_Does_not_create_segment_without_a_transaction():
```
hmstepanek:
```suggestion
def test_does_not_create_segment_without_a_transaction():
```

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
```diff
@@ -0,0 +1,115 @@
...
+        assert not current_transaction()
+
+
+# Creates OpenTelemetry segment in a transaction
+@validate_transaction_metrics(name="Foo", background_task=True)
+@validate_span_events(
+    exact_intrinsics={"name": "Function/Bar", "category": "generic"}, expected_intrinsics=("parentId",)
+)
+@validate_span_events(exact_intrinsics={"name": "Function/Foo", "category": "generic", "nr.entryPoint": True})
+def test_Creates_OpenTelemetry_segment_in_a_transaction():
```
hmstepanek:
```suggestion
def test_creates_opentelemetry_segment_in_a_transaction():
```

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
```diff
@@ -0,0 +1,115 @@
...
+    application = application_instance(activate=False)
+
+    with BackgroundTask(application, name="Foo"):
+        with tracer.start_as_current_span(name="Bar", kind=otel_api_trace.SpanKind.INTERNAL):
```
hmstepanek, line 49: General question: are there cross agent tests for this? Seems like we might want some...

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
```diff
@@ -0,0 +1,115 @@
...
+# Creates New Relic span as child of OpenTelemetry span
+@validate_transaction_metrics(name="Foo", background_task=True)
+@validate_span_events(
+    exact_intrinsics={"name": "Function/Baz", "category": "generic"}, expected_intrinsics=("parentId",)
+)
+@validate_span_events(
+    exact_intrinsics={"name": "Function/Bar", "category": "generic"}, expected_intrinsics=("parentId",)
+)
+@validate_span_events(exact_intrinsics={"name": "Function/Foo", "category": "generic"})
+def test_Creates_New_Relic_span_as_child_of_OpenTelemetry_span():
```
hmstepanek:
```suggestion
def test_creates_new_relic_span_as_child_of_open_telemetry_span():
```

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
```diff
@@ -0,0 +1,115 @@
...
+    application = application_instance(activate=False)
+
+    with BackgroundTask(application, name="Foo"):
+        with tracer.start_as_current_span(name="Bar", kind=otel_api_trace.SpanKind.INTERNAL):
+            with FunctionTrace(name="Baz"):
+                # OpenTelemetry API and New Relic API report the same traceId
+                assert otel_api_trace.get_current_span().get_span_context().trace_id == int(
+                    current_transaction()._trace_id, 16
```
hmstepanek: Use the public api here and in other places:
```suggestion
                    current_transaction().trace_id, 16
```

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
```diff
@@ -0,0 +1,115 @@
...
+
+                # OpenTelemetry API and New Relic API report the same spanId
+                assert otel_api_trace.get_current_span().get_span_context().span_id == int(current_trace().guid, 16)
+
+
+# OpenTelemetry API can add custom attributes to spans
+@validate_transaction_metrics(name="Foo", background_task=True)
+@validate_span_events(exact_intrinsics={"name": "Function/Baz"}, exact_users={"spanNumber": 2})
+@validate_span_events(exact_intrinsics={"name": "Function/Bar"}, exact_users={"spanNumber": 1})
+def test_OpenTelemetry_API_can_add_custom_attributes_to_spans():
```
hmstepanek:
```suggestion
def test_opentelemetry_api_can_add_custom_attributes_to_spans():
```

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
```diff
@@ -0,0 +1,115 @@
...
+            otel_api_trace.get_current_span().set_attribute("spanNumber", 1)
+
+
+# OpenTelemetry API can record errors
+@validate_transaction_metrics(name="Foo", background_task=True)
+@validate_error_event_attributes(
+    exact_attrs={"agent": {}, "intrinsic": {"error.message": "Test exception message"}, "user": {}}
+)
+@validate_span_events(exact_intrinsics={"name": "Function/Bar"})
+def test_OpenTelemetry_API_can_record_errors():
```
hmstepanek:
```suggestion
def test_opentelemetry_api_can_record_errors():
```

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
```diff
@@ -0,0 +1,115 @@
...
+            with FunctionTrace(name="Baz"):
+                otel_api_trace.get_current_span().set_attribute("spanNumber", 2)
+
+            otel_api_trace.get_current_span().set_attribute("spanNumber", 1)
+
+
+# OpenTelemetry API can record errors
+@validate_transaction_metrics(name="Foo", background_task=True)
+@validate_error_event_attributes(
+    exact_attrs={"agent": {}, "intrinsic": {"error.message": "Test exception message"}, "user": {}}
```
hmstepanek, line 101: I think there should be an error.class attribute on here too right?

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
```diff
@@ -0,0 +1,115 @@
...
+@validate_span_events(exact_intrinsics={"name": "Function/Bar"})
+def test_OpenTelemetry_API_can_record_errors():
+    application = application_instance(activate=False)
+
+    with BackgroundTask(application, name="Foo"):
+        with tracer.start_as_current_span(name="Bar", kind=otel_api_trace.SpanKind.INTERNAL):
+            try:
+                raise Exception("Test exception message")
+            except Exception as e:
+                otel_api_trace.get_current_span().record_exception(e)
```
hmstepanek, line 112: Do you have a test anywhere for making sure we don't do anything if hybrid agent is disabled?

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,476 @@
...
+#   "asgi": ASGI Web Application
+#   "other": Background Task
+#   "message": Message Queue
+# Traces:
+#   "other": Function Trace
+#   "db": Database/Datastore
+#   "message": Message Queue
+#
+INSTRUMENTING_MODULE_TYPE = {
+    "Redis": "db",
```
hmstepanek: Maybe change db to ds (datastore).

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
lrafeei: I was going to include that in the PR where I change the setting names, but I can do that in this one too (since it's literally just one test)

newrelic/api/opentelemetry.py
lrafeei: Exactly.  I am hoping to eventually do away with this dictionary once I complete the configuration setup portion of this.  Sadly, that will be a little bit down the road, so this mapping placeholder will fulfill that purpose.

lrafeei: Actually, I realized that with some logic rewriting, I don't need to have that at all.  REMOVED.

newrelic/api/opentelemetry.py
lrafeei: The agent creates a trace (in the form of a Sentinel) if a transaction already exists and will not create a new transaction, so I did not create separate logic for that part of it

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
lrafeei: Yes, and the tests in this file are directly from those.  I have not finished the script to generate these on the fly for if/when they change, so I left the original JSON out of this PR.  I'll just put it into the next PR anyway to make it clear where these came from

tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
lrafeei: For these cross agent tests, I just used what was in the JSON files, but once I work more on the status functionality, I will have more tests for errors as well.

newrelic/api/opentelemetry.py
hmstepanek: In that case maybe do one of those Please report this error to newrelic log statements and don't crash.

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,438 @@
...
+            #    to the newly created one.
+            # 3. Only a Sentinel Trace exists so far while still having a
+            #    remote parent. From OTel's end, this will be represented
+            #    as a `NonRecordingSpan` (and be seen as `None` at this
+            #    point). This covers cases where span is remote.
+            self.nr_parent = current_nr_trace
+        else:
+            # Not sure if there is a usecase where we could get in here
+            # but for debug purposes, we will log this with details
+            _logger.warning(
```
```suggestion
            _logger.error(
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,438 @@
...
+                "OpenTelemetry span (%s) and NR trace (%s) do not match nor correspond to a remote span.",
```
```suggestion
                "OpenTelemetry span (%s) and NR trace (%s) do not match nor correspond to a remote span. Open Telemetry span will not be reported to New Relic. Please report this problem to New Relic.",
```

newrelic/api/opentelemetry.py
```diff
@@ -0,0 +1,438 @@
...
+                self.otel_parent,
+                current_nr_trace,
+            )
```
hmstepanek, line 104:
```suggestion
            )
            return
```

PR #1589, opened 2025-11-21
Hybrid agent context management

lrafeei:
This PR contains a barebones implementation for Distributed Tracing/context propagation using OTel and NR APIs

### DIFF
--- a/newrelic/api/opentelemetry.py
+++ b/newrelic/api/opentelemetry.py
@@ -231,12 +231,12 @@ class Span(otel_api_trace.Span):
             self.nr_trace.name = self._name
 
     def is_recording(self):
-        # TODO: Refine this logic further if needed.
-        # If the trace has an end time set then it is done recording. Otherwise, if it does not have an end time set and the transaction's priority has not been set yet or it is set to something other than 0 then it is also still recording.
+        # If the trace has an end time set then it is done recording. Otherwise,
+        # if it does not have an end time set and the transaction's priority
+        # has not been set yet or it is set to something other than 0 then it
+        # is also still recording.
         if getattr(self.nr_trace, "end_time", None):
             return False
-        if not getattr(self.nr_transaction, "priority", None):
-            self.nr_transaction._make_sampling_decision()
 
         return getattr(self.nr_transaction, "priority", 1) > 0
 


newrelic/api/opentelemetry.py
```diff
@@ -28,11 +32,137 @@
...
+        return nr_tracestate
+
+    def extract(self, carrier, context=None, getter=None):
+        # We need to make sure that the carrier goes out
+        # in OTel format.  However, we want to convert this to
+        # NR to use the `accept_distributed_trace_headers` API
+        transaction = current_transaction()
+        tracestate_key = None
+        tracestate_headers = None
+        for key in self.LIST_OF_TRACESTATE_KEYS:
```
```suggestion
        for tracestate_key in self.LIST_OF_TRACESTATE_KEYS:
```

newrelic/api/opentelemetry.py
```diff
@@ -28,11 +32,137 @@
...
+            if key in carrier:
+                tracestate_key = key
+                tracestate_headers = carrier[tracestate_key]
+                break
```
```suggestion
                break
        else:
            tracestate_key = None 
```

newrelic/api/opentelemetry.py
```diff
@@ -28,11 +32,137 @@
...
+        # If we are passing into New Relic, traceparent and/or tracestate's keys also need to be NR compatible.
+        if tracestate_headers:
+            # Check to see if in NR or OTel format
+            if "@nr=" in tracestate_headers:
+                # NR format
+                # Reformatting DT keys in case they are in the HTTP_* format:
+                nr_headers = carrier.copy()
```
hmstepanek: Instead of all these loops with if statements and stuff, I think it would be simpler/more efficient to do something like this in both here and the else clause:
```
HEADER_NAME_TO_HTTP_MAP = dict((LIST_OF_TRACEPARENT_KEYS, LIST_OF_TRACESTATE_KEYS, ("newrelic", None))

nr_headers = {lowercase_name: carrier.get(lowercase_name) or carrier.get(lowercase_name, "") for lowercase_name, http_name in header_name_to_http_map}
transaction.accept_distributed_trace_headers(nr_headers)

otel_tracestate = self._convert_nr_to_otel(tracestate_headers)
carrier[tracestate_key] = otel_tracestate
```

hmstepanek (commented):
Just gonna post this anyway if there's anything that is still relevant after the changes we discussed.

newrelic/api/transaction.py
```diff
@@ -1438,6 +1438,18 @@ def accept_distributed_trace_headers(self, headers, transport_type="HTTP"):
                             data.update(tracestate_data)
                         else:
                             self._record_supportability("Supportability/TraceContext/TraceState/InvalidNrEntry")
+                    elif not payload and (tracestate == self.tracestate):
+                        self._record_supportability("Supportability/TraceContext/TraceState/NoNrEntry")
+                        self._record_supportability("Supportability/TraceContext/TraceState/OtelEntry")
+                        try:
+                            vendors["sa"] = True if vendors.get("sa").lower() == "true" else False
```
lrafeei: We might not even need this since the sampled flag already gets picked up by the traceparent.


```suggestion
                            vendors.pop("sa", None)
```

newrelic/api/opentelemetry.py
```diff
@@ -26,29 +30,69 @@
...
+    def inject(self, carrier, context=None, setter=None):
+        transaction = current_transaction()
+        # Only insert headers if we have not done so already this transaction
+        # Distributed Trace State will have the following states:
+        #   0 if not set
+        #   1 if already accepted
+        #   2 if inserted but not accepted
+
+        if transaction and not transaction._distributed_trace_state:
+            if isinstance(carrier, dict):
```
```suggestion
            nr_headers = transaction._generate_distributed_trace_headers()
            for name, value in nr_headers:
                setter.set(carrier, name, value)
```

newrelic/api/opentelemetry.py
```diff
@@ -26,29 +30,69 @@
...
 
+class NRTraceContextPropagator(TraceContextTextMapPropagator):
+    LIST_OF_TRACEPARENT_KEYS = ("traceparent", "HTTP_TRACEPARENT")
+    LIST_OF_TRACESTATE_KEYS = ("tracestate", "HTTP_TRACESTATE")
+    HEADER_KEY_MAPPING = dict((LIST_OF_TRACEPARENT_KEYS, LIST_OF_TRACESTATE_KEYS, ("newrelic", "HTTP_NEWRELIC")))
+
+    def extract(self, carrier, context=None, getter=None):
+        # If we are passing into New Relic, traceparent 
+        # and/or tracestate's keys also need to be NR compatible.
+        nr_headers = {lowercase_name: carrier.get(lowercase_name, carrier.get(http_name, "")) for lowercase_name, http_name in self.HEADER_KEY_MAPPING.items()}
```
```suggestion
        nr_headers = [(lowercase_name, getter.get(carrier, lowercase_name, "") for lowercase_name, http_name in self.HEADER_KEY_MAPPING.items()]
```

newrelic/api/opentelemetry.py
```diff
@@ -26,29 +30,69 @@
...
+        return super().extract(carrier=carrier, context=context, getter=getter)
+    
+
+    def inject(self, carrier, context=None, setter=None):
+        transaction = current_transaction()
+        # Only insert headers if we have not done so already this transaction
+        # Distributed Trace State will have the following states:
+        #   0 if not set
+        #   1 if already accepted
+        #   2 if inserted but not accepted
```
```suggestion
        #   2 if inserted but not accepted -- Do not insert the headers twice.
```

newrelic/api/opentelemetry.py
```diff
@@ -26,29 +30,69 @@
...
+        
+        return super().extract(carrier=carrier, context=context, getter=getter)
+    
+
+    def inject(self, carrier, context=None, setter=None):
+        transaction = current_transaction()
+        # Only insert headers if we have not done so already this transaction
+        # Distributed Trace State will have the following states:
+        #   0 if not set
+        #   1 if already accepted
```
```suggestion
        #   1 if already accepted - We have read inbound headers and are ready to insert outbound headers.
```

newrelic/api/opentelemetry.py
```diff
@@ -26,29 +30,69 @@
...
+        accept_distributed_trace_headers(nr_headers)
+        
+        return super().extract(carrier=carrier, context=context, getter=getter)
+    
+
+    def inject(self, carrier, context=None, setter=None):
+        transaction = current_transaction()
+        # Only insert headers if we have not done so already this transaction
+        # Distributed Trace State will have the following states:
+        #   0 if not set
```
```suggestion
        #   0 if not set - Inbound headers have not been read or there were no headers to read so we insert outbound headers anyway.
```

newrelic/api/opentelemetry.py
```diff
@@ -26,29 +30,69 @@
...
+        #   1 if already accepted
+        #   2 if inserted but not accepted
+
+        if transaction and not transaction._distributed_trace_state:
```
```suggestion
        if transaction._distributed_trace_state < 2:
```

newrelic/api/opentelemetry.py
```diff
@@ -26,29 +30,69 @@
...
+                insert_distributed_trace_headers(carrier)
+            else:
+                raise TypeError("Unsupported carrier type")
+            
+            return super().inject(carrier=carrier, context=context, setter=setter)
+        
+        elif not transaction:
+            return super().inject(carrier=carrier, context=context, setter=setter)
+        
+        else:
```
```suggestion
```

newrelic/api/opentelemetry.py
```diff
@@ -26,29 +30,69 @@
...
+                nr_headers = list(carrier.items())
+                insert_distributed_trace_headers(nr_headers)
+            elif isinstance(carrier, list):
+                insert_distributed_trace_headers(carrier)
+            else:
+                raise TypeError("Unsupported carrier type")
+            
+            return super().inject(carrier=carrier, context=context, setter=setter)
+        
+        elif not transaction:
```
hmstepanek: Move this up before the if statement as an early exit condition like we normally have.

newrelic/api/opentelemetry.py
```diff
@@ -26,29 +30,77 @@
...
+class NRTraceContextPropagator(TraceContextTextMapPropagator):
+    HEADER_KEYS = ("traceparent", "tracestate", "newrelic")
+
+    def extract(self, carrier, context=None, getter=None):
+        transaction = current_transaction()
+        # If we are passing into New Relic, traceparent 
+        # and/or tracestate's keys also need to be NR compatible.
+        
+        if transaction:
+            nr_headers = {header_key: ("" if not getter.get(carrier, header_key) else getter.get(carrier, header_key)[0]) for header_key in self.HEADER_KEYS}
```
hmstepanek: Can we have a comment here explaining why we grab header value [0]?

Thought this might be simpler (comprehensions support having an if clause):
```suggestion
            nr_headers = {header_key: getter.get(carrier, header_key)[0] for header_key in self.HEADER_KEYS if getter.get(carrier, header_key)}
```

newrelic/api/opentelemetry.py
```diff
@@ -26,29 +30,77 @@
...
+        #       outbound headers.
+
+        if not transaction:
+            return super().inject(carrier=carrier, context=context, setter=setter)
+        
+        if transaction._distributed_trace_state < 2:
+            nr_headers = []
+            transaction.insert_distributed_trace_headers(nr_headers)
+            for key, value in nr_headers:
+                setter.set(carrier, key, value)
```
hmstepanek, line 86: This looks good now-so much simpler!

A couple additional thoughts/considerations here:

What if the carrier already has that header in it? I think it makes sense to override the parent header but it might make more sense to add to/combine the tracestate header value? I'm not sure how we would add to it though given that we don't have a getter function so this is prob fine.

[This](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/propagation/tracecontext.py#L104) overrides the parent header value that we generate - I assume this is what is called when we super - is the override always guaranteed to match the value we generate?

newrelic/api/opentelemetry.py
```diff
@@ -26,29 +30,77 @@
...
+            return super().inject(carrier=carrier, context=context, setter=setter)
+        
+        # If distributed_trace_state == 2 or 3, do not inject headers.
+
+
+# Context and Context Propagator Setup
+otel_context_propagator = CompositePropagator(
+    propagators=[
+        NRTraceContextPropagator(),
+        W3CBaggagePropagator(),
```
hmstepanek: Given my comment above-I wonder if we need this one since we already propagate the W3C headers?

newrelic/api/opentelemetry.py
```diff
@@ -141,24 +193,20 @@ def __init__(
...
-        # can also mean the following:
-        # 1. There was not a context passed in that explicitly has sampling disabled.
-        #   This flag would be found in the traceparent or traceparent and tracespan headers.
-        # 2. Transaction was not created where DT headers are accepted during __init__
-        #   Therefore, we will treat a value of `None` as `True` for now.
-        #
-        # The primary reason for this behavior is because Otel expects to
-        # only be able to record information like events and attributes
-        # when `is_recording()` == `True`
+        # TODO: Refine this logic to respect the parent
```
hmstepanek: I think you can get rid of this comment since we already respect the remote parent flags now based on what I've already implemented.

newrelic/api/opentelemetry.py
```diff
@@ -141,24 +193,20 @@ def __init__(
...
+        # trace/span, even if the parent is from OTel.
+        # priority: except for always_on or always off, 
+        # sampled flag from parent if parent exists takes
+        # precedence over whatever sampling algorithm is
+        # enabled.
 
         if self.otel_parent:
```
hmstepanek: As I understand it from Chris - we need to always return sampled=True in order to fake OTEL into thinking the span is being sampled and collect data on it. Then we use our internal sampling decision later to decide whether we are actually going to sample it (aka add it to our reservoir).

newrelic/api/opentelemetry.py
```diff
@@ -208,7 +254,13 @@ def update_name(self, name):
             self.nr_trace.name = self._name
 
     def is_recording(self):
-        return self._sampled() and not (getattr(self.nr_trace, None), "end_time", None)
+        # TODO: Refine this logic along with `_sampled()`
+        if getattr(self.nr_trace, "end_time", None):
```
hmstepanek, line 238: Is there a case where a span has been created but recording hasn't started yet that we need to handle here?

newrelic/api/opentelemetry.py
```diff
@@ -208,7 +254,13 @@ def update_name(self, name):
...
+            return False
+        if not getattr(self.nr_transaction, "priority", None):
```
hmstepanek: I'm confused why we are looking at the priority here? Is this just being used to determine whether the transaction has ended or not?

newrelic/api/opentelemetry.py
```diff
@@ -296,6 +360,10 @@ def start_span(
                 port = self.attributes.get("net.host.port")
                 request_method = self.attributes.get("http.method")
                 request_path = self.attributes.get("http.route")
+
+                update_sampled_flag = False if headers else True
+                headers = headers if headers else _headers
```
```suggestion
                headers = headers or {}
```

newrelic/api/opentelemetry.py
```diff
@@ -338,6 +415,9 @@ def start_span(
                     request_method = self.attributes.get("http.method")
                     request_path = self.attributes.get("http.route")
 
+                    update_GUID_flag = False if headers else True
```
hmstepanek: Do we really need this and the one above? Can't we just say if headers down below?

newrelic/api/opentelemetry.py
```diff
@@ -338,6 +415,9 @@ def start_span(
...
+                    headers = headers if headers else _headers
```
```suggestion
                    headers = headers or headers
```

newrelic/api/opentelemetry.py
```diff
@@ -296,6 +360,10 @@ def start_span(
                 port = self.attributes.get("net.host.port")
                 request_method = self.attributes.get("http.method")
                 request_path = self.attributes.get("http.route")
+
+                update_sampled_flag = False if headers else True
```
hmstepanek: Wondering if we need this one?

newrelic/api/opentelemetry.py
```diff
@@ -286,7 +342,15 @@ def start_span(
         if parent_span_context is None or not parent_span_context.is_valid:
             parent_span_context = None
 
+        # If parent_span_context exists, we can create traceparent
+        # and tracestate headers
+        _headers = {}
```
```suggestion
```

newrelic/api/opentelemetry.py
```diff
@@ -286,7 +342,15 @@ def start_span(
...
+        if parent_span_context and self.nr_application.settings.distributed_tracing.enabled:
```
hmstepanek, line 324: Do we need to default these to None above this if statement so we don't run into a case where the variable isn't defined?

newrelic/api/opentelemetry.py
```diff
@@ -348,6 +428,14 @@ def start_span(
                         request_path=request_path,
                         headers=headers,
                     )
+
+                    # If headers do not contain the traceparent/tracestate
+                    # the transaction GUID needs to be updated to that of
+                    # the parent span.
+                    if update_GUID_flag and parent_span_context:
+                        guid = parent_span_trace_id >> 64
+                        transaction.guid = f"{guid:x}"
```
hmstepanek: The transaction guid consists of part of trace id but the transaction also contains another attribute with the full trace id-it seems that both need to be updated here. I think what we might need to do here is pass the trace_id into the `Transaction.__init__` (aka update the `__init__` to optionally accept a value for `trace_id`).

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -13,25 +13,74 @@
...
 from newrelic.common.object_wrapper import wrap_function_wrapper
 from newrelic.common.signature import bind_args
 from newrelic.core.config import global_settings
 
 _logger = logging.getLogger(__name__)
 _TRACER_PROVIDER = None
 
+# Enable OpenTelemetry Bridge to capture HTTP 
+# request/response headers as span attributes:
+os.environ["OTEL_INSTRUMENTATION_HTTP_CAPTURE_HEADERS_SERVER_REQUEST"] = ".*"
```
hmstepanek, line 30: Is it safe to do this here? What if the customer is running otel but doesn't have our hybrid agent enabled?

newrelic/api/opentelemetry.py
lrafeei: I'm having a flashback to why I had omitted the `super()` call in for this specific condition.  It will not hurt anything (the results for traceparent and tracestate are the same or that the results for traceparent are the same and that the OTel span does not have a tracestate whereas NR passed in its own tracestate, so that result does not get overridden).  It's just redundant.

I'll put in a comment here so that I don't forget why I excluded it (again).

newrelic/api/opentelemetry.py
lrafeei: Yes--In most of the OTel library instrumentations, `span.is_recording()` is called almost immediately after creation.  [This](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-wsgi/src/opentelemetry/instrumentation/wsgi/__init__.py#L695) is an example of that happening.

newrelic/api/opentelemetry.py
lrafeei: Oh, the Baggage Propagator?  That's a separate thing that we should still keep (that is basically an OTel exclusive feature)

newrelic/api/opentelemetry.py
lrafeei: You're right, in OTel, this is just used to determine if the span has ended or not.  We end up not incorporating an `end_time` attribute for spans (and instead allowing the NR trace and transactions to deal with that).  This call is usually made almost immediately after starting a span (it is also usually called at the end of the span).  If it is at the beginning of the start of the span, there will not be an `nr_trace` attribute yet.  There will be an `nr_transaction` attribute, though.

For the second part of the logic here, my thought process here is that if an OTel header is being passed into the Hybrid agent, it may have a sampled flag but not a priority flag.  So, in order to respect the sampling decision configured in the Hybrid agent, if there is no priority flag, it should run `_make_sampling_decision()`

OTel has a concept of recording but not sampling (which still allows for context propagation), so even if the transaction is not sampled, if the priority is greater than 0, it is still a GO for recording.

newrelic/api/opentelemetry.py
lrafeei: Ahh, you were talking about the line below that one.  Derp.  Yep, we do!  Done

newrelic/hooks/hybridagent_opentelemetry.py
lrafeei: This logic is a placeholder until the configuration set up has been integrated

newrelic/api/opentelemetry.py
lrafeei: That's actually a fair point.  Either way, it's going to get there...

newrelic/api/opentelemetry.py
lrafeei: We do not--this was the result of some residual logic I had where the `_headers` variable above created a traceparent header from the parent span's span_context.  Since then, that has become unnecessary, so this goes away

newrelic/api/opentelemetry.py
lrafeei: Ooh, good callout.  I can also just check to see if the `transaction.guid == transaction.trace_id[:16]` and if it is, update the `transaction._trace_id`.  If not, this is part of a transaction trace.

newrelic/api/opentelemetry.py
lrafeei: Removed all together!

newrelic/api/opentelemetry.py
lrafeei: Removed all together!

newrelic/api/opentelemetry.py
lrafeei: Not anymore--removed!

newrelic/api/opentelemetry.py
lrafeei: I may have also misunderstood earlier.

I want to make sure that the default propagator includes the NRTraceContextPropagator.  I could just have set the global textmap to that custom NR propagator directly, but I wanted to keep in the scaffolding to use the `CompositePropagator` instead.  With the `CompositePropagator`, additional propagators can be added to the list while behaving like one propagator.  The idea is that with the configuration implementation, the ability to customize the global propagators can be defined there.

Baggage integration is also on the horizon.  Right now, it does not hurt anything for it to be in here, since we are neither sending data nor attempting to do anything with the data.

newrelic/api/opentelemetry.py
lrafeei: To clarify, it seems pretty common for it to call `is_recording()` immediately after the creation of the span and right before the end of the span in the framework instrumentation.

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -46,17 +95,17 @@ def wrap_set_tracer_provider(wrapped, instance, args, kwargs):
...
 def wrap_get_tracer_provider(wrapped, instance, args, kwargs):
-    settings = global_settings()
-
-    if not settings.otel_bridge.enabled:
-        return wrapped(*args, **kwargs)
-
     # This needs to act as a singleton, like the agent instance.
     # We should initialize the agent here as well, if there is
     # not an instance already.
     application = application_instance(activate=False)
```
hmstepanek: Change this to grab the application with activate=True and then activate if it's not active.

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -112,14 +155,26 @@ def set_attributes(self, attributes):
...
 
-    otel_tracestate_headers = None
-
+    if transaction.settings.distributed_tracing.enabled:
+        if not isinstance(span, otel_api_trace.NonRecordingSpan):
+            # Use the Otel trace and span ids if current span
+            # is not a NonRecordingSpan.  Otherwise, we need to 
+            # override the current span with the transaction
+            # and trace guids.
+            transaction._trace_id = f'{span.get_span_context().trace_id:x}'
```
hmstepanek: I don't think we need to update the transaction id here-we just need to update the trace id to match-the transaction id can be whatever random value.

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -112,14 +155,26 @@ def set_attributes(self, attributes):
...
                 add_custom_span_attribute(key, value)
 
-    otel_tracestate_headers = None
-
+    if transaction.settings.distributed_tracing.enabled:
+        if not isinstance(span, otel_api_trace.NonRecordingSpan):
+            # Use the Otel trace and span ids if current span
+            # is not a NonRecordingSpan.  Otherwise, we need to 
+            # override the current span with the transaction
+            # and trace guids.
```
hmstepanek, line 166: Probably be a good idea to cash this to save on compute time.
```suggestion
            # and trace guids.
            trace_id = span.get_span_context().trace_id
```

newrelic/api/opentelemetry.py
```diff
@@ -208,7 +233,13 @@ def update_name(self, name):
...
 
     def is_recording(self):
-        return self._sampled() and not (getattr(self.nr_trace, None), "end_time", None)
+        # TODO: Refine this logic further if needed.
+        if getattr(self.nr_trace, "end_time", None):
+            return False
+        if not getattr(self.nr_transaction, "priority", None):
+            self.nr_transaction._make_sampling_decision()
+
+        return self.nr_transaction.priority > 0
```
```suggestion
        return getattr(self.nr_transaction, "priority", 1) > 0
```

newrelic/api/opentelemetry.py
```diff
@@ -208,7 +233,13 @@ def update_name(self, name):
             self.nr_trace.name = self._name
 
     def is_recording(self):
-        return self._sampled() and not (getattr(self.nr_trace, None), "end_time", None)
+        # TODO: Refine this logic further if needed.
+        if getattr(self.nr_trace, "end_time", None):
+            return False
+        if not getattr(self.nr_transaction, "priority", None):
+            self.nr_transaction._make_sampling_decision()
```
```suggestion
```

newrelic/api/opentelemetry.py
```diff
@@ -208,7 +233,13 @@ def update_name(self, name):
             self.nr_trace.name = self._name
 
     def is_recording(self):
-        return self._sampled() and not (getattr(self.nr_trace, None), "end_time", None)
+        # TODO: Refine this logic further if needed.
+        if getattr(self.nr_trace, "end_time", None):
+            return False
+        if not getattr(self.nr_transaction, "priority", None):
```
```suggestion
```

newrelic/api/opentelemetry.py
```diff
@@ -208,7 +233,13 @@ def update_name(self, name):
             self.nr_trace.name = self._name
 
     def is_recording(self):
-        return self._sampled() and not (getattr(self.nr_trace, None), "end_time", None)
+        # TODO: Refine this logic further if needed.
+        if getattr(self.nr_trace, "end_time", None):
```
hmstepanek, line 238:
```suggestion
        # If the trace has an end time set then it is done recording. Otherwise, if it does not have an end time set and the transaction's priority has not been set yet or it is set to something other than 0 then it is also still recording.
        if getattr(self.nr_trace, "end_time", None):
```

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -112,14 +155,26 @@ def set_attributes(self, attributes):
...
+            transaction._trace_id = f'{span.get_span_context().trace_id:x}'
+            guid = span.get_span_context().trace_id >> 64
+            transaction.guid = f'{guid:x}'
+        elif not transaction.priority:
```
```suggestion
```

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -112,14 +155,26 @@ def set_attributes(self, attributes):
...
+            # This is a NonRecordingSpan and the transaction
+            # sample and priority have not been computed yet.
+            transaction._make_sampling_decision()
```
```suggestion
```

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -112,14 +155,26 @@ def set_attributes(self, attributes):
...
+        
     span_context = otel_api_trace.SpanContext(
         trace_id=int(transaction.trace_id, 16),
         span_id=int(trace.guid, 16),
         is_remote=span.get_span_context().is_remote,
-        trace_flags=otel_api_trace.TraceFlags(span.get_span_context().trace_flags),
-        trace_state=otel_api_trace.TraceState(otel_tracestate_headers),
+        trace_flags=otel_api_trace.TraceFlags(0x01 if transaction.sampled else 0x00),
```
hmstepanek: Just hardcode this to be always sampled.

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -144,12 +199,10 @@ def wrap_start_internal_or_server_span(wrapped, instance, args, kwargs):
     if context_carrier:
         if ("HTTP_HOST" in context_carrier) or ("http_version" in context_carrier):
             # This is an HTTP request (WSGI, ASGI, or otherwise)
-            nr_environ = context_carrier.copy()
-            attributes["nr.http.headers"] = nr_environ
+            attributes["nr.http.headers"] = context_carrier
```
hmstepanek, line 199: Just wondering why we had/got rid of the copy here?

tests/hybridagent_opentelemetry/test_context_propagation.py
```diff
@@ -0,0 +1,126 @@
...
+
+_override_settings = {
+    "trusted_account_key": "1",
+    "distributed_tracing.enabled": True,
+    "span_events.enabled": True,
+}
+
+@pytest.mark.parametrize("telemetry", ["newrelic", "otel"])
+@pytest.mark.parametrize("propagation", [accept_distributed_trace_headers, PROPAGATOR.extract])
+def test_distributed_trace_header_compatibility_full_granularity(
```
hmstepanek: Looks like there's some megalinter complaints here.

newrelic/hooks/hybridagent_opentelemetry.py
lrafeei: I realized I was not actually manipulating the header data anymore, so I could just pass in the contents of the carrier directly as an attribute

PR #1594, opened 2025-11-25
Add grouping categories to import hooks

lrafeei:
This PR adds a new feature to the agent: the ability to disable instrumentation for an entire group/package/module.

Use case: If instrumentation for Django is not desired, we would have to add 22 sections into the *.ini file, disabling each individual hook

```
[import-hook:django.core.handlers.base]
enabled = false
[import-hook:django.django.core.handlers.asgi]
enabled = false
[import-hook:django.core.handlers.wsgi]
enabled = false
```

et cetera.

With this feature, we can simply use 
```
[import-hook:django]
enabled = false
```

And it will disable instrumentation for any Django hook.

Notes:
- The grouping disable settings will take precedence over the individual import hook settings.
- `asyncio` and the logging libraries do not have a separate grouping.

### DIFF
--- a/.github/workflows/addlicense.yml
+++ b/.github/workflows/addlicense.yml
@@ -39,7 +39,7 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # 5.0.1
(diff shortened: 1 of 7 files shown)


lrafeei:
Closing this PR.  The OTel configurations have to be separate from this for several reasons (the biggest of which involves context propagation and inclusion/exclusion of transaction creation).

TL;DR: back to the drawing board.

PR #1597, opened 2025-12-03
Add trace id ratio based sampler

hmstepanek:
# Overview
* Add trace id ratio based sampler.
* Add support for partial granularity ratio = full gran ratio + partial gran ratio (when both set to ratio).

### DIFF
--- a/newrelic/api/opentelemetry.py
+++ b/newrelic/api/opentelemetry.py
@@ -272,7 +272,6 @@ class Tracer(otel_api_trace.Tracer):
         *args,
         **kwargs,
     ):
-
         nr_trace_type = FunctionTrace
         transaction = current_transaction()
         self.nr_application = application_instance()
--- a/newrelic/core/config.py
+++ b/newrelic/core/config.py
@@ -701,6 +701,7 @@ def _environ_as_sampler(name, default):
         return default
     return val
 
+
 def _environ_as_int(name, default=0):
     val = os.environ.get(name, default)
     try:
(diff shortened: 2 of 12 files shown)


tests/agent_unittests/test_distributed_tracing_settings.py
```diff
@@ -144,11 +192,47 @@ def test_distributed_trace_setings(ini, env, expected_format, global_settings):
...
+            INI_FILE_EMPTY,
+            {
+                "NEW_RELIC_ENABLED": "true",
+                "NEW_RELIC_DISTRIBUTED_TRACING_SAMPLER_FULL_GRANULARITY_ROOT_TRACE_ID_RATIO_BASED_RATIO": "5",
+                "NEW_RELIC_DISTRIBUTED_TRACING_SAMPLER_FULL_GRANULARITY_REMOTE_PARENT_SAMPLED_TRACE_ID_RATIO_BASED_RATIO": "10",
+                "NEW_RELIC_DISTRIBUTED_TRACING_SAMPLER_FULL_GRANULARITY_REMOTE_PARENT_NOT_SAMPLED_TRACE_ID_RATIO_BASED_RATIO": "0",
+            },
+            ("default", "default", "default", None, None, None),
+        ),
+        (  # Ignores ratio sampler when ratio is not defined.
```
hmstepanek:
```suggestion
        (  # Ignores ratio sampler when invalid ratio path is provided.
```

newrelic/api/opentelemetry.py
```diff
@@ -436,4 +436,4 @@ def get_tracer(
         *args,
         **kwargs,
     ):
-        return Tracer(resource=self._resource, instrumentation_library=instrumenting_module_name, *args, **kwargs)
\ No newline at end of file
+        return Tracer(resource=self._resource, instrumentation_library=instrumenting_module_name, **kwargs, *args)
```
lrafeei: Not sure if this was some regex gone wrong, but this should probably be changed back

```suggestion
        return Tracer(resource=self._resource, instrumentation_library=instrumenting_module_name, *args, **kwargs)
```

newrelic/core/config.py
```diff
@@ -649,6 +694,13 @@ class OtelBridgeSettings(Settings):
 _settings.audit_log_file = os.environ.get("NEW_RELIC_AUDIT_LOG", None)
 
 
+def _environ_as_sampler(name, default):
+    val = os.environ.get(name, default)
+    # The trace_id_ratio_based value can only bet set by setting the ratio"
```
lrafeei: Super nitpicky, but


```suggestion
    # The trace_id_ratio_based value can only be set by setting the ratio
```

hmstepanek: Wow this was all sorts of miss-typed. 😳

newrelic/api/opentelemetry.py
hmstepanek: Megalinter was complaining but my fix was incorrect. I've adjusted it so hopefully it makes more sense. The `*args` needs to be moved before `resource=...`.

PR #1598, opened 2025-12-03
Prioritize full granularity over partial granularity

hmstepanek:
# Overview
Prioritize full granularity over partial granularity:

* priority +=2 when full granularity and sampled=true
* priority +=1 when partial granularity and sampled=true
* always_on sampler for full granularity sets priority=3 instead of 2
* priority=random and sampled=false when dt disabled or dt enabled but partial and full disabled
* when only sampled flag is sent in inbound tracestate header, the priority needs to be set and incremented appropriately.

### DIFF
--- a/tests/agent_features/test_distributed_tracing.py
+++ b/tests/agent_features/test_distributed_tracing.py
@@ -1446,7 +1446,7 @@ def test_distributed_trace_enabled_settings_set_correct_sampled_priority(
     _test()
 
 
-def test_distributed_trace_priority_set_when_only_sampled_set_in_headers(monkeypatch):
+def test_distributed_trace_priority_set_when_only_sampled_set_in_tracestate_header(monkeypatch):
     monkeypatch.setattr(random, "random", lambda *args, **kwargs: 0.123)
 
     @override_application_settings(_override_settings)


tests/agent_features/test_distributed_tracing.py
```diff
@@ -878,8 +880,7 @@ def _test():
     "full_granularity_enabled,full_granularity_remote_parent_sampled_setting,partial_granularity_enabled,partial_granularity_remote_parent_sampled_setting,expected_sampled,expected_priority,expected_adaptive_sampling_algo_called",
     (
         (True, "always_off", True, "adaptive", None, None, True),  # Uses adaptive sampling algo.
-        (True, "always_on", True, "adaptive", True, 2, False),  # Uses adaptive sampling algo.
-        (False, "always_on", False, "adaptive", False, 0, False),  # Uses adaptive sampling algo.
```
hmstepanek, line 882: I moved this into the new test case below so we can actually check the value of priority.

tests/agent_features/test_distributed_tracing.py
```diff
@@ -1394,3 +1395,68 @@ def _test():
...
+    @override_application_settings(test_settings)
+    @validate_transaction_object_attributes({"sampled": expected_sampled, "priority": expected_priority})
+    @background_task(name="test_distributed_trace_attributes")
+    def _test():
+        pass
+
+    _test()
+
+
+def test_distributed_trace_priority_set_when_only_sampled_set_in_headers(monkeypatch):
```
hmstepanek:
```suggestion
def test_distributed_trace_priority_set_when_only_sampled_set_in_tracestate_header(monkeypatch):
```

PR #1599, opened 2025-12-04
Hybrid agent set status

lrafeei:
This PR adds status setting capabilities as well as adding to the `record_exception` functionality.

(Note that `record_exception` is not fully complete at this point.)

### DIFF
--- a/tests/hybridagent_opentelemetry/test_status.py
+++ b/tests/hybridagent_opentelemetry/test_status.py
@@ -13,13 +13,12 @@
 # limitations under the License.
 
 import pytest
-
 from opentelemetry.trace.status import Status, StatusCode
-from newrelic.api.background_task import background_task
-
 from testing_support.fixtures import dt_enabled
-from testing_support.validators.validate_error_event_attributes import validate_error_event_attributes
 from testing_support.util import conditional_decorator
+from testing_support.validators.validate_error_event_attributes import validate_error_event_attributes
+
+from newrelic.api.background_task import background_task
 
 
 # `set_status` takes in a status argument that can be either
@@ -150,13 +149,37 @@ def test_set_status_with_start_span(tracer, _record_exception, _set_status_on_ex
     "status_to_set,description,expected_status_code",
     [
         (Status(StatusCode.OK), None, StatusCode.OK),  # OK_Status_no_description, no description
-        (Status(StatusCode.OK), "I will be ignored in set_status", StatusCode.OK),  # OK_Status_no_description, description
-        (Status(StatusCode.OK, "I will be ignored in Status"), None, StatusCode.OK),  # OK_Status_with_description, no description
-        (Status(StatusCode.OK, "I will be ignored in Status"), "I will be ignored in set_status", StatusCode.OK),  # OK_Status_with_description, description
+        (
+            Status(StatusCode.OK),
+            "I will be ignored in set_status",
+            StatusCode.OK,
+        ),  # OK_Status_no_description, description
+        (
+            Status(StatusCode.OK, "I will be ignored in Status"),
+            None,
+            StatusCode.OK,
+        ),  # OK_Status_with_description, no description
+        (
+            Status(StatusCode.OK, "I will be ignored in Status"),
+            "I will be ignored in set_status",
+            StatusCode.OK,
+        ),  # OK_Status_with_description, description
         (Status(StatusCode.ERROR), None, StatusCode.ERROR),  # Error_Status_no_description, no description
-        (Status(StatusCode.ERROR), "I will be ignored in set_status", StatusCode.ERROR),  # Error_Status_no_description, description
-        (Status(StatusCode.ERROR, "This is where I belong"), None, StatusCode.ERROR),  # Error_Status_with_description, no description
-        (Status(StatusCode.ERROR, "This is where I belong"), "I will be ignored in set_status", StatusCode.ERROR),  # Error_Status_with_description, description
+        (
+            Status(StatusCode.ERROR),
+            "I will be ignored in set_status",
+            StatusCode.ERROR,
+        ),  # Error_Status_no_description, description
+        (
+            Status(StatusCode.ERROR, "This is where I belong"),
+            None,
+            StatusCode.ERROR,
+        ),  # Error_Status_with_description, no description
+        (
+            Status(StatusCode.ERROR, "This is where I belong"),
+            "I will be ignored in set_status",
+            StatusCode.ERROR,
+        ),  # Error_Status_with_description, description
         (StatusCode.OK, None, StatusCode.OK),  # OK_StatusCode, no description
         (StatusCode.OK, "I will be ignored in Status", StatusCode.OK),  # OK_StatusCode, description
         (StatusCode.ERROR, None, StatusCode.ERROR),  # Error_StatusCode, no description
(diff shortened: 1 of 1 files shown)


newrelic/api/opentelemetry.py
```diff
@@ -238,19 +244,53 @@ def is_recording(self):
         if getattr(self.nr_trace, "end_time", None):
             return False
 
-        return getattr(self.nr_transaction, "priority", 1) > 0
+        # If priority is either not set at this point
```
hmstepanek, line 247: I think this is perhaps a merge conflict cause this is the same logic as the previous code except if it's None we set it to 1?

newrelic/api/opentelemetry.py
```diff
@@ -238,19 +244,53 @@ def is_recording(self):
...
+        https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/trace/__init__.py#L979
+
+        Additional Notes:
+        1. Ignore future calls if status is already set to OK
+            since span should be completed if status is OK.
+        2. Similarly, ignore calls to set to StatusCode.UNSET
+            since this will be either invalid or unnecessary.
+        """
+        if isinstance(status, Status):
+            if (self.status and self.status.status_code is StatusCode.OK) or status.is_unset:
```
hmstepanek: Is it even possible for self.status to be None here?

newrelic/api/opentelemetry.py
```diff
@@ -238,19 +244,53 @@ def is_recording(self):
...
+                return
+            if description is not None:
```
hmstepanek, line 267: Is this just a limitation of the `Status` type? Maybe add a comment here explaining this.

newrelic/api/opentelemetry.py
```diff
@@ -238,19 +244,53 @@ def is_recording(self):
...
 
     def record_exception(self, exception, attributes=None, timestamp=None, escaped=False):
         error_args = sys.exc_info() if not exception else (type(exception), exception, exception.__traceback__)
 
-        if not hasattr(self, "nr_trace"):
-            notice_error(error_args, attributes=attributes)
+        # `escaped` indicates whether the exception has not
+        # been unhandled by the time the span has ended.
+        if attributes:
+            attributes.update({"exception.escaped": escaped})
```
```suggestion
            attributes["exception.escaped"] = escaped
```

newrelic/api/opentelemetry.py
```diff
@@ -238,19 +244,53 @@ def is_recording(self):
...
         else:
-            self.nr_trace.notice_error(error_args, attributes=attributes)
+            attributes = {"exception.escaped": escaped}
```
hmstepanek, line 293: Is this even possible that attributes is None? It would also enter this else clause if attributes was empty which seems unnecessary.

newrelic/api/opentelemetry.py
```diff
@@ -238,19 +244,53 @@ def is_recording(self):
...
+
+        self.set_attributes(attributes)
+
+        notice_error(error_args, attributes=attributes)
```
hmstepanek, line 297: Do we need to map anything to expected or ignored params in notice error here?

newrelic/api/opentelemetry.py
```diff
@@ -283,6 +323,20 @@ def end(self, end_time=None, *args, **kwargs):
...
 
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        """
+        Ends context manager and calls `end` on the `Span`.
+        This is used when span is called as a context manager
+        i.e. `with tracer.start_span() as span:`
+        """
+        if exc_val and self.is_recording():
+            if self._record_exception:
+                self.record_exception(exception=exc_val, escaped=True)
```
hmstepanek, line 338: I know otel does things differently and records the traceback oftentimes in events on the span so I assume that's why we aren't doing anything with the traceback here-is that correct?

tests/hybridagent_opentelemetry/test_attributes.py
```diff
@@ -69,6 +71,7 @@ def newrelic_function_trace():
         otel_span = trace.get_current_span()
         otel_span.set_attribute("otel_span_attribute_FT", "OTel span attribute from FT")
 
+    @dt_enabled
```
hmstepanek, line 74: What's going on with these dt_enabled's?

tests/hybridagent_opentelemetry/test_status.py
```diff
@@ -0,0 +1,140 @@
...
+def conditional_decorator(decorator, condition):
+    def _conditional_decorator(func):
+        if not condition:
+            return func
+        return decorator(func)
+
+    return _conditional_decorator
+
+
+@pytest.mark.parametrize(
```
hmstepanek, line 29: Maybe have a test that tests providing a description?

newrelic/api/opentelemetry.py
lrafeei: I realized that this does something slightly different than that: the `transaction.priority` attribute is always going to exist, but it is initially set to `None`.  This `getattr` function only returns 1 if it does not have the `priority` attribute at all.

So, when it is set to `None` and it tries to do a comparison, it fails.

newrelic/api/opentelemetry.py
lrafeei: For framework instrumentations, no.  If custom instrumentation is used, it is possible not to have attributes defined when starting a span.

newrelic/api/opentelemetry.py
lrafeei: In this case, values like the traceback were recorded by `notice_error()`, so it seemed to be redundant logic.

tests/hybridagent_opentelemetry/test_attributes.py
lrafeei: I had neglected to put them in alongside the `validate_span_events` validators and it just happened to not fail until now.

newrelic/api/opentelemetry.py
lrafeei: I got the impression that these were slightly different interpretations.  My understanding was that 
1. OpenTelemetry's `expected` flag was to denote whether `record_exception()` was called explicitly within a piece of code as opposed to being called within `__end__()`
2. New Relic's `expected` flag was something explicitly passed in through a `notice_error()` call within a piece of code.  If the `expected` flag is not set, it then goes to the application's configuration to see if the error class or error status code are already set in a user's ignore list(s).

Based on my understanding of this, I did not think we should be passing that in under the hood.

newrelic/api/opentelemetry.py
lrafeei: Honestly, I couldn't think of a scenario where it would be, but I got a bit nervous about missing a use case.

newrelic/api/opentelemetry.py
lrafeei: The `set_status` function is a bit confusing initially; it can take in a `Status` (which has attributes `StatusCode` and an optional `description`) or a `StatusCode`.  So, if just the `StatusCode` is passed in, we can pass in a `description` as well.  If a `Status` type is passed in, it should already have the `StatusCode` and `description` in there, so passing in another `description` is redundant at best and conflicting at worse.  So, the `description` being passed in this case is ignored.

newrelic/api/opentelemetry.py
hmstepanek: Ooooh oops-you are right!

newrelic/api/opentelemetry.py
hmstepanek: I'm thinking of a case like this:
```
try:
   raise Exception()
except:
   args = sys.exc_info()

span.__exit__(*args)  # sys.exc_info() = (None, None, None) here.
```

PR #1607, opened 2025-12-12
Hybrid agent WSGI traces

lrafeei:
This PR
* adds WSGI instrumentation capabilities
* a supportability metric for Otel Bridge enabled flag
* changes the name of the enabled/disabled setting from `otel_bridge.enabled` to `opentelemetry.enabled`

### DIFF
--- a/newrelic/api/opentelemetry.py
+++ b/newrelic/api/opentelemetry.py
@@ -356,7 +356,7 @@ class Tracer(otel_api_trace.Tracer):
         self.resource = resource
         self.instrumentation_library = instrumentation_library.split(".")[-1].capitalize()
 
-    def _web_transaction_calls(self):
+    def _create_web_transaction(self):
         if "nr.wsgi.environ" in self.attributes:
             # This is a WSGI request
             transaction = WSGIWebTransaction(self.nr_application, environ=self.attributes.pop("nr.wsgi.environ"))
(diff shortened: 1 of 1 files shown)


newrelic/api/opentelemetry.py
```diff
@@ -432,30 +446,36 @@ def start_span(
...
-                        self.nr_application,
-                        name=name,
-                        scheme=scheme,
-                        host=host,
-                        port=port,
-                        request_method=request_method,
-                        request_path=request_path,
-                        headers=headers,
-                    )
+                    if "nr.wsgi.environ" in self.attributes:
```
hmstepanek: This looks like the same code duplicated-do you think it would be beneficial to refactor this into a function?

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -127,8 +128,13 @@ def wrap_get_current_span(wrapped, instance, args, kwargs):
     if not transaction:
         return span
 
-    settings = transaction.settings or global_settings()
-    if not settings.otel_bridge.enabled:
+    # Do not allow the wrapper to continue if
+    # the Hybrid Agent setting is not enabled
+    application = application_instance(activate=False)
+    app_settings = global_settings() if not application else application.settings
+    settings = transaction.settings or app_settings
```
hmstepanek: Aren't the transaction settings also the application settings? Wondering if this is redundant.

newrelic/api/opentelemetry.py
lrafeei: That's a good point.  And it's only going to get worse when there's ASGI support as well.  Done!

newrelic/hooks/hybridagent_opentelemetry.py
lrafeei: That's a fair point, it is completely redundant since that type of setting is not going to change from transaction to transaction.

newrelic/api/opentelemetry.py
```diff
@@ -356,6 +356,34 @@ def __init__(self, resource=None, instrumentation_library=None, *args, **kwargs)
         self.resource = resource
         self.instrumentation_library = instrumentation_library.split(".")[-1].capitalize()
 
+    def _web_transaction_calls(self):
```
hmstepanek: The only suggestion I have here is maybe name this something different like: `_create_web_transaction` or something similar?

lrafeei: Done!

PR #1608, opened 2025-12-13
Add attrs to partial gran spans & tracking metrics

hmstepanek:
# Overview
* Add fragmentation tracking metric for when we run out of space to add more `nr.ids` in compact mode.
* Add `nr.pg` to entry span to track when a transaction is partial granularity.
* Keep error attributes on partial granularity spans.
* Only record partial granularity type supportability metric if partial granularity transactions are actually sent.
* Small fixup to logic to make it more clear (`not a==b` => `a!=b`).
* Fix bug in metric reporting for kept and instrumented.

### DIFF
--- a/newrelic/api/opentelemetry.py
+++ b/newrelic/api/opentelemetry.py
@@ -21,6 +21,7 @@ from opentelemetry.baggage.propagation import W3CBaggagePropagator
 from opentelemetry.propagate import set_global_textmap
 from opentelemetry.propagators.composite import CompositePropagator
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
+from opentelemetry.trace.status import Status, StatusCode
 
 from newrelic.api.application import application_instance
 from newrelic.api.background_task import BackgroundTask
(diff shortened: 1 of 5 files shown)


newrelic/core/node_mixin.py
```diff
@@ -79,71 +79,91 @@ def span_event(
...
-                # of ids on the seen span, compute the new duration & start time, and
-                # return None.
+                return [i_attrs, {}, {key: a_attrs.get(key) for key in exit_span_error_attrs_present | {"nr.pg"}}]
+        # If the span is not an exit span, skip it by returning None.
+        if not exit_span_attrs_present:
+            return None
+        # If the span is an exit span and we are in reduced mode (meaning no attribute dropping),
+        # just return the exit span as is.
+        if partial_granularity_type == "reduced":
+            ct_exit_spans["kept"] += 1
```
hmstepanek, line 108: There was a bug here that I fixed which is why this is here now. I had to move these into here because gets incremented multiple times if it's put in span_events because span_events loops over the children and then the children of the children (so it gets incremented multiple times each time is yields the node up the chain) where as here it is guaranteed that it is only incremented once.

PR #1611, opened 2025-12-16
Remove Python 3.8 Support

TimPansino:
# Overview

* Drop support for Python 3.8.
* Remove all code paths for Python 3.8 and below.
* Run linters/formatters to upgrade syntax where appropriate.


umaannamalai:
I was double checking the repo for references to Python 3.8 and noticed this function in `newrelic/core/environment.py` which could be updated to cover multiple version removals we've already done as well as 3.8:

```
def _get_stdlib_builtin_module_names():
    builtins = set(sys.builtin_module_names)
    # Since sys.stdlib_module_names is not available in versions of python below 3.10,
    # use isort's hardcoded stdlibs instead.
    python_version = sys.version_info[0:2]
    if python_version < (3,):
        stdlibs = isort_stdlibs.py27.stdlib
    elif (3, 7) <= python_version < (3, 8):
        stdlibs = isort_stdlibs.py37.stdlib
    elif python_version < (3, 9):
        stdlibs = isort_stdlibs.py38.stdlib
    elif python_version < (3, 10):
        stdlibs = isort_stdlibs.py39.stdlib
    elif python_version >= (3, 10):
        stdlibs = sys.stdlib_module_names
    else:
        _logger.warning("Unsupported Python version. Unable to determine stdlibs.")
        return builtins
    return builtins | stdlibs
```

We could also update this comment in `pyproject.toml`:

`"UP006",  # non-pep585-annotation (not compatible with Python 3.8)`

PR #1613, opened 2025-12-19
[DRAFT] Rethink Azure Functions Utilization

TimPansino:
# Overview

* Guard against crashes in Azure Functions Utilization
* Add regression tests for this behavior

### DIFF
--- a/newrelic/hooks/framework_azurefunctions.py
+++ b/newrelic/hooks/framework_azurefunctions.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import os
-import re
 import urllib.parse as urlparse
 
 from newrelic.api.application import application_instance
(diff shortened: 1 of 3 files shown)


newrelic/hooks/framework_azurefunctions.py
```diff
@@ -35,18 +33,12 @@ def intrinsics_populator(application, context):
...
-    elif website_owner_name:
-        resource_group_name = AZURE_RESOURCE_GROUP_NAME_PARTIAL_RE.search(website_owner_name).group(1)
-    else:
-        resource_group_name = os.environ.get("WEBSITE_RESOURCE_GROUP", "Unknown")
+
+    subscription_id = "Unknown"
+    if website_owner_name and "+" in website_owner_name:
+        subscription_id = website_owner_name.split("+")[0] or "Unknown"
+
+    resource_group_name = os.environ.get("WEBSITE_RESOURCE_GROUP", "Unknown")
```
lrafeei, line 41: The biggest issue here is that the `WEBSITE_RESOURCE_GROUP` environment variable is not guaranteed to be present.

lrafeei: More specifically, it is not present in the Consumer plan and it is not always present in the Flex Consumer plan at the time we would need the value of this.

tests/framework_azurefunctions/test_utilization.py
```diff
@@ -0,0 +1,39 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from newrelic.common.utilization import AzureFunctionUtilization
+
+
+def test_utilization(monkeypatch):
+    monkeypatch.setenv("REGION_NAME", "EastUS2")
```
lrafeei, line 19: Turns out this is case sensitive when being passed into the application: it should be all lowercase


```suggestion
    monkeypatch.setenv("REGION_NAME", "eastus2")
```

lrafeei (changes requested):

PR #1617, opened 2025-12-29
Hybrid agent ASGI, DB Traces, and External Traces support

lrafeei:
This PR adds support and testing for the following:

- ASGI Web Transactions (SERVER kind)
- DB Traces (CLIENT kind)
- External Traces (CLIENT kind)

### DIFF
--- a/newrelic/core/application.py
+++ b/newrelic/core/application.py
@@ -597,17 +597,9 @@ class Application:
                 internal_metric("Supportability/Python/AzureFunctionMode/enabled", 1)
 
             # OpenTelemetry Bridge toggle metric
-            opentelemetry_bridge = configuration.opentelemetry.enabled
-            internal_metric(
-                f"Supportability/Tracing/Python/OpenTelemetryBridge/{'enabled' if opentelemetry_bridge else 'disabled'}",
-                1,
-            )
-
-            # OpenTelemetry Bridge toggle metric
             opentelemetry_bridge = "enabled" if configuration.opentelemetry.enabled else "disabled"
-            f"Supportability/Tracing/Python/OpenTelemetryBridge/{opentelemetry_bridge}"
             internal_metric(
-                f"Supportability/Tracing/Python/OpenTelemetryBridge/{'enabled' if opentelemetry_bridge else 'disabled'}",
+                f"Supportability/Tracing/Python/OpenTelemetryBridge/{opentelemetry_bridge}",
                 1,
             )
 


newrelic/api/opentelemetry.py
```diff
@@ -216,10 +217,17 @@ def set_attributes(self, attributes):
             self.set_attribute(key, value)
 
     def _set_attributes_in_nr(self, otel_attributes=None):
-        if not (otel_attributes and hasattr(self, "nr_trace") and self.nr_trace):
+        if not otel_attributes or not getattr(self, "nr_trace"):
```
hmstepanek: You have to pass the default arg in otherwise it errors. The only one that doesn't do this is the dict.get method-that one automatically returns None if the key does not exist.
```suggestion
        if not otel_attributes or not getattr(self, "nr_trace", None):
```

newrelic/api/opentelemetry.py
```diff
@@ -216,10 +217,17 @@ def set_attributes(self, attributes):
...
             return
+        
+        # If these attributes already exist in NR's intrinsic or agent
+        # attributes, keep the attributes in the OTel span, but do not
+        # add them to NR's user attributes to avoid sending the same
+        # data multiple times.
+        combined_attrs = {**self.nr_trace.agent_attributes, **self.nr_trace.user_attributes}
```
hmstepanek: You might be able to do this logic more simply with a set:
```suggestion
        otel_user_attrs = (set(self.nr_trace.agent_attributes) | set(self.nr_trace.user_attributes)) & set(otel_attributes)
        self.nr_trace.add_custom_attributes({key: otel_attributes[key] for key in otel_user_attrs}) 
```

newrelic/api/opentelemetry.py
```diff
@@ -216,10 +217,17 @@ def set_attributes(self, attributes):
...
         for key, value in otel_attributes.items():
-            self.nr_trace.add_custom_attribute(key, value)
+            if key not in combined_attrs:
```
hmstepanek: Question: If the nr trace has an existing user attribute that matches an otel attribute name, this does not override the nr user attribute with the otel attribute value-is that the correct behavior that we want?

newrelic/api/opentelemetry.py
```diff
@@ -296,6 +304,63 @@ def record_exception(self, exception, attributes=None, timestamp=None, escaped=F
...
+            return self.attributes.get("db.system").capitalize()
+        
+        def _dynamodb_attribute_mapping():
+            host = _host()
+            region = self.attributes.get("cloud.region")
+            operation = self.attributes.get("db.operation")
+            target = self.attributes.get("aws.dynamodb.table_names", [None])[-1]
+            account_id = self.nr_transaction.settings.cloud.aws.account_id
+
+            if region and account_id and target:
```
hmstepanek: Can you abstract this logic into a function inside a common file and then import and use it in here-that way if we ever have to change this it's in one spot instead of multiple locations? Maybe put it in `newrelic/common/aws.py` and call it `generate_dynamodb_arn`.

newrelic/api/opentelemetry.py
```diff
@@ -296,6 +304,63 @@ def record_exception(self, exception, attributes=None, timestamp=None, escaped=F
 
         notice_error(error_args, attributes=attributes)
 
+    def _database_attribute_mapping(self):
```
hmstepanek, line 306: I like where you are going with this in terms of wrapping all the attributes up in some sort of mapper container. I think maybe we should roll with this concept but make it a bit more official. Can you move this outside of this class and make it a function that returns a dictionary map of the attributes? So something like:
```
def map_database_attributes(attributes):
   return {
       host: attributes.get("net.peer.name") or attributes.get("server.address"),
       ...
    }
```

Then inside here you can just do something like:
```
attributes = map_database_attributes(self.attributes)
self.nr_trace._add_agent_attributes({
    "cloud.region": region,
    "aws.requestId": "aws.request_id",
})
if mapped_attrs["host"]:
    self.nr_trace.host = mapped_attrs["host"]
...
```

And we should probably do this for all the different types of traces and their attr maps as well.

newrelic/api/opentelemetry.py
```diff
@@ -296,6 +304,63 @@ def record_exception(self, exception, attributes=None, timestamp=None, escaped=F
...
+        
+        self.nr_trace.database_name = _database_name() or self.nr_trace.database_name
+        self.nr_trace.host = _host() or self.nr_trace.host
+        self.nr_trace.port_path_or_id = _port_path_or_id() or self.nr_trace.port_path_or_id
+        self.nr_trace.product = _product() or self.nr_trace.product.capitalize()
+
+        db_statement = self.attributes.get("db.statement")
+        
+        if isinstance(db_statement, (str, bytes)):
+            operation = _parse_operation(db_statement)
```
hmstepanek: Put this in `newrelic/core/database_utils.py`:
```suggestion
def get_database_target_from_statement(db_statement):
            operation = _parse_operation(db_statement)
            target = _parse_target(db_statement, operation)
            return target
```

newrelic/api/opentelemetry.py
```diff
@@ -305,36 +355,30 @@ def end(self, end_time=None, *args, **kwargs):
...
         self.set_status(StatusCode.OK if not error[0] else StatusCode.ERROR)
 
-        if ("exception.escaped" in self.attributes) or (
-            self.kind in (otel_api_trace.SpanKind.SERVER, otel_api_trace.SpanKind.CONSUMER)
-            and isinstance(current_trace(), Sentinel)
-        ):
+        if ("exception.escaped" in self.attributes) or (self.kind in (otel_api_trace.SpanKind.SERVER, otel_api_trace.SpanKind.CONSUMER) and isinstance(current_trace(), Sentinel)):
             # We need to end the transaction as well
-            self.nr_transaction.__exit__(*error)
+            self.nr_transaction.__exit__(*sys.exc_info())
```
hmstepanek: I think you set `error = sys.exc_info` up above so I think this should still be `*error` here?
```
self.nr_transaction.__exit__(*error)
```

newrelic/api/opentelemetry.py
lrafeei: That's fair.  I suppose if a user explicitly sets that attribute, we shouldn't override it.  I just had not thought of that option and was trying to eliminate redundant looping logic.  I'll just get rid of the existing user attributes check and only check for the agent attributes.

newrelic/api/opentelemetry.py
```diff
@@ -360,6 +413,29 @@ def _create_web_transaction(self):
         if "nr.wsgi.environ" in self.attributes:
             # This is a WSGI request
             transaction = WSGIWebTransaction(self.nr_application, environ=self.attributes.pop("nr.wsgi.environ"))
+        elif "nr.asgi.scope" in self.attributes:
+            # This is an ASGI request
+            scope = self.attributes.pop("nr.asgi.scope")
+            scheme = scope.get("scheme", "http")
+            if "server" in scope and scope["server"] is not None:
```
```suggestion
            server = scope.get("server") or (None, None)
            host, port = scope["server"] = tuple(server)
```

newrelic/api/opentelemetry.py
```diff
@@ -360,6 +413,29 @@ def _create_web_transaction(self):
...
+                host, port = scope["server"] = tuple(scope["server"])
+            else:
+                host, port = None, None
+            request_method = scope["method"]
```
hmstepanek: Should we be using scope.get in case these keys don't exist?

newrelic/core/application.py
```diff
@@ -603,6 +603,13 @@ def connect_to_data_collector(self, activate_agent):
                 1,
             )
 
+            # OpenTelemetry Bridge toggle metric
+            opentelemetry_bridge = configuration.opentelemetry.enabled
```
hmstepanek: Maybe just put the logic outside of the string above-I find it's a bit easier to read that way.
```suggestion
            opentelemetry_bridge = "enabled" if configuration.opentelemetry.enabled else "disabled"
            f"Supportability/Tracing/Python/OpenTelemetryBridge/{opentelemetry_bridge}"
```

newrelic/core/database_utils.py
```diff
@@ -419,6 +419,12 @@ def _parse_operation(sql):
     return operation if operation in _operation_table else ""
 
 
+def _parse_operation_otel(sql):
+    match = _parse_operation_re.search(sql)
+    operation = (match and match.group(1).lower()) or ""
```
hmstepanek: I don't think we need to do the `or ""` twice.
```suggestion
    operation = (match and match.group(1).lower())
```

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -137,9 +132,8 @@ def wrap_get_tracer_provider(wrapped, instance, args, kwargs):
 
     if _TRACER_PROVIDER is None:
         from newrelic.api.opentelemetry import TracerProvider
-
-        hybrid_agent_tracer_provider = TracerProvider("hybrid_agent_tracer_provider")
-        _TRACER_PROVIDER = hybrid_agent_tracer_provider
+        _TRACER_PROVIDER = TracerProvider()
```
hmstepanek, line 135: Just making sure we don't still need to pass `"hybrid_agent_tracer_provider"` into here?

tests/hybridagent_fastapi/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+    "debug.record_transaction_failure": True,
+    "debug.log_autorum_middleware": True,
+    "opentelemetry.enabled": True,
+}
+
+collector_agent_registration = collector_agent_registration_fixture(
+    app_name="Python Agent Test (Hybrid Agent, FastAPI)", default_settings=_default_settings
+)
+
+os.environ["NEW_RELIC_CONFIG_FILE"] = str(Path(__file__).parent / "newrelic_fastapi.ini")
```
hmstepanek, line 41: Maybe take a look at [this fixture](https://github.com/newrelic/newrelic-python-agent/blob/main/tests/agent_unittests/conftest.py#L48) and see if you can re-use it or do something similar instead of having these .ini files and setting env var variables/etc.

Usage example [here](https://github.com/newrelic/newrelic-python-agent/blob/main/tests/agent_unittests/test_distributed_tracing_settings.py#L30).

newrelic/api/opentelemetry.py
lrafeei: Whoops, you're right.  Not sure why I changed it...

newrelic/hooks/hybridagent_opentelemetry.py
lrafeei: Oh, I changed this logic slightly to be more in line with the actual TracerProvider's functionality.  I'll flesh it out some more in the future, but basically, the OTel TracerProvider doesn't really take in a name.

tests/hybridagent_fastapi/conftest.py
lrafeei: This is a fair point.  I figured this would be going away soon anyway, so I didn't bother to try to clean up that logic.  Basically, this will be part of the internal configuration setup and it will not be necessary to do this whole setting of the ini files.

hmstepanek (approved):

PR #1619, opened 2026-01-03
Hybrid agent Message Queue Traces and Transactions

lrafeei:
This PR adds support and testing for the following:

- Makes use of the sentinel span for trace/span mapping
- MessageTraces and MessageTransactions (PRODUCER/CONSUMER kind)

### DIFF
--- a/newrelic/api/opentelemetry.py
+++ b/newrelic/api/opentelemetry.py
@@ -12,10 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import logging
 import sys
 from contextlib import contextmanager
-import json
 
 from opentelemetry import trace as otel_api_trace
 from opentelemetry.baggage.propagation import W3CBaggagePropagator
@@ -24,7 +24,6 @@ from opentelemetry.propagators.composite import CompositePropagator
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
 from opentelemetry.trace.status import Status, StatusCode
 
-from newrelic.core.database_utils import generate_dynamodb_arn, get_database_operation_target_from_statement
 from newrelic.api.application import application_instance
 from newrelic.api.background_task import BackgroundTask
 from newrelic.api.datastore_trace import DatastoreTrace
@@ -35,6 +34,7 @@ from newrelic.api.message_transaction import MessageTransaction
 from newrelic.api.time_trace import current_trace, notice_error
 from newrelic.api.transaction import Sentinel, current_transaction
 from newrelic.api.web_transaction import WebTransaction, WSGIWebTransaction
+from newrelic.core.database_utils import generate_dynamodb_arn, get_database_operation_target_from_statement
 from newrelic.core.otlp_utils import create_resource
 
 _logger = logging.getLogger(__name__)
@@ -55,16 +55,15 @@ class NRTraceContextPropagator(TraceContextTextMapPropagator):
         }
 
         extracted_context = super().extract(carrier=carrier, context=context, getter=getter)
-        
+
         if transaction:
             transaction.accept_distributed_trace_headers(nr_headers)
         else:
             # A situation where extract() is called BEFORE a span
             # creation (such as in the kafka consumer instrumentation)
             extracted_context._nr_headers_outside_transaction = nr_headers
-            
+
         return extracted_context
-        
 
     def inject(self, carrier, context=None, setter=None):
         transaction = current_transaction()
@@ -343,28 +342,24 @@ class Span(otel_api_trace.Span):
                 db_statement = db_statement.string
             operation, target = get_database_operation_target_from_statement(db_statement)
             target = target or self.attributes.get("db.mongodb.collection")
-            span_obj_attrs.update({
-                "operation": operation,
-                "target": target,
-            })
+            span_obj_attrs.update({"operation": operation, "target": target})
         elif span_obj_attrs["product"] == "Dynamodb":
             region = self.attributes.get("cloud.region")
             operation = self.attributes.get("db.operation")
             target = self.attributes.get("aws.dynamodb.table_names", [None])[-1]
             account_id = self.nr_transaction.settings.cloud.aws.account_id
             resource_id = generate_dynamodb_arn(span_obj_attrs["host"], region, account_id, target)
-            agent_attrs.update({
-                "aws.operation": self.attributes.get("db.operation"),
-                "cloud.resource_id": resource_id,
-                "cloud.region": region,
-                "aws.requestId": self.attributes.get("aws.request_id"),
-                "http.statusCode": self.attributes.get("http.status_code"),
-                "cloud.account.id": account_id,
-            })
-            span_obj_attrs.update({
-                "target": target,
-                "operation": operation,
-            })
+            agent_attrs.update(
+                {
+                    "aws.operation": self.attributes.get("db.operation"),
+                    "cloud.resource_id": resource_id,
+                    "cloud.region": region,
+                    "aws.requestId": self.attributes.get("aws.request_id"),
+                    "http.statusCode": self.attributes.get("http.status_code"),
+                    "cloud.account.id": account_id,
+                }
+            )
+            span_obj_attrs.update({"target": target, "operation": operation})
 
         # We do not want to override any agent attributes
         # with `None` if `value` does not exist.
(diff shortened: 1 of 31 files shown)


tests/hybridagent_opentelemetry/test_hybrid_cross_agent.py
```diff
@@ -135,12 +137,10 @@ def test_opentelemetry_api_can_add_custom_attributes_to_spans(tracer):
 def test_opentelemetry_api_can_record_errors(tracer):
     application = application_instance(activate=False)
 
-    with BackgroundTask(application, name="Foo"):
-        with tracer.start_as_current_span(name="Bar", kind=otel_api_trace.SpanKind.INTERNAL):
-            try:
+    with pytest.raises(Exception):
```
hmstepanek:
```suggestion
    with pytest.raises(Exception, match=r"Test exception message"):
```

tests/hybridagent_fastapi/test_otel_application.py
```diff
@@ -51,15 +51,15 @@ def test_application(caplog, app, endpoint):
...
         exact_intrinsics=_exact_intrinsics,
         expected_intrinsics=_expected_child_intrinsics,
         unexpected_intrinsics=_unexpected_child_intrinsics,
     )
     @validate_transaction_metrics(
         transaction_name,
-        scoped_metrics=[(f"Function/{transaction_name}", 1)],
-        rollup_metrics=_test_application_rollup_metrics + [(f"Function/{transaction_name}", 1)],
+        scoped_metrics=[(f"Function/{transaction_name} http send", 2)],
+        rollup_metrics=_test_application_rollup_metrics + [(f"Function/{transaction_name} http send", 2)],
```
hmstepanek: Fixup for megalinter:
```suggestion
        rollup_metrics=[(f"Function/{transaction_name} http send", 2), *_test_application_rollup_metrics],
```

newrelic/api/opentelemetry.py
```diff
@@ -148,16 +159,28 @@ def __init__(
...
                 self.otel_parent,
                 current_nr_trace,  # NR parent trace
             )
             return
 
-        if nr_trace_type == FunctionTrace:
+        if not create_nr_trace:
+            # Do not create a New Relic trace for this OTel span.
+            # While this OTel span exists it will not be explicitly
+            # translate to a NR trace.  This may occur during the
```
```suggestion
            # translated to a NR trace.  This may occur during the
```

newrelic/api/opentelemetry.py
```diff
@@ -148,16 +159,28 @@ def __init__(
...
+            # creation of a Transaction, which will create the root
+            # span.  This may also occur during special cases, such
```
hmstepanek, line 175: Question about this-the spec says the following:

> - Creating a span with a span kind of `SpanKind.CONSUMER` will start a `OtherTransaction/*` transaction 
and create a corresponding segment.
> - If the `CONSUMER` span occurs within an already existing NR transaction, it will be included in
the transaction trace.

Wouldn't the above mean that a nr trace should be created on multiple calls to the consumer?

newrelic/api/opentelemetry.py
```diff
@@ -353,10 +378,9 @@ def _database_attribute_mapping(self):
     def end(self, end_time=None, *args, **kwargs):
         # We will ignore the end_time parameter and use NR's end_time
 
-        # Check to see if New Relic trace ever existed or,
-        # if it does, that trace has already ended
-        nr_trace = hasattr(self, "nr_trace")
-        if not nr_trace or (nr_trace and getattr(nr_trace, "end_time", None)):
+        # Check to see if New Relic trace ever existed
+        # or, if it does, that trace has already ended
+        if getattr(self.nr_trace, "end_time", None):
```
hmstepanek: This isn't quite the same behavior as the previous. In the previous if the self.nr_trace was None it would also return. Do we want to match the previous behavior still?

```suggestion
        if not self.nr_trace or getattr(self.nr_trace, "end_time", None):
```

newrelic/api/opentelemetry.py
```diff
@@ -369,20 +393,36 @@ def end(self, end_time=None, *args, **kwargs):
...
 
-        # External specific attributes
+        # External/Web specific attributes
         self.nr_trace._add_agent_attribute("http.statusCode", self.attributes.get("http.status_code"))
 
+        # Message specific attributes
+        if self.attributes.get("messaging.system"):
+            destination_name = self.attributes.get("messaging.destination")
+            self.nr_transaction.destination_name = destination_name
+            bootstrap_servers = ast.literal_eval(self.attributes.get("messaging.url", "[]"))
```
hmstepanek: Is the eval necessary here?

newrelic/api/opentelemetry.py
```diff
@@ -369,20 +393,36 @@ def end(self, end_time=None, *args, **kwargs):
         if self.attributes.get("db.system"):
             self._database_attribute_mapping()
 
-        # External specific attributes
+        # External/Web specific attributes
         self.nr_trace._add_agent_attribute("http.statusCode", self.attributes.get("http.status_code"))
 
+        # Message specific attributes
+        if self.attributes.get("messaging.system"):
+            destination_name = self.attributes.get("messaging.destination")
```
hmstepanek, line 396: Just directly store this-it's easier to read that way.
```suggestion
            self.nr_transaction.destination_name = self.attributes.get("messaging.destination")
```

newrelic/api/opentelemetry.py
```diff
@@ -369,20 +393,36 @@ def end(self, end_time=None, *args, **kwargs):
...
+            self.nr_transaction.destination_name = destination_name
```
hmstepanek, line 397:
```suggestion
```

newrelic/api/opentelemetry.py
```diff
@@ -369,20 +393,36 @@ def end(self, end_time=None, *args, **kwargs):
...
-            # We need to end the transaction as well
-            self.nr_transaction.__exit__(*error)
-
+        
+        # Only if unhandled exception do we want to abruptly end.
+        # Otherwise, ensure that the span is the last one to end.
+        if getattr(self.attributes, "exception.escaped", False) or (self.kind in (otel_api_trace.SpanKind.SERVER, otel_api_trace.SpanKind.CONSUMER) and isinstance(current_trace(), Sentinel)):
+            # We need to end the transaction, which will
+            # end the sentinel trace as well.
+            self.nr_transaction.__exit__(*sys.exc_info())
```
```suggestion
            self.nr_transaction.__exit__(*error)
```

newrelic/api/opentelemetry.py
```diff
@@ -237,8 +262,8 @@ def add_link(self, context=None, attributes=None):
         raise NotImplementedError("Not implemented yet.")
 
     def update_name(self, name):
-        # Sentinel, MessageTrace, DatastoreTrace, and ExternalTrace
-        # types do not have a name attribute
+        # NOTE: Sentinel, MessageTrace, DatastoreTrace, and
+        # ExternalTrace types do not have a name attribute.
         self._name = name
         if hasattr(self, "nr_trace") and hasattr(self.nr_trace, "name"):
```
hmstepanek, line 267: It looks like we are missing coverage for this function.

newrelic/api/opentelemetry.py
hmstepanek: It looks like we are missing coverage in the tests for this case.

tests/hybridagent_kafkapython/test_producer.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+    @background_task(name="test_distributed_tracing_headers_under_terminal")
+    @cache_kafka_producer_headers
+    @validate_messagebroker_headers
+    def test():
+        with FunctionTrace(name="terminal_trace", terminal=True):
+            send_producer_message()
+
+    test()
+
```
hmstepanek, line 88: Wondering if we should have a test here to validate that a transaction and trace are not created for a producer in the case that the span exists outside a transaction?

> - Creating a span with a span kind of `SpanKind.PRODUCER` will not start a NR transaction.
> - If the `PRODUCER` span occurs within an already existing NR transaction, a segment will
be created and it will be included in the transaction trace.

hmstepanek (changes requested):
In general implementation looks good. I think the main thing on my mind for this PR is the  [consumer and producer parts of the agent spec](https://source.datanerd.us/agents/agent-specs/blob/main/otel_bridge/Tracing-API.md#spankindproducer) as compared to the tests and some of the comments here-they seem to be a bit at odds with each other but maybe I'm just missing something. I vaguely remember our behavior being different than some of the other agent teams in regards to when we create a trace vs a transaction for message brokers so that might be where my confusion stems from.

newrelic/api/opentelemetry.py
lrafeei: Yeah, that was super frustrating.  It returns a string of the list.  So, instead of 
`bootstrap_servers = ["some_ip_address", "another_ip_address"]`
it returns
`bootstrap_servers = '["some_ip_address", "another_ip_address"]'`

newrelic/api/opentelemetry.py
lrafeei: Yep!  For the rest of the message queues that is what we do.  For Kafka, we ended up treating the consumer as a transaction and the producer as a trace.  But yes, I have another PR for the rest of the queues that tweaks this logic.

newrelic/api/opentelemetry.py
lrafeei: Oof, not sure how that piece of the logic got removed--that should have stayed in there...

newrelic/api/opentelemetry.py
lrafeei: Oh, I remember why I stored it as a separate variable instead of directly putting it in as a transaction attribute: I also use that name in the Kafka Nodes metric a few lines below.

tests/hybridagent_kafkapython/test_producer.py
lrafeei: That's fair.  I'll put in a test for the first scenario and the second scenario I will use an existing consumer test to prove that this is the case

newrelic/api/opentelemetry.py
hmstepanek: Oh--oops missed that. Ok fine as is then!

PR #1622, opened 2026-01-09
Add support and tests for RabbitMQ

lrafeei:
This PR adds tests/additional support for RabbitMQ


newrelic/api/opentelemetry.py
```diff
@@ -201,10 +173,10 @@ def __init__(
         elif nr_trace_type == MessageTrace:
             trace_kwargs = {
                 "library": self.instrumenting_module,
-                "operation": "Produce",
-                "destination_type": "Topic",
+                "operation": "Produce" if self.kind == otel_api_trace.SpanKind.PRODUCER else "Consume",
+                "destination_type": "Exchange", # For Kafka, this will be overridden
                 "destination_name": self.name,
-                "params": self.attributes,
+                "params": self.attributes,
```
hmstepanek: You would think it would remove the extra space automatically. 
```suggestion
                "params": self.attributes,
```

newrelic/api/opentelemetry.py
```diff
@@ -327,6 +299,75 @@ def record_exception(self, exception, attributes=None, timestamp=None, escaped=F
 
         notice_error(error_args, attributes=attributes)
 
+    def _messagequeue_attribute_mapping(self):
+        host = self.attributes.get("net.peer.name") or self.attributes.get("server.address")
+        port = self.attributes.get("net.peer.port") or self.attributes.get("server.port")
+        name = self.name.split()[0] # OTel's format for this is "name operation"
```
hmstepanek: Thinking this could improve performance slightly as it won't have to keep searching after it finds the space:
```suggestion
        name = self.name.split(maxsplit=1)[0] # OTel's format for this is "name operation"
```

newrelic/api/opentelemetry.py
```diff
@@ -327,6 +299,75 @@ def record_exception(self, exception, attributes=None, timestamp=None, escaped=F
...
+        
+        # Logic for Pika/RabbitMQ
+        span_obj_attrs = {
+            "library": self.attributes.get("messaging.system").capitalize(),
```
hmstepanek: We don't need to handle the case where messaging.system doesn't exist right?

newrelic/api/opentelemetry.py
```diff
@@ -327,6 +299,75 @@ def record_exception(self, exception, attributes=None, timestamp=None, escaped=F
...
+            "destination_name": name,   # OTel's format for this is "name operation"
+        }
+        
+        # Keep this to simplify Transaction naming logic later on
+        if span_obj_attrs["library"] == "Rabbitmq":
```
hmstepanek: It looks like we use "RabbitMQ" in our code today. Seems like we might want this to match for consistency.

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -234,6 +226,56 @@ def wrap_start_internal_or_server_span(wrapped, instance, args, kwargs):
...
+    elif not channel:
+        # This is a consumer generator call
+        # Create a new transaction only.
+        # if transaction already exists, a new one
+        # will not be created and nothing will occur.
+        # This is the current behavior that Kafka has
+        tracer._create_consumer_trace = False
+    
+    params = {"task_name": task_name}
+    for property in properties_to_extract:
```
```suggestion
    for _property in properties_to_extract:
```

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -234,6 +226,56 @@ def wrap_start_internal_or_server_span(wrapped, instance, args, kwargs):
...
+        value = getattr(properties, property, None)
```
```suggestion
        value = getattr(properties, _property, None)
```

newrelic/hooks/hybridagent_opentelemetry.py
```diff
@@ -234,6 +226,56 @@ def wrap_start_internal_or_server_span(wrapped, instance, args, kwargs):
...
+        if properties and value:
+            params[property] = value
```
```suggestion
            params[_property] = value
```

hmstepanek (changes requested):
Just a couple minor things other wise looks good.

newrelic/api/opentelemetry.py
lrafeei: This seems to be the convention for OTel's message queues (also, in this specific case, it will never get here if `messaging.system` does not exist because that is the conditional required to get into this function)

newrelic/api/opentelemetry.py
lrafeei: I thought about this and I realized the issue here is that OTel seems to keep these lowercased.  I am not sure at what point I am just overriding the existing OTel behavior.  I feel as though at that point, I have to do this for every library.

hmstepanek: I guess it depends on if it's used for anything important in the UI/pipeline where the difference in this value would cause a mismatch in something important like an entity connection, etc. I don't see that it's used in the entity def tables so probably we could just leave the value as the original from OTEL. I wonder if there's any reason to bother capitalizing it if it's still not going to match?

lrafeei: For some reason I thought that was an agent spec requirement (for the name to be capitalized) but I can't seem to find where it says that, so maybe I just hallucinated that requirement all together...

hmstepanek (approved):

PR #1623, opened 2026-01-09
Improve Strands Tool Error Capturing

TimPansino:
# Overview

* Add additional instrumentation to Strands Agents to capture errors in synchronous tool functions decorated with `@tool`.
* Change coverage of all tools in Strands to track errors emitted from `stream()` instead of relying on the `_tool_func` attribute for better coverage.
* Overhaul Strands Agents testing to better cover the entire matrix of tool types and invocation functions.

### DIFF
--- a/newrelic/common/utilization.py
+++ b/newrelic/common/utilization.py
@@ -233,21 +233,29 @@ class AzureFunctionUtilization(CommonUtilization):
     HEADERS = {"Metadata": "true"}  # noqa: RUF012
     VENDOR_NAME = "azurefunction"
 
-    @staticmethod
-    def fetch():
+    @classmethod
+    def fetch(cls):
         cloud_region = os.environ.get("REGION_NAME")
         website_owner_name = os.environ.get("WEBSITE_OWNER_NAME")
         azure_function_app_name = os.environ.get("WEBSITE_SITE_NAME")
 
         if all((cloud_region, website_owner_name, azure_function_app_name)):
-            if website_owner_name.endswith("-Linux"):
-                resource_group_name = AZURE_RESOURCE_GROUP_NAME_RE.search(website_owner_name).group(1)
-            else:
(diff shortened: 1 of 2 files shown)


newrelic/hooks/mlmodel_strands.py
```diff
@@ -464,6 +477,15 @@ def wrap_bedrock_model__stream(wrapped, instance, args, kwargs):
     return wrapped(*args, **kwargs)
 
 
+def wrap_decorated_function_tool__wrap_tool_result(wrapped, instance, args, kwargs):
+    transaction = current_transaction()
+    if transaction:
+        # notice_error will handle pulling from sys.exc_info() if there's an active exception.
+        transaction.notice_error()
```
umaannamalai: Do we have a handle to the `tool_id` here (stored in the `strands_attrs`? We typically pass the `tool_id`/ `agent_id` in as an attribute to the `notice_error` call.

newrelic/hooks/mlmodel_strands.py
```diff
@@ -415,11 +415,24 @@ async def aclose(self):
 
 
 def wrap_ToolRegister_register_tool(wrapped, instance, args, kwargs):
-    bound_args = bind_args(wrapped, args, kwargs)
+    try:
+        from strands.tools.decorator import DecoratedFunctionTool
```
umaannamalai, line 420: Is it worth adding a log message here so we know why we early exited in case things move around in Strands?

newrelic/hooks/mlmodel_strands.py
TimPansino: I was able to add `tool_id`, no handle to `agent_id` though.

newrelic/hooks/mlmodel_strands.py
TimPansino: Done

PR #1628, opened 2026-01-15
Hybrid Agent SpanLink and SpanEvent Events Implementation

lrafeei:
This PR provides support for SpanLink and SpanEvent Events as well as modifies the instrumentation for `get_tracer_provider` and `set_tracer_provider`


newrelic/core/stats_engine.py
```diff
@@ -1189,7 +1209,8 @@ def record_transaction(self, transaction):
                     self._span_stream.put(event)
             elif transaction.sampled:
                 for event in transaction.span_events(self.__settings):
-                    self._span_events.add(event, priority=transaction.priority)
+                    new_event = self._add_spanlink_and_spanevent_events(event, transaction.priority)
```
hmstepanek: I think this entire function is unecessary. All it seems to do is parse the span links and events from a list back into a list. I was thinking filter could end up dropping span links and so you end up with the following:
```
[event, [], span_events] -> [event, span_events]
```
but I think that's ok as we don't really need to know what is what after we add it to the queue.

newrelic/api/time_trace.py
```diff
@@ -215,6 +217,73 @@ def add_code_level_metrics(self, source):
...
+        settings = self.settings
+        if not settings:
+            return
+
+        if not settings.opentelemetry.enabled:
+            return
+
+        if len(self.span_link_events) >= 100:
+            self.transaction._record_supportability("Supportability/SpanEvent/Links/Dropped")
+            _logger.debug(
```
hmstepanek: I wonder if we really need this log given we have a metric for it?

tests/hybridagent_pika/test_pika_blocking_connection_consume_generator.py
```diff
@@ -85,7 +82,7 @@ def test_blocking_connection_consume_connection_close(producer):
     channel = connection.channel()
 
     try:
-        for method_frame, _properties, body in channel.consume(QUEUE):
+        for _, _properties, body in channel.consume(QUEUE):
```
hmstepanek, line 85: lol I just fixed this in my PR too!

newrelic/api/time_trace.py
lrafeei: We probably don't.  I was thinking it would be something to have in debug mode if the customer wanted to see which log specifically was dropped.

newrelic/core/stats_engine.py
lrafeei: Oof, that's a good point.  That was a leftover from when I was turning it into the SampledSet within the SampledSet.  Removed!

hmstepanek (approved):

PR #1629, opened 2026-01-20
Add cross agent test support for DT and sampling

hmstepanek:
# Overview
Add 3 different types of cross agent tests. See READMEs for details.
Also add safety guard to sampler proxy.

### DIFF
--- a/tests/cross_agent/test_harvest_sampling_rates.py
+++ b/tests/cross_agent/test_harvest_sampling_rates.py
@@ -63,7 +63,7 @@ def parse_to_config_paths(settings, setting, config):
                 new_setting = f"{setting}._{key}"
                 v = list(value.keys())[0] if isinstance(value, dict) else value
                 settings[new_setting] = v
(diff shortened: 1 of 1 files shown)


newrelic/core/samplers/sampler_proxy.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, settings):
             # If the ratio is not defined fallback to adaptive sampler.
             if (
                 settings.distributed_tracing.sampler._root == "trace_id_ratio_based"
-                and settings.distributed_tracing.sampler.root.trace_id_ratio_based.ratio
+                and settings.distributed_tracing.sampler.root.trace_id_ratio_based.ratio is not None
```
lrafeei: Do we need to check to see if the setting is explicitly `None`?  Couldn't we just do a truthy/falsy check still?  It seems that if it is not defined, the value will be `None` and if it is defined but not a valid value, it will be assigned to "0.0" which would still be falsy.

PR #1630, opened 2026-01-21
LangChain / LangGraph Agent Instrumentation

TimPansino:
# Overview

* Overhaul LangChain instrumentation to conform to newer standards.
* Add new instrumentation for `create_agent` in LangChain (which in turn uses LangGraph).
* Overhaul testing structure for LangChain with large matrixing.

### DIFF
--- a/newrelic/common/llm_utils.py
+++ b/newrelic/common/llm_utils.py
@@ -13,18 +13,27 @@
 # limitations under the License.
 
 import itertools
+import logging
 
 from newrelic.api.transaction import current_transaction
 from newrelic.common.object_wrapper import ObjectProxy
 
+_logger = logging.getLogger(__name__)
+
 
 def _get_llm_metadata(transaction):
-    # Grab LLM-related custom attributes off of the transaction to store as metadata on LLM events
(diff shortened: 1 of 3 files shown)


TimPansino:
Just reviewed this myself and fixed or tweaked everything I found. Assigning this for full review.

newrelic/common/llm_utils.py
```diff
@@ -22,3 +27,73 @@ def _get_llm_metadata(transaction):
         llm_metadata_dict.update(llm_context_attrs)
 
     return llm_metadata_dict
+
+
+class GeneratorProxy(ObjectProxy):
```
lrafeei, line 41: Would it be possible to change the name of this to something like `LLMGeneratorProxy` (and same with the `AsyncGeneratorProxy`)?  I feel as though this might get confusing later on since there is another function in newrelic/common/async_proxy.py called `GeneratorProxy`

lrafeei: (Or is the intent to not draw attention to the fact that they are slightly different?)

PR #1632, opened 2026-01-22
Guard Azure Functions Utilization

TimPansino:
# Overview

* Add guard against crashes in Azure Functions Utilization detection.
* Add regression test.

### DIFF
--- a/newrelic/hooks/external_botocore.py
+++ b/newrelic/hooks/external_botocore.py
@@ -270,8 +270,6 @@ def create_chat_completion_message_event(
 
         if settings.ai_monitoring.record_content.enabled:
             chat_completion_message_dict["content"] = content
-        if request_timestamp:
-            chat_completion_message_dict["timestamp"] = request_timestamp
 
         chat_completion_message_dict.update(llm_metadata_dict)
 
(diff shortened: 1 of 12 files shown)


newrelic/common/utilization.py
```diff
@@ -233,21 +233,29 @@ class AzureFunctionUtilization(CommonUtilization):
...
+                faas_app_name = f"/subscriptions/{subscription_id}/resourceGroups/{resource_group_name}/providers/Microsoft.Web/sites/{azure_function_app_name}"
+                # Only send if all values are present
+                return (faas_app_name, cloud_region)
+            except Exception:
+                _logger.debug(
+                    "Unable to determine Azure Functions subscription id from WEBSITE_OWNER_NAME. %r",
+                    website_owner_name,
+                )
+
+        return None
```
umaannamalai, line 258: Was this return statement meant to be inside of the except block?

TimPansino: No, it covers the if and the except block both by being there. Also, it's just there for clarity anyway since hitting the end of the function body would do the exact same thing.

There's probably a linter rule somewhere that would tell me not to do that but I don't really care until we turn it on.

PR #1670, opened 2026-02-25
Fix package reporting implementation

lrafeei:
This PR attempts to fix an issue where some modules are not loaded in the environment section of the UI (the modules do not get passed into the `update_loaded_modules` endpoint).  It also fixes a bug where it does not reload the modules upon agent restart.

Previously the logic attempted to load as many modules as possible during a 2 second window by iterating through a plugins generator.  However, it seems that this may miss a few modules if the total loading time is longer than this. (Disclaimer: This was not the intended operation)

The new logic will upload at least one module per fast harvest cycle but as many as it can within a 0.5 second window.


newrelic/core/application.py
```diff
@@ -1284,27 +1284,23 @@ def harvest(self, shutdown=False, flexible=False):
...
-                        self.modules = []
+                    # Send environment plugin.
+                    # If a module takes less than 0.5 seconds to load,
+                    # continue to upload.  Do this until we reach 5.0
+                    # seconds or until a module has taken more than
+                    # 0.5 seconds to upload.  Then, wait for the next
+                    # harvest cycle before resuming.
+                    if self._remaining_plugins and self.configuration and self.configuration.package_reporting.enabled:
+                        start = stopwatch_start = time.time()
+                        while ((time.time() - stopwatch_start) < 0.5) and (
```
hmstepanek: I wonder-can you just simplify this to be the following and leave MAX_PACKAGE_CAPTURE_TIME_PER_SLOW_HARVEST=2:
```suggestion
                        while (time.time() - start) < MAX_PACKAGE_CAPTURE_TIME_PER_SLOW_HARVEST:
```
This way it will at least load 1 module before exiting. I realize this is slow harvest but I wonder if 5s is a bit long.

newrelic/core/application.py
```diff
@@ -1284,27 +1284,23 @@ def harvest(self, shutdown=False, flexible=False):
...
-                        # Reset the modules list every harvest cycle
-                        self.modules = []
+                    # Send environment plugin.
+                    # If a module takes less than 0.5 seconds to load,
+                    # continue to upload.  Do this until we reach 5.0
+                    # seconds or until a module has taken more than
+                    # 0.5 seconds to upload.  Then, wait for the next
+                    # harvest cycle before resuming.
+                    if self._remaining_plugins and self.configuration and self.configuration.package_reporting.enabled:
+                        start = stopwatch_start = time.time()
```
hmstepanek: This seems unlikely but I wonder if what's happening for the AI team who is testing this is they are never entering the while loop because of that less than .5 check? Are they seeing any modules reporting at all? Another way you could implement this that might be simpler and avoid this potential issue is:
```
start = time.time()
for module in self.plugins:
    self._active_session.send_loaded_modules([module])
    if time.time() - start > MAX_PACKAGE_CAPTURE_TIME_PER_SLOW_HARVEST:
        break
else:
    self._remaining_plugins = False
```

newrelic/core/application.py
```diff
@@ -1745,15 +1741,12 @@ def internal_agent_shutdown(self, restart=False):
...
-        while self.configuration and self.configuration.package_reporting.enabled and self._remaining_plugins:
-            try:
-                module_info = next(self.plugins)
-                self.modules.append(module_info)
-            except StopIteration:
-                self._remaining_plugins = False
-                if self.modules:
-                    self._active_session.send_loaded_modules(self.modules)
-                    self.modules = []
+        if self._remaining_plugins and self.configuration and self.configuration.package_reporting.enabled:
```
hmstepanek: I'm wondering if we really need the self._remaining_plugins or can we just set self.plugins = False?

hmstepanek (changes requested):
A couple thoughts here-I'm not sure exactly why it would not be reporting the module(s) again but I had one thought. Seems unlikely to be the reason but I didn't see anything else that would cause that. You could add some log statements or a metric in or something to maybe help you debug what's going on.

newrelic/core/application.py
lrafeei: Oh, like override the generator object to be a boolean when it runs out and use that as the conditional?

newrelic/core/application.py
```diff
@@ -1284,27 +1282,20 @@ def harvest(self, shutdown=False, flexible=False):
...
-                        self.modules = []
+                    # Send environment plugin.
+                    # If a module takes less than 0.5 seconds to load,
+                    # continue to upload.  Do this until we reach 5.0
+                    # seconds or until a module has taken more than
+                    # 0.5 seconds to upload.  Then, wait for the next
+                    # harvest cycle before resuming.
+                    if self.plugins and self.configuration and self.configuration.package_reporting.enabled:
+                        start = time.time()
+                        while (time.time() - start) < MAX_PACKAGE_CAPTURE_TIME_PER_SLOW_HARVEST:
```
hmstepanek:
```
logger.debug("Looping over plugins.")
while package in self.plugins:
   logger.debug(f"Recording package {package}")
   self._active_session.send_loaded_modules([next(package)])
   if (time.time() - start) < MAX_PACKAGE_CAPTURE_TIME_PER_SLOW_HARVEST
       logger.debug("Exiting plugins loop.)
       break

```

newrelic/core/application.py
```diff
@@ -49,7 +49,7 @@
 
 _logger = logging.getLogger(__name__)
 
-MAX_PACKAGE_CAPTURE_TIME_PER_SLOW_HARVEST = 2.0
+MAX_PACKAGE_CAPTURE_TIME_PER_SLOW_HARVEST = 3.0
```
hmstepanek:
```suggestion
MAX_PACKAGE_CAPTURE_TIME_PER_SLOW_HARVEST = 2.0
```

JiwaniZakir (commented):
The new harvest loop in `application.py` around line 1291 has an inverted condition that effectively breaks after sending a single package every harvest cycle. The line `if (time.time() - start) < MAX_PACKAGE_CAPTURE_TIME_PER_SLOW_HARVEST: break` causes an immediate break because the first iteration almost never exceeds 3 seconds — the `<` should be `>=` to match the original behavior of stopping once the time budget is exhausted. This is a semantic inversion of the old `while ... (elapsed < MAX)` guard.

Additionally, the new code calls `send_loaded_modules([package])` individually per package rather than batching, which introduces significantly more network overhead compared to the original approach of accumulating and sending once per cycle. If the intent was to stream packages incrementally, that trade-off should be documented.

Using `next(self.plugins, False)` with `False` as the sentinel is unconventional and could mask bugs if `package` itself is falsy for any reason — `None` is the standard sentinel for `next()`.

The addition of `self.plugins = plugins()` on restart is correct and closes a gap where a restarted agent would have had an exhausted generator, but it would benefit from a comment explaining *why* the reset is scoped to the restart path rather than being part of a broader state reset.

tests/agent_unittests/test_harvest_loop.py
hmstepanek:
```suggestion
        # span_event_data is the 4th to last endpoint called
```

newrelic/core/application.py
```diff
@@ -1515,6 +1491,18 @@ def harvest(self, shutdown=False, flexible=False):
 
                             self._active_session.send_errors(error_data)
 
+                    # Send environment plugin.
+
+                    if configuration and configuration.package_reporting.enabled:
+                        package_list = []
+                        start = time.time()
+                        while package := next(self.plugins, None):
```
hmstepanek: It doesn't matter too much but it might be slightly more readable to do this instead:
```
for package in self.plugins:
    package_list.append(package)
```

hmstepanek (changes requested):
A couple suggestions but nothing major.

tests/agent_unittests/test_harvest_loop.py
lrafeei: Whoops, good call.

hmstepanek (approved):

PR #1682, opened 2026-03-10
Llm response token counts

umaannamalai:
This PR includes changes to send token counts pulled from LLM response objects directly from the agent.


JiwaniZakir (commented):
In `create_chat_completion_message_event`, the `all_token_counts` parameter is added to the signature and checked for truthiness, but its value is never actually used — both the request message (line ~231) and response message (line ~274) unconditionally set `"token_count": 0` whenever the flag is truthy. This looks like a bug where the actual per-message token count should be extracted from `all_token_counts` rather than hardcoded to zero. If `all_token_counts` is a dict or list keyed by message index, the lookup logic is missing entirely.

Additionally, `extract_bedrock_titan_embedding_model_response` sets only `response.usage.total_tokens` (equal to `inputTextTokenCount`), while the text model response handler sets all three — `prompt_tokens`, `completion_tokens`, and `total_tokens`. The inconsistency is reasonable for embeddings (no output), but it would be worth a comment clarifying the intentional omission of `prompt_tokens` and `completion_tokens` to avoid confusion for future maintainers.

In `extract_bedrock_titan_text_model_streaming_response`, the accumulation pattern (`bedrock_attrs.get(..., 0) + value`) assumes invocation metrics may appear across multiple chunks, but `amazon-bedrock-invocationMetrics` is typically only present in the final chunk — so for most chunks `prompt_tokens` and `completion_tokens` will be `0`, meaning the accumulation adds unnecessary overhead without risk of double-counting. A guard checking whether the metrics key exists before updating would make the intent clearer.

PR #1705, opened 2026-04-06
Fix Wagtail transaction names

### LINKED ISSUE #1573: Transaction names are inconsistent using Wagtail
## Summary

We use Wagtail (https://wagtail.org/) and have noticed that the transaction names reporting in NR are inconsistent. Some pages are grouped nicely and reference their page name while others are grouped under "wagtail.views:serve".

## Desired Behavior

All the pages that we would like grouped in NR inherit from `wagtail.models.pages.Page` but it seems that only those pages that include the `RoutablePageMixin` have a specific transaction name. Ideally we would like it so any custom page (i.e. a page we override manually in our app code as opposed to pages defined purely in the CMS) gets an appropriate transaction name in NR.

## Possible Solution

I'm not familiar enough with the NR agent internals to suggest a solution but happy to work with you (and possibly Wagtail) on one.

## Additional context

We think we've found a workaround by using the `RoutablePageMixin` on every page we customise but it's not ideal.

hmstepanek:
# Overview
Fix Wagtail transaction names.

# Related Github Issue
Fixes #1573.


tests/framework_wagtail/dummy_app/models.py
```diff
@@ -0,0 +1,40 @@
...
+
+from django.db import models
+from wagtail.models import Page
+from wagtail.contrib.routable_page.models import RoutablePage, re_path
+from wagtail.fields import RichTextField
+
+class HomePage(Page):
+    body = RichTextField(blank=True)
+
+    content_panels = Page.content_panels + ["body"]
```
hmstepanek:
```suggestion
    content_panels = [*Page.content_panels, "body"]
```

tests/framework_wagtail/dummy_app/models.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class RoutablePage(RoutablePage):
+    body = RichTextField(blank=True)
+
+    content_panels = Page.content_panels + ["body"]
```
hmstepanek:
```suggestion
    content_panels = [*Page.content_panels, "body"]
```

newrelic/hooks/framework_django.py
```diff
@@ -484,7 +484,21 @@ def wrapper(wrapped, instance, args, kwargs):
...
+        # is located on the base class which results in transaction having the same
+        # name. Use the child class instead of the base class name. Set the priority=6
+        # to override the priority set in other parts of this hook file so that the
+        # more explicit name takes precedence.
+        new_name = name
+        try:
+            from wagtail.models.pages import Page
+        except:
+            Page = None
+        if instance and isinstance(instance, Page):
```
TimPansino: Can we maybe move this import to an import-hook and export it to the module level to avoid the overhead of constantly trying to import this for the vast majority of customers that don't have wagtail? We can change this `if` to check `if Page is not none and ...` and set Page to None by default.

hmstepanek: Because there was already the same import hook in the wagtail hook module I had to move the wagtail import hook to the django module to make this work. Let me know your thoughts and if you'd prefer a different approach.

tests/framework_wagtail/test_application.py
```diff
@@ -0,0 +1,114 @@
...
+    "transaction_tracer.transaction_threshold": 0.0,
+    "transaction_tracer.stack_trace_threshold": 0.0,
+    "debug.log_data_collector_payloads": True,
+    "debug.record_transaction_failure": True,
+    "debug.log_autorum_middleware": True,
+}
+
+collector_agent_registration = collector_agent_registration_fixture(
+    app_name="Python Agent Test (framework_wagtail)", default_settings=_default_settings, scope="module"
+)
```
TimPansino: Why is this stuff in here and not the conftest file?

hmstepanek: Good question. I think claude may have moved these and I didn't notice. Moved them back to conftest.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-597286

TimPansino (approved):

PR #1707, opened 2026-04-09
Guard Garbage Collection Metrics Against Re-Entrant Calls

TimPansino:
# Overview

* Add flag to prevent re-entrant calls to `_GCDataSource.record_gc()` to prevent infinite recursion on Python 3.13+.
* Improve caching of IS_PYPY in `gc_data.py` for performance.
* Add regression test that bypasses garbage collection infrastructure and directly unit tests re-entrant calls to `_GCDataSource.record_gc()`.
  * If we can find a reliable way to trigger the original issue in testing, we could add a better regression test here as well.

# Related Github Issue

Closes #1706

# Testing

1. Revert the changes made to `gc_data.py`.
2. Run the `agent_unittests` suite. Observe the tests hanging forever when simulating re-entrant invocations of `record_gc`.
3. Restore the changes to `gc_data.py`.
4. Run the `agent_unittests` suite. Observe the tests passing due to the new guard in place preventing re-entrant calls.

### DIFF
--- a/newrelic/samplers/gc_data.py
+++ b/newrelic/samplers/gc_data.py
@@ -56,8 +56,9 @@ class _GCDataSource:
         if self.__recording:
             return
 
-        # __recording flag is used to prevent re-entrant calls to record_gc. This could theoretically result in missing
-        # some GC metrics, but would more likely result in infinite recursion that crashes the entire interpreter.
+        # The self.__recording flag is used to prevent re-entrant calls to record_gc. This could theoretically result
+        # in missing some GC metrics, but trying to record the metrics instead would likely result in infinite
+        # recursion that crashes the entire interpreter.
         self.__recording = True
         try:
             if not self.enabled:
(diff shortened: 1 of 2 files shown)


newrelic/samplers/gc_data.py
```diff
@@ -49,24 +53,33 @@ def top_object_count_limit(self):
...
-        elif phase == "stop":
-            total_time = time.time() - self.start_time
-            self.gc_time_metrics.record_custom_metric(f"GC/time/{self.pid}/all", total_time)
-            for gen in range(0, 3):
-                if gen <= current_generation:
-                    self.gc_time_metrics.record_custom_metric(f"GC/time/{self.pid}/{gen}", total_time)
-                else:
-                    self.gc_time_metrics.record_custom_metric(f"GC/time/{self.pid}/{gen}", 0)
+        # __recording flag is used to prevent re-entrant calls to record_gc. This could theoretically result in missing
+        # some GC metrics, but would more likely result in infinite recursion that crashes the entire interpreter.
```
hmstepanek:
```suggestion
        # some GC metrics, but would more likely prevent  infinite recursion that crashes the entire interpreter.
```

TimPansino: Fixed

hmstepanek (approved):

PR #1712, opened 2026-04-20
Anthropic Create Instrumentation

TimPansino:
# Overview

* Add instrumentation and tests for Anthropic create.
  * This explicitly does **not** cover response streaming.

### DIFF
--- a/newrelic/hooks/mlmodel_anthropic.py
+++ b/newrelic/hooks/mlmodel_anthropic.py
@@ -1,0 +1,434 @@
+# Copyright 2010 New Relic, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


newrelic/hooks/mlmodel_anthropic.py
```diff
@@ -0,0 +1,434 @@
...
+            "request.max_tokens": request_max_tokens,
+            "vendor": "anthropic",
+            "ingest_source": "Python",
+            "duration": ft.duration * 1000,
+            "response.model": response_model,
+            "response.choices.finish_reason": stop_reason,
+            "response.number_of_messages": number_of_messages,
+            # "response.usage.total_tokens": total_tokens,
+            # "response.usage.prompt_tokens": input_tokens,
+            # "response.usage.completion_tokens": output_tokens,
```
lrafeei, line 317: I assume these are streaming specific attributes that will be re-enabled in the future?

TimPansino: They are and are also going to be after streaming is done, as they're part of token counting. I still have to work out all the details for token counting so that will likely be the last thing done for Anthropic.

lrafeei (approved):

PR #1715, opened 2026-04-20
Anthropic Streaming Instrumentation

TimPansino:
# Overview

* Initial Streaming Support for Anthropic SDK Instrumentation.
  * Intentionally does not cover response token counting as this feature is not entirely complete on `main` either.


newrelic/hooks/mlmodel_anthropic.py
```diff
@@ -21,15 +21,22 @@
...
 from newrelic.api.time_trace import get_trace_linking_metadata
 from newrelic.api.transaction import current_transaction
-from newrelic.common.object_wrapper import wrap_function_wrapper
+from newrelic.common.llm_utils import AsyncLLMStreamProxy, LLMStreamProxy
+from newrelic.common.object_wrapper import ObjectProxy, wrap_function_wrapper
 from newrelic.common.package_version_utils import get_package_version
 from newrelic.core.config import global_settings
 
 ANTHROPIC_VERSION = get_package_version("anthropic")
+EXCEPTION_HANDLING_FAILURE_LOG_MESSAGE = (
```
hmstepanek, line 30: Do you need the "\n"'s at the ends of these two error message lines?

hmstepanek (changes requested):
Looks good. I just had a question about the \n.

newrelic/hooks/mlmodel_anthropic.py
TimPansino: I removed the extra newlines, they were just copied from previous instances when we weren't using `exc_info=True`.

hmstepanek (approved):

PR #1721, opened 2026-05-02
Exclude newrelic header by default

hmstepanek:
# Overview
Exclude `newrelic` DT header by default.


TimPansino (dismissed):

TimPansino:
I think you need to edit most of these failing tests to check for a different header name other than `newrelic`.

tests/framework_grpc/test_distributed_tracing.py
```diff
@@ -122,28 +122,16 @@ def _test():
...
             nr_tracestate = list(W3CTraceState.decode(metadata["tracestate"]).values())[0]
             nr_tracestate = NrTraceState.decode(nr_tracestate, None)
-            w3c_data.update(nr_tracestate)
 
-            # Remove all trust keys
-            decoded["d"].pop("tk", None)
-            w3c_data.pop("tk")
-
-            # Round priority of newrelic header to 6 decimal places so it match tracestate.
-            decoded["d"]["pr"] = f"{decoded['d']['pr']:.6f}"
```
hmstepanek, line 142: There's not as much stuff being tested here but before it wasn't actually testing the values but testing that the newrelic header values matched the tracestate header values. We test the priority and the rest of this in the cross agent tests extensively so I think it's ok to not do that here.

TimPansino (approved):

PR #1727, opened 2026-05-08
OpenAI collect chat completion at done

lrafeei:
OpenAI reintroduced logic that closes the streaming response even if the consumer hasn't read all data. The Python Agent previously relied on reaching the stream's end before sending LLM data to the collector.  This resulted in the chat completion data not being collected at the closing of the stream.  Logic has now been added to account for this.


tests/mlmodel_langchain/test_agents.py
```diff
@@ -181,3 +248,77 @@ def _test():
...
+# using a context propagator to propagate the trace.
+@reset_core_stats_engine()
+@validate_custom_events(chat_completion_recorded_events)
+@background_task()
+def test_state_graph_with_state_invoke(chat_openai_client, exercise_graph):
+    from langgraph.graph import END, START, MessagesState, StateGraph
+
+    class GraphContext(TypedDict):
+        nr_context: TimeTrace
+
```
TimPansino, line 266: If some of this doesn't currently work without custom instrumentation, let's only test the pieces that do. I'd prefer not to test anything that doesn't work out of the box. Let's comment out the rest and mark it as not yet supported, so we can add those tests in the future.

TimPansino (approved):

TimPansino:
Merging this as is, and going to iterate on it in follow up PRs.

PR #1733, opened 2026-05-13
LangChain & LangGraph Transaction Context Propagation

TimPansino:
# Overview

* Add additional transaction context propagation to LangChain and LangGraph to support a wider range of application structures.
* Provide more thorough testing for LangGraph's StateGraph based invocation patterns.
* Add missing model attributes back to LangChain's Chain based invocation patterns.
* Reformat LangChain instrumentation to only allow keyword arguments to complex function signatures to prevent silent mistakes.


newrelic/hooks/mlmodel_langchain.py
```diff
@@ -359,7 +359,25 @@ def wrap_ContextThreadPoolExecutor_submit(wrapped, instance, args, kwargs):
...
+        executor_or_config, func, args, kwargs = bind_run_in_executor(*args, **kwargs)
+    except Exception:
+        return await wrapped(*args, **kwargs)
+
+    func = context_wrapper(func, trace=trace, strict=True)
+    return await wrapped(executor_or_config, func, *args, **kwargs)
+
+
+def _create_error_vectorstore_events(*, transaction, search_id, args, kwargs, linking_metadata, wrapped):
     settings = transaction.settings if transaction.settings is not None else global_settings()
```
hmstepanek, line 381:
```suggestion
    settings = transaction.settings or global_settings()
```

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -965,13 +1015,39 @@ def add_nr_completion_id(run_args, completion_id):
...
+def _get_chain_request_model(instance):
+    # A best effort attempt to pull the request model from the chain or any of
+    # its steps for better observability. This is not guaranteed to work in all
+    # cases as it depends on how the chain and steps are implemented, and it can
+    # only pull the first model it finds. The request model is not a guaranteed
+    # attribute on chains or steps, but some implementations may have it.
+    try:
+        llm = getattr(instance, "llm", None)
+        if llm is not None:
+            for attr in ("model_name", "model"):
```
hmstepanek: Wondering if this might be a little more efficient?
```suggestion
            name = getattr(llm, "model_name", None) or getattr(llm, "model", None)
            if isinstance(name, str):
                return name
```

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -985,6 +1061,8 @@ def _create_error_chain_run_events(transaction, instance, run_args, completion_i
             "virtual_llm": True,
             "request_id": run_id,
             "duration": duration,
+            "request.model": model,
+            "response.model": model,
```
hmstepanek: Should the response model be populated in an error situation?

newrelic/hooks/mlmodel_langchain.py
```diff
@@ -1054,6 +1134,8 @@ def _create_successful_chain_run_events(
...
```
hmstepanek: Traditionally the request model and response model are different values/come from different sources. Does it even make sense for this to have a model given it's a layer around a llm call/virtual call and the llm call typically has the model not this wrapper code?

hmstepanek: I could see if it's requesting a specific model, reporting the model (that appears to be what's happening here) but it doesn't seem to be responding with one so it seems like it would be more accurate to leave this blank rather than use the requested model as the response model? Unfortunately the spec seems kinda vague about this and it appears we inconsistently do this today. Older mlmodel hooks we seem to leave it off but newer ones we seem to have started adding it for some reason. I see this particularly when it comes to the error events.

hmstepanek (changes requested):
My main concern is the inconsistency in behavior here around attaching model data. And the spec provides no explicit clarity on this though I would interpret it to mean we should only be attaching the request.model when we have a requested model and a response.model when the response contains a model. Some of the code here and code in more recently implemented mlmodel hook files seems to be at odds with this. I would like to see some explicit direction on the behavior here rather than what we have now which seems a bit adhoc. 

Lalleh said she wanted to test this against her test app too to make sure this fixes the issue she was seeing so I've added her as a reviewer for this as well.

newrelic/hooks/mlmodel_langchain.py
TimPansino: I don't disagree with this, but this construct is repeated >30 times in the codebase and we should probably only replace it all at once.

newrelic/hooks/mlmodel_langchain.py
TimPansino: As far as I can tell, we should report the `response.model` if we have it. If we get a response with the model but there's an error somewhere else in the chain, we can end up with both of these attrs still.

newrelic/hooks/mlmodel_langchain.py
TimPansino: I think this is worth discussing internally with leadership. These models used to be reported, but we removed them because we thought it didn't make sense. Now we've gotten a request to add them back on both request and response attributes, even though we don't necessarily have a model in the response.

We should get a definite plan here for this section.

TimPansino: I've pulled out the model attribute changes from this PR, it's now entirely about the context propagation component. We'll separate that discussion to a new PR.

newrelic/hooks/mlmodel_langchain.py
hmstepanek: I generally agree with that. ^

lrafeei (approved):

hmstepanek (approved):

PR #1737, opened 2026-05-20
Fix inconsistencies around harvest limit config settings

hmstepanek:
# Overview
* Fix serverside ml_event harvest limits ordering issue.
   * Previously the ml_event would not be added to the allow list which would reject ml_events from being recorded.
* Remove logging an error message when we attempt to delete a setting on the settings object that doesn't exist.
* Only report non-deprecated harvest limit setting names in log.


newrelic/config.py
```diff
@@ -355,6 +355,103 @@ def _process_setting(section, option, getter, mapper):
...
+        # value_config takes precendence over value_stored
+        try:
+            value_stored = getattr(_config_object, getter)(section, option_stored)
+        except configparser.NoOptionError:
+            value_stored = None
+        try:
+            value_config = getattr(_config_object, getter)(section, option_config)
+        except configparser.NoOptionError:
+            value_config = None
+        value = value_config or value_stored
```
TimPansino: Confirmed bug from hand testing: Setting a literal value of 0 for the new setting but a non-zero value for the older settings causes the older setting to not only win, but also somehow bypasses the warning log message. A setting of 0 should be disabled for `event_harvest_config` settings, so this is a definite bug.

```suggestion
        # Use an explicit "is not None" check rather than "or"
        # so a legitimate value of 0 is not dropped.
        value = value_config if value_config is not None else value_stored
```

newrelic/config.py
```diff
@@ -355,6 +355,103 @@ def _process_setting(section, option, getter, mapper):
...
+        if value_stored and value_config is None:
+            _logger.info(
+                "Deprecated setting found: %r. Please use new setting: %r. Applying value of deprecated setting %r to %r.",
+                value_stored,
+                value_config,
+                value_stored,
+                value_config,
+            )
+        elif value_stored and value_config:
+            _logger.info("Ignoring deprecated setting: %r. Using new setting: %r.", value_stored, value_config)
```
TimPansino: This section is using the incorrect values to log here. Previously it used settings names, but this is set up to use the values from those settings instead. 

This is also subject to the previous comment of respecting a setting of 0 and not just evaluating truthiness. Not checking `is not None` is causing the log messages to be silently skipped when both are set, but one setting is set to a literal 0. This should definitely still issue a log message, so all these checks need to be using `None` comparisons.

```
2026-05-29 15:28:11,271 (27434/MainThread) newrelic.config INFO - Deprecated setting found: 100. Please use new setting: None. Applying value of deprecated setting 100 to None.
2026-05-29 15:28:11,271 (27434/MainThread) newrelic.config INFO - Deprecated setting found: 100. Please use new setting: None. Applying value of deprecated setting 100 to None.
2026-05-29 15:28:11,271 (27434/MainThread) newrelic.config INFO - Deprecated setting found: 100. Please use new setting: None. Applying value of deprecated setting 100 to None.
2026-05-29 15:28:11,271 (27434/MainThread) newrelic.config INFO - Deprecated setting found: 100. Please use new setting: None. Applying value of deprecated setting 100 to None.
2026-05-29 15:28:11,272 (27434/MainThread) newrelic.config INFO - Deprecated setting found: 100. Please use new setting: None. Applying value of deprecated setting 100 to None.
```

```suggestion
        if value_stored is not None and value_config is None:
            _logger.info(
                "Deprecated setting found: %r. Please use new setting: %r. Applying value of deprecated setting %r to %r.",
                option_stored,
                option_config,
                option_stored,
                option_config,
            )
        elif value_stored is not None and value_config is not None:
            _logger.info("Ignoring deprecated setting: %r. Using new setting: %r.", option_stored, option_config)
```

TimPansino (changes requested):
There are 3 issues in this PR that were surfaced by hand testing. 

1. Log messages now incorrectly use the setting *value* instead of setting *names*.

The changes incorrectly assume truthiness is a valid way of comparing settings, when a literal 0 is a valid setting with a real meaning. This causes the following 2 issues.
 
2. Precedence is being ignored where the deprecated setting can win over the new setting if the new setting is set to 0.
3. Log messages detailing which settings are in use are incorrect if either is set to 0.

As part of the review, I added 2 test cases to the existing tests to validate that the value of 0 is being respected over both the default settings, and the deprecated settings. I pushed that in a new commit.

TimPansino:
Another issue worth considering, we seem to have missed the setting for ml_event_data. We still use the old naming convention entirely for that setting. Would it be worth changing `event_harvest_config.harvest_limits.ml_event_data` -> `ml_events.max_samples_stored` for consistency?

That would require moving a fair bit of code around to fix, so maybe it would be best for a follow up. The other problem with this is noise for customers who have it set, but this setting is incredibly rare in the wild.

newrelic/core/config.py
```diff
@@ -1666,9 +1666,25 @@ def apply_server_side_settings(server_side_config=None, settings=_settings):
...
-    harvest_limits = event_harvest_config.get("harvest_limits", ())
+    harvest_limits = event_harvest_config.get("harvest_limits", {})
+
+    # Override this setting here so that the allowlist that is pulled from the
+    # the harvest_limits includes the ml_event harvest limit.
+    # Since the server does not override this setting as it's an OTLP setting,
+    # we must override it here manually by converting it into a per harvest cycle
+    # value.
+    # override ml_events / (60s/5s) harvest
+    harvest_limits["ml_event_data"] = settings_snapshot.event_harvest_config.harvest_limits.ml_event_data / 12
```
TimPansino: This was a pre-existing bug, but this should be floor division and not floating point. Previously we would see numbers with floating point values like this reported in `agent_settings`.

```
'harvest_limits': {
  'analytic_event_data': 16,
  'custom_event_data': 16,
  'span_event_data': 199,
  'error_event_data': 8,
  'ml_event_data': 16.583333333333332,
  'log_event_data': 16
}
```

```suggestion
    harvest_limits["ml_event_data"] = settings_snapshot.event_harvest_config.harvest_limits.ml_event_data // 12
```

TimPansino:
A note on the PR description:

> Previously the ml_event would not be added to the allow list which would reject ml_events from being recorded.

The ml_event_data was never expected to be in the allowlist before, and we don't ever check the presence of it in the allowlist. Adding it there only makes it show up in the UI under the agent settings now, which is less confusing.

It's still not checked to this day, as the `reset_ml_events` function is actually gated by `custom_event_data` in the allow list, as this was originally an extension onto custom events.

The customer who originally reported the missing method from the allowlist was upset about their LLM data not being visible in the UI. That data does not have any relevance to the ml_event_data settings, as that's exclusively used for SKLearn instrumentation and ML happening on host. (That customer's data was also being reported in the first place as we found the historical data. They were having an issue with the UI tab not appearing which was later fixed without our intervention.)

newrelic/config.py
```diff
@@ -355,6 +355,103 @@ def _process_setting(section, option, getter, mapper):
...
+                break
+            target = getattr(target, fields[0])
+            fields = fields[1].split(".", 1)
+
+        # Cache the configuration so can be dumped out to
+        # log file when whole main configuration has been
+        # processed. This ensures that the log file and log
+        # level entries have been set.
+
+        _cache_object.append((option_stored, value))
```
TimPansino: I'm not sure this line is correct, or else this whole function is confused. This is putting the deprecated setting name in my logs instead of the current setting.

```
2026-05-29 16:34:04,075 (48081/MainThread) newrelic.config DEBUG - agent config event_harvest_config.harvest_limits.analytic_event_data = 299
```

The problem I'm seeing is that this was meant to be displaying the "new" settings names (`max_samples_stored`) that we prefer from an end user perspective, but the final output of this function is supposed to be storing it in the `event_harvest_config` format that the collector still expects.

I'm questioning the design choice here anyway, as the original intent of the settings printout was to list every user setting that was parsed and applied. If the user is using a deprecated setting, that translation historically happened after we had already printed out the settings dump.

My suggestion would be to set the name to whichever option was originally read from the config and respected.

```suggestion
        option_used = option_config if value_config is not None else option_stored
        _cache_object.append((option_used, value))
```

This is a really weird issue though and that may not be the correct design.

hmstepanek: It was intentional to use the old deprecated name as we take the user's value and we store it into the old deprecated setting name. Similarly the server side config uses the old name. I understand and agree with your point though-I'll change it to log it as the old name in the cache even though we technically store it into the deprecated name.

TimPansino: Ok so to make sure we're on the same page here, the behavior I'm seeing now is that no matter which setting name you use - the deprecated or the non-deprecated one - we're going to log it as the non-deprecated setting name when we dump our debug logs. Is that correct?

I'm not 100% sold on that but it's just for the debug logs so it's not a huge deal. But want to be certain the behavior is what we were intending.

TimPansino (approved):
Approving, with the note that the `ml_event_data` setting name is out of sync with the others but that's out of scope for this PR. We'll likely want to change that in a future major version.

newrelic/config.py
TimPansino: Resolving this as that is the intended behavior

PR #1741, opened 2026-05-27
chore: Initial work to add agent config schema

mvicknr:
# Overview
Adds a script to automatically generate and update a configuration schema for the agent. This PR has the initial schema generated from the script.

The script will automatically run on local build and on pushes. It will determine if there are config changes and if so, it will update the schema. On local, the dev should include these changes in their commit. On push, if the changes aren't already made in the schema, it will make the changes and add them to the PR. 

The schema version will be updated as part of the regular agent release. - This will need to happen after the release tag has been generated but before it has been applied.

See the readme in the schemaGeneration directory to understand how the schema is generated.

# Testing
Github Actions when a pull request is made. More details on running the tests locally can be found in our
[testing guidelines](https://github.com/newrelic/newrelic-python-agent/blob/main/CONTRIBUTING.md#testing-guidelines),
exercise your changes.


mvicknr:
this commit shows that the schema generation on push works: [chore: regenerate Fleet Control config schema](https://github.com/newrelic/newrelic-python-agent/pull/1741/commits/0cff9756d60565508361b66122d930da731d25d0)

.fleetControl/schemas/config.json
```diff
@@ -0,0 +1,1114 @@
...
+      "items": {
+        "type": "string"
+      },
+      "default": []
+    },
+    "log_file": {
+      "type": "string",
+      "description": "Sets the name of a file to log agent messages to. Whatever you set this to, you must ensure that the permissions for the containing directory and the file itself are correct, and that the user that your web application runs as can write out to the file. If not able to out a log file, it is also possible to say \"stderr\" and output to standard error output. This would normally result in output appearing in your web server log. It can also be set using the NEW_RELIC_LOG environment variable."
+    },
+    "log_level": {
```
mvicknr, line 664: other agents are not using an enum here because enum forces case sensitivity - should I update this to a string for you as well?

umaannamalai (approved):

PR #1748, opened 2026-06-01
LangChain Chain Model Reporting

TimPansino:
# Overview

* Add a new strategy back to LangChain for capturing model attributes on chain invocations.


hmstepanek (commented):
I'm trying to wrap my head around why we need this callback pattern but I really don't understand. Can't we just call the function where we need it after we get the response back instead of all this passing around of a handler callback function and class?

hmstepanek (approved):

PR #1749, opened 2026-06-01
Anthropic Token Counting

TimPansino:
# Overview

* Add token counting to Anthropic SDK for all LLM invocations.

### DIFF
--- a/newrelic/hooks/mlmodel_anthropic.py
+++ b/newrelic/hooks/mlmodel_anthropic.py
@@ -425,6 +425,8 @@ def _record_completion_error(*, transaction, linking_metadata, completion_id, kw
             request_model=request_model,
             llm_metadata=llm_metadata,
             response_content=None,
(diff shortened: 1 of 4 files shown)


newrelic/hooks/mlmodel_anthropic.py
```diff
@@ -455,10 +458,39 @@ def _record_completion_success(
...
+                content
+                for msg in messages
+                if (
+                    content := _extract_message_content(
+                        msg.get("content") if isinstance(msg, dict) else getattr(msg, "content", None)
+                    )
+                )
+            )
+            if input_message_content:
+                response_prompt_tokens = token_count_callback(request_model, input_message_content)
```
umaannamalai: I wonder if we should put the token count callback invocations inside of a specific try-except block so we can log out an error message with details if the customer's callback function failed.

TimPansino: I've added an error message for this.

umaannamalai (approved):

PR #1750, opened 2026-06-02
Gemini Token Counting

TimPansino:
# Overview

* Add token counting to Google GenAI SDK for all LLM invocations.


newrelic/hooks/mlmodel_gemini.py
```diff
@@ -197,20 +197,24 @@ def _record_embedding_success(*, transaction, embedding_id, linking_metadata, kw
...
-                settings.ai_monitoring.llm_token_count_callback(request_model, embedding_content)
-                if settings.ai_monitoring.llm_token_count_callback
-                else None
-            ),
             "request.model": request_model,
             "duration": ft.duration * 1000,
             "vendor": "gemini",
             "ingest_source": "Python",
         }
+        if embedding_token_count:
```
hmstepanek: You don't have to move this outside of the dictionary as attributes with value of None are dropped as part of our attribute filtering.

newrelic/hooks/mlmodel_gemini.py
```diff
@@ -492,15 +496,13 @@ def _record_generation_error(*, transaction, linking_metadata, completion_id, kw
...
-    if generation_config:
-        request_temperature = getattr(generation_config, "temperature", None)
-        request_max_tokens = getattr(generation_config, "max_output_tokens", None)
-    else:
-        request_temperature = None
-        request_max_tokens = None
+    # Extract data from generation config object
+    request_temperature, request_max_tokens = _extract_generation_config(kwargs)
 
+    # Prepare error attributes
```
hmstepanek: I don't really feel like this comment is adding anything.
```suggestion
```

newrelic/hooks/mlmodel_gemini.py
```diff
@@ -540,17 +542,19 @@ def _record_generation_error(*, transaction, linking_metadata, completion_id, kw
...
+            chat_completion_id=completion_id,
+            span_id=span_id,
+            trace_id=trace_id,
             # Passing the request model as the response model here since we do not have access to a response model
-            request_model,
-            request_model,
-            llm_metadata,
-            output_message_list,
-            request_timestamp,
+            response_model=request_model,
```
hmstepanek, line 548: It looks like we dropped the request model here and only have the response model with the request model value-is that intentional?

newrelic/hooks/mlmodel_gemini.py
```diff
@@ -644,13 +651,37 @@ def _record_generation_success(
...
+            if output_message_content:
+                response_completion_tokens = token_count_callback(response_model, output_message_content)
+
+        # Prefer the sum of individual counts as the total whenever both are available.
+        # This ensures consistency in the event that the token counting callback has reported
+        # different values for prompt or completion tokens.
+        if response_prompt_tokens and response_completion_tokens:
+            response_total_tokens = response_prompt_tokens + response_completion_tokens
+
+        all_token_counts = bool(response_prompt_tokens and response_completion_tokens and response_total_tokens)
```
hmstepanek, line 677: Not sure the casting is really necessary here.
```suggestion
        all_token_counts = response_prompt_tokens and response_completion_tokens and response_total_tokens
```

newrelic/hooks/mlmodel_gemini.py
```diff
@@ -672,26 +703,57 @@ def _record_generation_success(
             "time_to_first_token": time_to_first_token,
         }
 
+        if all_token_counts:
```
hmstepanek, line 701: Just double checking the logic here-if we have only response completion tokens we don't want to add those at all? Its a attach all or nothing situation?

hmstepanek: If this is correct then I think couldn't we just use `if response_total_tokens` instead of creating a new variable?

newrelic/hooks/mlmodel_gemini.py
```diff
@@ -672,26 +703,57 @@ def _record_generation_success(
...
+def _parse_input_message(input_message):
+    # The input_message will be a string if generate_content was called directly. In this case, we don't have
+    # access to the role, so we default to user since this was an input message
+    if isinstance(input_message, str):
+        return input_message, "user"
+    # The input_message will be a Google Content type if send_message was called, so we parse out the message
+    # text and role (which should be "user")
+    elif isinstance(input_message, google.genai.types.Content):
+        return input_message.parts[0].text, input_message.role
+    else:
```
hmstepanek, line 736: Since we have this else clause if input_message was None it would fall into the else clause in here which means all the places where we are checking if input_message is None before calling this function are unnecessary.

hmstepanek (changes requested):
A couple minor things-otherwise looks good.

newrelic/hooks/mlmodel_gemini.py
TimPansino: I prefer the bool casting so it's clear this is a behavioral flag and not a count.

newrelic/hooks/mlmodel_gemini.py
TimPansino: This is actually more subtle, a value of 0 should also be dropped for token counts. I can take out the if statement with an `or` though.

newrelic/hooks/mlmodel_gemini.py
TimPansino: Yes, since it was unused. The request_model is not attached to the message events, just the summary.

We've been attaching the request model AS the response model attribute though since the real response model was unavailable.

newrelic/hooks/mlmodel_gemini.py
TimPansino: 1. Yes, the token counts should only be attached from the usage metadata if we have access to both prompt and completion counts. If we only have 1 and not both, we don't attach anything. This is done so the pipeline's tokenizer can estimate token counts if the info isn't fully available, but partial availability can produce nonsensical results where the numbers don't add up to the sum.
2. If we pull the total tokens from the usage metadata, it's possible we pull total tokens but not the prompt tokens and we end up with a partial count attachment. It's generally safer to have a separate and explicit flag that we've retrieved all relevant token counts.

newrelic/hooks/mlmodel_gemini.py
hmstepanek: Oooh ok that makes sense.

newrelic/hooks/mlmodel_gemini.py
hmstepanek: Ah I see ok-sounds good.

hmstepanek (approved):

PR #1767, opened 2026-06-16
Google ADK Agents Instrumentation

TimPansino:
# Overview

* Add instrumentation for Google ADK Agents

### DIFF
--- a/newrelic/hooks/mlmodel_googleadk.py
+++ b/newrelic/hooks/mlmodel_googleadk.py
@@ -1,0 +1,152 @@
+# Copyright 2010 New Relic, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import logging
+import sys
+import uuid
+
+from newrelic.api.function_trace import FunctionTrace
+from newrelic.api.time_trace import get_trace_linking_metadata
+from newrelic.api.transaction import current_transaction
+from newrelic.common.llm_utils import AsyncLLMStreamProxy, _get_llm_metadata
+from newrelic.common.object_names import callable_name
+from newrelic.common.object_wrapper import wrap_function_wrapper
+from newrelic.common.package_version_utils import get_package_version
+from newrelic.core.config import global_settings
+
+_logger = logging.getLogger(__name__)
+GOOGLEADK_VERSION = get_package_version("google-adk")
+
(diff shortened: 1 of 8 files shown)


newrelic/hooks/mlmodel_googleadk.py
```diff
@@ -0,0 +1,152 @@
...
+        return wrapped(*args, **kwargs)
+
+    settings = transaction.settings or global_settings()
+    if not settings.ai_monitoring.enabled:
+        return wrapped(*args, **kwargs)
+
+    transaction.add_ml_model_info("GoogleADK", GOOGLEADK_VERSION)
+    transaction._add_agent_attribute("llm", True)
+
+    func_name = callable_name(wrapped)
```
umaannamalai: In our agent spec, it's not a requirement that we include the full callable name in the function trace name. Since this one is pretty long, I'm wondering if we're better off just using the name of the wrapped function. The goal with using the full name was more so to disambiguate the function name so the customer is clear what function was wrapped.

newrelic/hooks/mlmodel_googleadk.py
```diff
@@ -0,0 +1,152 @@
...
+    agent_name = getattr(instance, "name", "agent")
+    function_trace_name = f"{func_name}/{agent_name}"
+    agentic_subcomponent_data = {"type": "APM-AI_AGENT", "name": agent_name}
```
umaannamalai, line 49: Going forward, we will ideally only want to send the subcomponent attr if we have a local agent executing (rather than an invocation by a local or remote agent). Is there a way in the ADK repo to distinguish between remote and local agents and executions vs invocations?

newrelic/hooks/mlmodel_googleadk.py
```diff
@@ -0,0 +1,152 @@
...
+        agent_event_dict.update(_get_llm_metadata(transaction))
+    except Exception:
+        _logger.warning(AGENT_EVENT_FAILURE_LOG_MESSAGE, exc_info=True)
+        agent_event_dict = {}
+
+    return agent_event_dict
+
+
+def instrument_googleadk_agents_llm_agent(module):
+    if hasattr(module, "LlmAgent") and hasattr(module.LlmAgent, "_run_async_impl"):
```
umaannamalai, line 149: Is there any risk of this function getting called multiple times for a single agent execution (ex: re-entrance)? I'm thinking about the case that sub-agents are used. Maybe a test case for sub-agents could help?

newrelic/hooks/mlmodel_googleadk.py
TimPansino: Remote agents are handled by a separate class entirely, and it looks like there's a `parent_agent` field on all agents that can determine if this is a sub agent and not a top level agent.

newrelic/hooks/mlmodel_googleadk.py
TimPansino: (This comment is showing up weirdly on GitHub so I'm assuming this is about the agent instrumentation.)

This shouldn't be an issue, the attrs are stored on the proxy, and there should be separate proxies for each call. I'd like to add multi-agent tests with sub agents, but I'm trying to get this all reviewed and merged incrementally. I was planning on doing that once we have Tools + Agents instrumentation merged.

umaannamalai (approved):

PR #1768, opened 2026-06-16
Google ADK Tool Instrumentation

TimPansino:
# Overview

* Add instrumentation for Google ADK Tools

### DIFF
--- a/newrelic/hooks/mlmodel_googleadk.py
+++ b/newrelic/hooks/mlmodel_googleadk.py
@@ -20,7 +20,7 @@ import uuid
 from newrelic.api.function_trace import FunctionTrace
 from newrelic.api.time_trace import get_trace_linking_metadata
 from newrelic.api.transaction import current_transaction
-from newrelic.common.llm_utils import _get_llm_metadata
+from newrelic.common.llm_utils import AsyncLLMStreamProxy, _get_llm_metadata
 from newrelic.common.object_names import callable_name
 from newrelic.common.object_wrapper import wrap_function_wrapper
 from newrelic.common.package_version_utils import get_package_version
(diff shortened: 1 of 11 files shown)


newrelic/hooks/mlmodel_googleadk.py
```diff
@@ -0,0 +1,167 @@
...
+            "agent_name": agent_name,
+            "vendor": "google_adk",
+            "ingest_source": "Python",
+        }
+        if error:
+            tool_event_dict["error"] = True
+
+        if settings.ai_monitoring.record_content.enabled:
+            tool_event_dict["input"] = str(tool_input) if tool_input else None
+            tool_event_dict["output"] = str(tool_output) if tool_output is not None else None
```
umaannamalai: Can this be shortened to remove the `is not None`?

newrelic/hooks/mlmodel_googleadk.py
```diff
@@ -0,0 +1,167 @@
...
+            tool_id=tool_id,
+            run_id=run_id,
+            tool_input=tool_input,
+            tool_output=response_dict,
+            agent_name=agent_name,
+            error=False,
+            transaction=transaction,
+            linking_metadata=linking_metadata,
+        )
+        tool_event_dict["duration"] = ft.duration * 1000
```
umaannamalai: If we hit an error in `_construct_base_tool_event_dict()`, then `tool_event_dict` will be an empty dict. Recording the duration alone in an empty dict that doesn't have the tool name, ID, or any other info may not be useful. We may want to verify that `tool_event_dict` came back non-empty before adding the duration to it.

newrelic/hooks/mlmodel_googleadk.py
```diff
@@ -0,0 +1,167 @@
...
+        if function_call is not None:
+            tool_name = getattr(function_call, "name", "tool") or "tool"
+            run_id = getattr(function_call, "id", "") or ""
+            tool_input = getattr(function_call, "args", None)
+        if agent is not None:
+            agent_name = getattr(agent, "name", "agent") or "agent"
+    except Exception:
+        _logger.warning(TOOL_EXTRACTOR_FAILURE_LOG_MESSAGE, exc_info=True)
+
+    func_name = callable_name(wrapped)
```
umaannamalai: In our agent spec, it's not a requirement that we include the full callable name in the function trace name. Since this one is pretty long, I'm wondering if we're better off just using the name of the wrapped function. The goal with using the full name was more so to disambiguate the function name so the customer is clear what function was wrapped.

newrelic/hooks/mlmodel_googleadk.py
```diff
@@ -0,0 +1,167 @@
...
+    function_trace_name = f"{func_name}/{tool_name}"
+    agentic_subcomponent_data = {"type": "APM-AI_TOOL", "name": tool_name}
+
+    ft = FunctionTrace(name=function_trace_name, group="Llm/tool/GoogleADK")
+    ft.__enter__()
+    ft._add_agent_attribute("subcomponent", json.dumps(agentic_subcomponent_data))
```
umaannamalai: Going forward, we will ideally only want to send the subcomponent attr if we have a local  tool executing. Is there a way in the ADK repo to distinguish between remote and local tools and executions vs invocations?

TimPansino: Yes to remote vs local, it's distinguishable by class. I've added a gate for this to only add subcomponent to local tools.

I'm not sure what executions vs invocations means in the context of tools here though.

newrelic/hooks/mlmodel_googleadk.py
```diff
@@ -145,6 +147,147 @@ def _construct_base_agent_event_dict(agent_name, agent_id, transaction, linking_
...
+    transaction = current_transaction()
+    if not transaction:
+        return await wrapped(*args, **kwargs)
+
+    settings = transaction.settings or global_settings()
+    if not settings.ai_monitoring.enabled:
+        return await wrapped(*args, **kwargs)
+
+    transaction.add_ml_model_info("GoogleADK", GOOGLEADK_VERSION)
+    transaction._add_agent_attribute("llm", True)
```
TimPansino, line 160: @umaannamalai Was this part necessary for tools? I'm pretty sure this is the result of a merge conflict but not certain.

umaannamalai: We actually have been emitting this metric in the tools instrumentation also for other agentic frameworks. The idea was that we'd still get the AI Responses tab to appear if the tool instrumentation was hit independent of the agents instrumentation.

umaannamalai (approved):

PR #1779, opened 2026-07-01
Google ADK LoopAgent Support

TimPansino:
# Overview

* Add instrumentation and tests for Google ADK's LoopAgent class.

### DIFF
--- a/tests/mlmodel_googleadk/_test_loop_agent.py
+++ b/tests/mlmodel_googleadk/_test_loop_agent.py
@@ -1,0 +1,104 @@
+# Copyright 2010 New Relic, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


tests/mlmodel_googleadk/_test_loop_agent.py
```diff
@@ -0,0 +1,104 @@
...
+            "span_id": None,
+            "trace_id": "trace-id",
+            "vendor": "google_adk",
+            "ingest_source": "Python",
+            "duration": None,
+        },
+    )
+]
+
+# TODO: validate_custom_events requires exactly one match per event, and
```
umaannamalai, line 64: Do we need this TODO here still? Validating the number of events and the contents of the child event should be sufficient.

TimPansino: I was going to leave it in there since we aren't able to validate the shape of the child `LlmAgent` event. I was hoping we'd be able to fix this at some point in the future by making that validator better, or making a more focused LLM validator but that's a not right now kind of problem.

umaannamalai (approved):

PR #1793, opened 2026-07-15
Add support for AIM SSC.

umaannamalai:
Support server side configuration of AIM settings in NR1

Precedence:
If ai_monitoring.enabled is present in the connect response, this value MUST take precedence to override the ai_monitoring.enabled value and fallback on using the value from collect_ai. For example, if collect_ai is False, and ai_monitoring.enabled locally is True, the agent would ultimately disable AI Monitoring. If both collect_ai is False and ai_monitoring.enabled is True in the connect response, ai_monitoring.enabled would ultimately enable AI Monitoring. If HSM is enabled, the appropriate AI monitoring settings SHOULD be set appropriately in the connect response. As a safeguard however, agents SHOULD also enforce HSM mode on AI settings they receive from the connect response.


newrelic/core/config.py
```diff
@@ -1653,6 +1653,16 @@ def apply_server_side_settings(server_side_config=None, settings=_settings):
         if value == "apdex_f":
             agent_config["transaction_tracer.transaction_threshold"] = None
 
+    # collect_ai is the account-level toggle to disable AIM.
+    # The value of ai_monitoring.enabled sent from SSC always takes
+    # precedence over the value of collect_ai.
+    # Apply collect_ai first so that when agent_config settings from SSC
+    # are applied, they will override collect_ai if needed.
+    collect_ai = server_side_config.get("collect_ai", None)
```
hmstepanek: Shouldn't this be `agent_config` instead of `server_side_config`?

hmstepanek: Do we need to do a pop here instead of a get to ensure it doesn't loop over this same setting and override the value out of order in the code below?

hmstepanek (changes requested):

newrelic/core/config.py
umaannamalai: `collect_ai` actually isn't sent up in the `agent_config` block. It is its own key within the connect response payload.  

I can add a pop for `collect_ai`. The call to `apply_config_setting(settings_snapshot, name, value)` in the server_side_config loop shouldn't actually affect any logic with `ai_monitoring.enabled` since it's looking for the name `collect_ai`. However, it isn't necessary to store the value of `collect_ai` in the `settings_snapshot` if we've already done the override on line 1663 so I can get rid of it.

hmstepanek (approved):

PR #1800, opened 2026-07-22
Fix Missing ChatCompletion events in LangChain when using MCP

lrafeei:
This PR resolves the issue of missing ChatCompletion events when using LangChain/LangGraph with MCP tools, specifically when the response format is not the default JSON.


Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-597287

tests/mlmodel_gemini/langchain_integration/conftest.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+# Unlike the rest of the gemini test suite, the output of these tests also
+# contains a randomly generated ID.  This ID is baked into the 'thoughtSignature'
+# which has a proprietary encryption.
+# To ensure that VCR_MATCH_ON is only affected in this specific conftest, we
+# must override the fixture itself.
+@pytest.fixture(autouse=True)
+def vcr_match_on():
+    return [matcher for matcher in VCR_MATCH_ON if matcher != "body"]
```
TimPansino: Are you sure that's necessary? Not matching on the body contents at all seems like a very large gap. The other gemini test suite has thoughtSignatures as well, but it comes from a response initially. The recorded responses will always be the same, so the thoughSignature should be consistent across runs.

tests/mlmodel_gemini/langchain_integration/conftest.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+# Initialize MCP Client and load tools
+@pytest.fixture(scope="session")
+def mcp_client():
+    mcp_server_location = (
+        "langchain_integration/mcp_server.py"
+        if os.getenv("GITHUB_ACTIONS")
+        else "tests/mlmodel_gemini/langchain_integration/mcp_server.py"
+    )
```
TimPansino: This seems like a bug that could be fixed by adding `__init__.py` files to the test suite folders to make sure they're importable.

tests/mlmodel_gemini/langchain_integration/conftest.py
lrafeei: In this case, the output also contains a randomly generated ID in its dictionary and the thoughtSignature encrypts that as well, so it is a different signature every time

tests/mlmodel_langchain/gemini_integration/conftest.py
```diff
@@ -0,0 +1,194 @@
...
+                                message.id = f"lc_run--ai-message-id-{step}"
+                            except AttributeError:
+                                # ToolMessage
+                                if isinstance(message.content, list):
+                                    message.content[-1]["id"] = f"lc_tool-message-content-id-{step}"
+                                else:
+                                    message.tool_call_id = f"tool-id-{step}"
+                                message.id = f"tool-message-id-{step}"
+                        yield {"messages": update["messages"]}
+                step += 1
```
TimPansino: What is all this supposed to be doing? Why are we editing the content of the returned messages? There aren't any assertions in here so I'm not following what the point of this is.

tox.ini
```diff
@@ -188,7 +188,7 @@ envlist =
     python-mlmodel_googleadk-{py310,py311,py312,py313,py314}-googleadklatest,
     python-mlmodel_strands-{py310,py311,py312,py313}-strandslatest,
     python-mlmodel_anthropic-{py39,py310,py311,py312,py313,py314,py314t},
-    python-mlmodel_gemini-{py39,py310,py311,py312,py313,py314,py314t},
+    python-mlmodel_gemini-{py310,py311,py312,py313,py314},
```
TimPansino: I don't think it's acceptable to reduce our coverage of Gemini to include LangChain in the tests. Either these tests need to go into the LangChain test suite and install gemini there, or there needs to be some more tox installation magic and `importorskip` calls to limit the langchain tests to only install and run when supported.

tox.ini
```diff
@@ -466,7 +466,11 @@ deps =
     mlmodel_autogen: mcp
     mlmodel_anthropic: anthropic
     mlmodel_gemini: pytest-recording
-    mlmodel_gemini: google-genai
+    mlmodel_gemini: google-genai<2.14
```
TimPansino: Does langchain require this pin? I don't like the idea of pinning this and not having visibility into the latest `google-genai` package versions. If this is required, this will have to either be moved into the langchain tests or made into a set of separate tox envs.

newrelic/hooks/mlmodel_gemini.py
```diff
@@ -761,7 +761,14 @@ def _on_stop_iteration(self, transaction):
...
                 # Concatenate all chunk texts together to get the full response text
-                full_content = "".join([chunk.text for chunk in streaming_events])
+                try:
+                    full_content = "".join([chunk.text for chunk in streaming_events])
+                except TypeError:
+                    # This is to account for tool calls, where the tool
+                    # call response contains the text that is required.
+                    # If not valid, this will trigger an AttributeError
+                    # and not record a streaming success (yet).
+                    full_content = kwargs["contents"][-1].parts[0].function_response.response["output"][0]["text"]
```
TimPansino, line 770: This line is pretty arcane, and I'm wondering how it will hold up to other `response_format` values. As far as I can tell we don't have coverage of any other formats other than this custom one, I'd like to at least be testing against the default tool format (JSON?).

tests/mlmodel_gemini/langchain_integration/conftest.py
```diff
@@ -0,0 +1,194 @@
...
+
+        agent_tools = loop.run_until_complete(mcp_client.get_tools())
+        Schema, State = schemas
+
+        return create_agent(
+            model=gemini_streaming_client,
+            name="my_agent",
+            tools=agent_tools,
+            state_schema=State,
+            response_format=ToolStrategy(Schema),
```
TimPansino: As I mentioned at the top of the PR I'm interested in seeing better coverage by parametrizing these 2 to be default values as well for the tests.

TimPansino (changes requested):

tox.ini
lrafeei: Oh, that was supposed to be reverted...let me change that.

tests/mlmodel_gemini/langchain_integration/conftest.py
lrafeei: Added a fixture for this to test with ToolStrategy as well as the default JSON

tox.ini
lrafeei: Moved this to the LangChain test suite

newrelic/hooks/mlmodel_gemini.py
```diff
@@ -713,12 +713,11 @@ def _parse_input_message(messages):
             return input_message, "user"
         # The input_message will be a Google Content type if send_message was called, so we parse out the message
         # text and role (which should be "user")
-        elif isinstance(input_message, google.genai.types.Content) and input_message.parts[0].text:
-            return input_message.parts[0].text, input_message.role
```
TimPansino, line 717: Is this supposed to be gone?

tests/testing_support/fixture/vcr.py
```diff
@@ -359,19 +359,26 @@ def pytest_collection_modifyitems(items):
...
 
 
-def pytest_recording_configure(config, vcr):
+# === Custom VCR Matchers ===
+
+
+def _match_method_case_insensitive(r1, r2):
     """
-    Register a case-insensitive method matcher with VCR.py so that requests with different
+    A case-insensitive method matcher fore VCR so that requests with different
```
TimPansino:
```suggestion
    A case-insensitive method matcher for VCR so that requests with different
```

newrelic/hooks/mlmodel_gemini.py
lrafeei: It is, and it was supposed to be removed in the missing User Input PR

lrafeei: To further explain, the logic below is intended to capture what this logic is trying to capture, but with some caveats:
1. The input message may not necessarily be the first one in the list of messages
2. `input_message.parts[0].text` will be available as an attribute but not necessarily populated.  This is an indicator that it is not, in fact, the input message that we are looking for and we should attempt to see if the next item in the list of messages is the Content type that contains our input message.

TimPansino (approved):

PR #1805, opened 2026-07-30
Swap to Trace Cache ID in LangChain and AgentFramework

TimPansino:
# Overview

* For context propagation purposes, swap the langchain and agent-framework instrumentation to only store the trace cache ID rather than the full trace in a ref.
  * This is to prevent these objects from getting serialized unexpectedly in the case of leaks. The ID is a much smaller object while the trace could contain MBs of data.


lrafeei (dismissed):

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-599926

lrafeei (approved):

PR #1813, opened 2026-08-06
OpenAI Responses API Instrumentation

TimPansino:
# Overview

* Instrument the newer `openai.responses` APIs.
* Add a new set of tests that mirrors the original `chat_completions` tests.


newrelic/hooks/mlmodel_openai.py
```diff
@@ -790,6 +799,172 @@ def _record_completion_error(*, transaction, linking_metadata, completion_id, kw
...
+    if kwargs.get("instructions"):
+        messages.append({"content": kwargs["instructions"], "role": "system"})
+    if isinstance(input_, str):
+        messages.append({"content": input_, "role": "user"})
+    elif isinstance(input_, (list, tuple)):
+        for item in input_:
+            if isinstance(item, dict) and item.get("role") is not None:
+                content = item.get("content")
+                # Content is a string, or a list of typed parts each carrying `text`.
+                if isinstance(content, (list, tuple)):
```
umaannamalai, line 815: For chat completions, we deliberately suppress content-less tool call messages (on line 595 by checking "tool_call" in `finish_reason`). Since we don't want to report every tool call as an `LlmChatCompletionMessage` with content: "", do we want to mirror that suppression behavior here if possible?

newrelic/hooks/mlmodel_openai.py
```diff
@@ -890,29 +1070,75 @@ def _record_stream_chunk(stream_proxy, return_val):
...
+    event_type = event.get("type")
+    if event_type == "response.output_text.delta":
+        delta = event.get("delta") or ""
+        if delta and "time_to_first_token" not in stream_proxy._nr_openai_attrs:
+            stream_proxy._nr_openai_attrs["time_to_first_token"] = (
+                int(1000.0 * time.time()) - stream_proxy._nr_request_timestamp
+            )
+        stream_proxy._nr_openai_attrs["content"] = stream_proxy._nr_openai_attrs.get("content", "") + delta
+        stream_proxy._nr_openai_attrs["role"] = stream_proxy._nr_openai_attrs.get("role") or "assistant"
+    elif event_type in ("response.completed", "response.incomplete", "response.failed"):
```
umaannamalai, line 1134: Are we capturing errors surfaced via `response.failed` somewhere? From my scan of the codebase, it seems like `APIErrors` are raised when the raw payload has a top-level `error` key. `response.failed` events nest errors inside `response.error` and the stream just ends normally afterwards. I just want to make sure we are capturing errors in this case also.

newrelic/hooks/mlmodel_openai.py
umaannamalai: Or is that why "role" is being checked for explicitly?

PR #1818, opened 2026-08-09
fix: pass application through get_linking_metadata

ckarnell:
`newrelic.agent.get_linking_metadata(application=...)` accepts the argument and drops it. `time_trace.py:746` calls `get_service_linking_metadata()` with nothing, so a caller passing an application still gets the default one's `entity.name` and `entity.guid`.

Three things in the same file say that was not deliberate. The helper it calls takes `application` and uses it to resolve settings. The identically named method on `TimeTrace` at line 694 passes it through to `_get_service_linking_metadata`. And `application=None` is the convention across a good number of public APIs here, including `notice_error` and `record_custom_event`, which this one member quietly breaks.

Nobody chose it. `git log -L` on the function puts it in `2c4f46bc`, "Record log event (#555)", the same commit that added `application` to the helper, and the line has not been touched since. No test pins the current behaviour: every test and benchmark calls it with no arguments.

The fix is one word.

Verified offline, no collector and no network. With a fake application whose settings name a different entity, the public function returns `None` for `entity.name` on main and `OTHER-APP` here, while the helper returns `OTHER-APP` both times. The added test fails without the source change and `tests/agent_features/test_logs_in_context.py` is 9 passed with it.

Severity is low. It only bites a process reporting to more than one application, since everyone else was getting the default they wanted anyway. Nothing leaks and nothing crashes, the metadata is just attributed to the wrong entity.

I did not run this against a collector or a real multi-application process, so the argument that it matters rests on reading, and I did not run the wider suite beyond that one file.

### DIFF
--- a/newrelic/api/time_trace.py
+++ b/newrelic/api/time_trace.py
@@ -744,7 +744,7 @@ def get_service_linking_metadata(application=None, settings=None):
 
 
 def get_linking_metadata(application=None):
-    metadata = get_service_linking_metadata()
+    metadata = get_service_linking_metadata(application)
     trace = current_trace()
     if trace:
         metadata.update(trace._get_trace_linking_metadata())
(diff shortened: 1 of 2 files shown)


TimPansino (approved):
LGTM, thanks for the fix!

PR #1822, opened 2026-08-12
Cache packages_distributions to Avoid Costly O(n^2) Performance Hit

### LINKED ISSUE #1817: packages_distributions() is called once per module name in _get_package_version(), rescanning all installed distributions
## Description

`_get_package_version()` is memoized per module name with `@lru_cache`, but the
`importlib_metadata.packages_distributions()` call inside it is not cached. That
function walks **every** installed distribution and reads its metadata, so the
agent repeats a full scan of the environment once for every distinct module name
it looks up.

https://github.com/newrelic/newrelic-python-agent/blob/main/newrelic/common/package_version_utils.py#L100-L104

```python
@lru_cache
def _get_package_version(name):
    ...
    if hasattr(importlib_metadata, "packages_distributions"):
        distributions = importlib_metadata.packages_distributions()   # full scan, not cached
        distribution_name = distributions.get(name, name)
```

The result of `packages_distributions()` is identical for every caller within a
process, so the cost scales as *(number of module names) x (number of installed
distributions)* when it could be paid once.

Two places make this add up:

- `newrelic/core/environment.py` `plugins()` iterates `sys.modules` and calls
  `get_package_version()` for each top-level module. `Application._harvest()`
  consumes that generator with a 2 second budget per harvest, so a large
  environment keeps scanning across many harvest cycles after startup.
- Instrumentation hooks look up versions when modules are first imported, which
  keeps triggering scans during request handling.

## Expected Behavior

`packages_distributions()` should be evaluated at most once per process (or the
top-level-package to distribution mapping computed once and reused), so version
lookups cost one scan in total rather than one scan per module name.

## Steps to Reproduce

Environment with 242 installed distributions, agent 10.16.0, Python 3.11.13.

```python
import importlib.metadata
import sys
import time

from newrelic.common.package_version_utils import get_package_version

calls = {"n": 0, "seconds": 0.0}
original = importlib.metadata.packages_distributions


def counting_packages_distributions():
    calls["n"] += 1
    start = time.perf_counter()
    try:
        return original()
    finally:
        calls["seconds"] += time.perf_counter() - start


importlib.metadata.packages_distributions = counting_packages_distributions

start = time.perf_counter()
one = original()
print(f"installed distributions      : {len(one)}")
print(f"one packages_distributions() : {(time.perf_counter() - start) * 1000:.0f} ms")

names = sorted({m for m in sys.modules if "." not in m and not m.startswith("_")})[:25]
start = time.perf_counter()
for name in names:
    get_package_version(name)
total = time.perf_counter() - start

print(f"\nget_package_version() for {len(names)} distinct names")
print(f"  wall time                   : {total * 1000:.0f} ms")
print(f"  packages_distributions calls: {calls['n']}")
print(f"  time inside those scans     : {calls['seconds'] * 1000:.0f} ms "
      f"({calls['seconds'] / total * 100:.0f}% of wall time)")
```

Output:

```
installed distributions      : 242
one packages_distributions() : 184 ms

get_package_version() for 25 distinct names
  wall time                   : 1625 ms
  packages_distributions calls: 23
  time inside those scans     : 1623 ms (100% of wall time)
```

23 scans for 25 names, and the scans account for essentially all of the wall
time. A real process resolves several hundred module names, not 25.

## Additional context

This is more than a startup cost on cooperatively scheduled servers. We run
gunicorn with `gevent` workers, where file I/O does not yield to the hub, so each
scan stalls every other greenlet in the worker for its whole duration. Using
gevent's own event loop monitor (`GEVENT_MONITOR_THREAD_ENABLE`) with a 0.2 s
threshold, the captured blocking stacks were dominated by this path:

```
pathlib.py:1044:open
  < pathlib.py:1058:read_text
  < importlib_metadata/__init__.py:955:read_text
  < importlib/metadata/__init__.py:1081:_top_level_declared

contextlib.py:446:__exit__
  < importlib_metadata/__init__.py:948:read_text
  < importlib_metadata/__init__.py:491:metadata
  < importlib/metadata/__init__.py:1076:packages_distributions

email/feedparser.py:84:readline
  < email/feedparser.py:128:__next__
  < email/feedparser.py:462:_parsegen
  < email/parser.py:178:_call_parse
```

Setting `NEW_RELIC_PACKAGE_REPORTING_ENABLED=false` removed the harvest-driven
scans, which confirmed that path, but version lookups from instrumentation hooks
still scan on first import of each module.

Caching `packages_distributions()` for the life of the process resolves it
locally, but that means monkeypatching a standard library function from
application code, which we would rather not carry.

## Your Environment

- Python agent: 10.16.0 (the same code is present on `main` as of this writing)
- Python: 3.11.13
- OS: Debian bookworm (container), also reproduced on macOS
- Installed distributions: 242
- Server: gunicorn 23.0.0 with `gevent` 25.9.1 workers, Django 4.2.21

TimPansino:
# Overview

* Reduce overhead of package lookups by caching the `importlib.metadata.packages_distributions()` dict after first lookup. 
  * As this scans the filesystem for every package on each version lookup, it was incurring a total O(n^2) time penalty scaling with the number of installed packages on agent initialization.
 
# Related Issues

Closes #1817


newrelic/common/package_version_utils.py
```diff
@@ -96,14 +103,16 @@ def _get_package_version(name):
...
-            distribution_name = distribution_name[0] if isinstance(distribution_name, list) else distribution_name
-        else:
-            distribution_name = name
+        # Cached lookup for packages_distributions() to avoid scanning the filesystem
+        # every time we need to check a package verison.
+        if not _packages_distributions and hasattr(importlib_metadata, "packages_distributions"):
+            _packages_distributions = importlib_metadata.packages_distributions()
+
+        # Try to grab the package's distribution name, and fallback to just the package name if we can't find it.
+        distribution_names = _packages_distributions.get(name, [])
```
hmstepanek: Seems like this would be simpler:

```suggestion
        distribution_names = _packages_distributions.get(name, [name])
```

newrelic/common/package_version_utils.py
```diff
@@ -96,14 +103,16 @@ def _get_package_version(name):
...
+        distribution_name = distribution_names[0] if distribution_names else name
```
hmstepanek:
```suggestion
        distribution_name = distribution_names[0]
```

hmstepanek (dismissed):
Looks good. Just had one suggestion to simplify the logic but not a big deal either way.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-608845

hmstepanek (dismissed):

hmstepanek (approved):

PR #1825, opened 2026-08-17
Instrument Azure CosmosDB

lrafeei:
Instrumentation for Azure CosmosDB and testing


.github/workflows/tests.yml
```diff
@@ -576,6 +577,101 @@ jobs:
...
+          # Add a timeout to avoid infinite loops
+          RETRIES=0
+          until curl -k -s https://cosmosdb:8081/_explorer/index.html > /dev/null; do
+            RETRIES=$((RETRIES + 1))
+            if [ "$RETRIES" -ge 30 ]; then
+              echo "ERROR: CosmosDB explorer not reachable after 60s"
+              exit 1
+            fi
+            sleep 2
+          done
```
TimPansino, line 631: Why is this section necessary? Shouldn't this be covered by the healthcheck?

tests/datastore_azurecosmos/test_database_to_container_calls.py
```diff
@@ -0,0 +1,194 @@
...
+
+@validate_transaction_metrics(
+    "test_database_to_container_calls:test_database_container_queries",
+    scoped_metrics=_scoped_container_query,
+    rollup_metrics=[*_scoped_container_query, *_base_container_query],
+    background_task=True,
+)
+@background_task()
+def test_database_container_queries(loop, async_database):
+    async def _test_database_container_queries():
```
TimPansino: You've got a name collision here, needs an `async_` in the name.

.github/workflows/tests.yml
lrafeei: For some reason, it seems as though the ports are not ready at the same time (this one takes extra long to be ready).  Should I just make the health check check to see if the exporter file is ready?  Rather than the built in health check, that is.

TimPansino (approved):

.github/workflows/tests.yml
TimPansino: If that's possible then yes, let's make docker handle as much of the service health lifecycle as possible.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-608853

PR #1841, opened 2026-09-03
Agent Framework Instrumentation for Agents

TimPansino:
# Overview

* Add new instrumentation for Agent Framework agents.
* Refactor test suite and test for new agent events.


newrelic/hooks/mlmodel_agentframework.py
```diff
@@ -13,26 +13,166 @@
...
+    agent_id = str(uuid.uuid4())
+
+    try:
+        return_val = wrapped(*args, **kwargs)
+    except Exception:
+        ft.__exit__(*sys.exc_info())
+        raise
+
+    # Handle streaming case by registering a cleanup hook, as proxying the response object is very difficult.
+    if hasattr(return_val, "with_cleanup_hook"):
```
umaannamalai, line 71: If `with_cleanup_hook()` itself raises, this calls ft.__exit__(*sys.exc_info()) which marks the span with an error even though the actual Agent.run() call already succeeded. Should this be `ft.__exit__(None, None, None)` i\since the underlying operation didn't fail?

Also, should we add a warning log here for consistency?

newrelic/hooks/mlmodel_agentframework.py
```diff
@@ -13,26 +13,166 @@
...
+    ft.__enter__()
+    ft._add_agent_attribute("subcomponent", json.dumps(agentic_subcomponent_data))
+
+    linking_metadata = get_trace_linking_metadata()
+    agent_id = str(uuid.uuid4())
+
+    try:
+        return_val = wrapped(*args, **kwargs)
+    except Exception:
+        ft.__exit__(*sys.exc_info())
```
umaannamalai, line 67: Should we have a  `_record_agent_event(error=True)` call here?

newrelic/hooks/mlmodel_agentframework.py
```diff
@@ -61,6 +201,12 @@ def wrap_BedrockChatClient__prepare_options(wrapped, instance, args, kwargs):
     return request
 
 
+def instrument_agent_framwork__agents(module):
```
umaannamalai, line 204:
```suggestion
def instrument_agent_framework__agents(module):
```

newrelic/hooks/mlmodel_agentframework.py
```diff
@@ -61,6 +201,12 @@ def wrap_BedrockChatClient__prepare_options(wrapped, instance, args, kwargs):
...
+    if hasattr(module, "Agent"):
+        if hasattr(module.Agent, "run"):
+            wrap_function_wrapper(module, "Agent.run", wrap_Agent_run)
+
+
 def instrument_agent_framwork_bedrock__chat_client(module):
```
umaannamalai, line 210:
```suggestion
def instrument_agent_framework_bedrock__chat_client(module):
```

newrelic/hooks/mlmodel_agentframework.py
```diff
@@ -13,26 +13,166 @@
...
+
+    try:
+        return_val = wrapped(*args, **kwargs)
+    except Exception:
+        ft.__exit__(*sys.exc_info())
+        raise
+
+    # Handle streaming case by registering a cleanup hook, as proxying the response object is very difficult.
+    if hasattr(return_val, "with_cleanup_hook"):
+        try:
```
umaannamalai, line 72: Would the cleanup hook fire when the stream doesn't close or get iterated on?  I know we had a somewhat related issue with Autogen previously.

PR #1856, opened 2026-09-18
Fix Anthropic token/content capture for langchain-anthropic (with_raw_response.create)

### LINKED ISSUE #1855: Anthropic instrumentation drops token/content capture for messages.with_raw_response.create (breaks langchain-anthropic / ChatAnthropic)
## Summary

When the Anthropic SDK is called through `client.messages.with_raw_response.create(...)`, the Python agent records an `LlmChatCompletionSummary` (`vendor = "anthropic"`) but with **no token/usage data** — `response.usage.*` is never populated, so AI Monitoring reports 0 tokens (and the completion content is not captured). A direct `client.messages.create(...)` call captures tokens and content correctly under identical conditions.

This matters in practice because **`langchain-anthropic`'s `ChatAnthropic` calls the SDK exclusively via `with_raw_response`**, so LangChain / LangGraph apps on Anthropic silently lose token + completion-content capture even though the summary events are recorded.

## Environment

- `newrelic==13.5.0`
- `anthropic==1.6.0`
- `langchain-anthropic==1.7.2`, `langchain-core==1.6.3` (the repro needs only `langchain-anthropic`)
- Python 3.14
- `ai_monitoring.enabled = true`, `ai_monitoring.record_content.enabled = true`, `ai_monitoring.streaming.enabled = true` (default)

## Reproduction

Run under `newrelic-admin run-program` (so import order is not a factor). Same process, same model; the two paths are tagged so they're separable in NRDB:

```python
import os
os.environ["ANTHROPIC_BASE_URL"] = "https://api.anthropic.com"

import newrelic.agent          # already initialized by newrelic-admin
from anthropic import Anthropic
from langchain_anthropic import ChatAnthropic

MODEL = "claude-haiku-4-5-20251001"
_sdk = Anthropic()
_lc = ChatAnthropic(model=MODEL, max_tokens=64)

@newrelic.agent.background_task(name="repro_sdk")
def via_sdk():
    newrelic.agent.add_custom_attribute("llm.path", "sdk")
    _sdk.messages.create(model=MODEL, max_tokens=64,
                         messages=[{"role": "user", "content": "say hi in 3 words"}])

@newrelic.agent.background_task(name="repro_chatanthropic")
def via_langchain():
    newrelic.agent.add_custom_attribute("llm.path", "chatanthropic")
    _lc.invoke("say hi in 3 words")

for _ in range(8):
    via_sdk(); via_langchain()
newrelic.agent.shutdown_agent(timeout=30)
```

```
newrelic-admin run-program python repro.py
```

Then query the events:

```sql
SELECT count(*) AS summaries,
       filter(count(*), WHERE `response.usage.total_tokens` IS NOT NULL) AS with_usage,
       sum(`response.usage.total_tokens`) AS total_tokens
FROM LlmChatCompletionSummary FACET `llm.path`, vendor SINCE 30 minutes ago
```

| llm.path | vendor | summaries | with `response.usage` | total_tokens |
|---|---|---|---|---|
| `sdk` | anthropic | 4 | **4** | 86 |
| `chatanthropic` | anthropic | 4 | **0** | 0 |

Both paths record `vendor = anthropic` summaries; only the direct-SDK path carries usage. (8 calls were issued per path; the ratio is the point — 100% of SDK-path summaries carry `response.usage`, 0% of ChatAnthropic-path summaries do.)

## Root cause

`newrelic/hooks/mlmodel_anthropic.py` (`instrument_anthropic_messages`) wraps `Messages.create` / `Messages.stream` (and the async variants). `langchain-anthropic` calls `self._client.messages.with_raw_response.create(**payload)` (`langchain_anthropic/chat_models.py`; there is also a `beta.messages.with_raw_response.create` variant).

In the SDK, `messages.with_raw_response.create` is `to_raw_response_wrapper(messages.create)`, so the New Relic wrapper on `Messages.create` **does** fire (that is why a `vendor=anthropic` summary is recorded at all) — but the object returned to the wrapper is the raw API response (`LegacyAPIResponse` / `APIResponse`), not a parsed `Message`. The hook then does:

```python
usage = getattr(response, "usage", None)          # mlmodel_anthropic.py (~L168 sync / ~L256 async)
...
input_tokens=getattr(usage, "input_tokens", None) if usage else None,
output_tokens=getattr(usage, "output_tokens", None) if usage else None,
```

On a raw-response wrapper `usage` is `None`, so no `response.usage.*` is recorded, and the assistant content isn't parsed from the raw response either.

## What this is *not*

- Reproduced under `newrelic-admin run-program`, so it is **not** the known LangChain import-priority issue.
- Reproduces with a bare `ChatAnthropic().invoke()` — no `create_agent`, no LangGraph — so it is below the agent layer; LangChain tool/agent instrumentation itself works (`LlmTool` events are recorded fine in a fuller app).
- The direct SDK path works identically in the same process, isolating the `with_raw_response` entrypoint as the only difference.

## Suggested fix

Handle the raw-response case in the Anthropic hook — e.g. instrument `MessagesWithRawResponse.create` (and the `beta` variants), or, when the wrapped call returns a raw API response, `.parse()` it (or read the underlying payload) before extracting `usage`/content.

adjohn:
## Summary

`langchain-anthropic`'s `ChatAnthropic` calls the Anthropic SDK via `client.messages.with_raw_response.create(...)` (and the `beta.messages` variant). That routes through `Messages.create` — so the instrumentation fires and an `LlmChatCompletionSummary` (`vendor=anthropic`) is recorded — but the object handed to the wrapper is the raw `APIResponse`/`LegacyAPIResponse`, not a parsed `Message`. `getattr(response, "usage")`, `.content`, `.id`, etc. are therefore `None`, so **token counts and completion content are silently dropped** for every LangChain / LangGraph app on Anthropic, even though the summary events show up.

Fixes #1855.

## Fix

Parse the raw response before extracting attributes; the original raw object is still returned to the caller untouched (`.parse()` caches, so the caller's own `.parse()` is unaffected). The async client is handled too, where `AsyncAPIResponse.parse()` is a coroutine and must be awaited. On any failure the code falls back to the original object so instrumentation never breaks the call.

## Test

Adds a `create.with_raw_response` case to the parametrized `interaction_method` fixture in `tests/mlmodel_anthropic/conftest.py`, so the raw-response path is exercised across the existing success and error tests. It fails before this fix (no `response.usage.*`) and passes after.

## Verification

Reproduced and verified end-to-end against a live (staging) account — bare `ChatAnthropic().invoke()` vs `Anthropic().messages.create()` in one process, tagged so the two paths are separable:

| path | summaries | with `response.usage` | total tokens |
|---|---|---|---|
| SDK `messages.create` | 10 | 10 | 217 |
| `ChatAnthropic` (before fix) | — | 0 | 0 |
| `ChatAnthropic` (after fix) | 10 | 10 | 214 |

Reproduced under `newrelic-admin run-program`, so this is **not** the known LangChain import-priority issue; and it reproduces with a bare `ChatAnthropic().invoke()` (no agent / LangGraph), so it is below the agent layer.

## Notes

- Environment: `anthropic==1.6.0`, `langchain-anthropic==1.7.2`, Python 3.14.
- I verified the fix empirically (above) but was not able to run the full `tox` suite locally — please run CI, especially the newly-parametrized `mlmodel_anthropic` error tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


adjohn:
Related to #1800, which fixed the same class of "missing LLM data when LangChain/LangGraph drives the model via a non-default response shape" for the **Gemini** hook (streaming tool-call content extraction).

This PR is the **Anthropic** counterpart: `langchain-anthropic`'s `ChatAnthropic` calls `messages.with_raw_response.create`, so `mlmodel_anthropic` receives a raw `APIResponse`/`LegacyAPIResponse` with no `.usage`/`.content` and token + completion capture goes missing (the non-streaming path). Same spirit of fix as #1800 — handle the alternate response shape in place (`.parse()` it) and fall back gracefully rather than breaking.

TimPansino:
This could really use regression tests in LangChain but I'll leave that for another PR, we'll have to expand the infra in those tests to include anthropic.

TimPansino (dismissed):

TimPansino (approved):

PR #1858, opened 2026-09-21
OpenSearch Instrumentation

TimPansino:
# Overview

* Add instrumentation for `opensearch-py` package.
* Add new test runners with OpenSearch as a container.
* Add full test suite for `opensearch-py` ported from existing `elasticsearch` test suite.


tests/datastore_elasticsearch/test_async_transport.py
```diff
@@ -1,6 +1,6 @@
 # Copyright 2010 New Relic, Inc.
 #
-# Licensed under the Apache License, ES_VERSION 2.0 (the "License");
```
lrafeei, line 3: Ha, good catch

lrafeei (approved):

